Issue #23, opened 2023-11-09
fmt_number is mutating source data, so multiple calls of same code produces error


machow:
I'll look into it..!

```python
from great_tables import GT
from great_tables.data import exibble

gt_ex = GT(exibble)

# works
gt_ex.fmt_number(columns="num", decimals=1, scale_by=1/1000, pattern="{x}K")

# errors because num is now strings
gt_ex.fmt_number(columns="num", decimals=1, scale_by=1/1000, pattern="{x}K")
```

rich-iannone:
@machow It still seems to be an issue with the `fmt_integer()` example at https://posit-dev.github.io/great-tables/examples-qmd/fmt-number.html

PR #24, opened 2023-11-09
fix: do not mutate sourcenotes or body

machow:
This PR fixes SourceNotes and Body code, so it does not mutate the underlying data. This was causing issues where re-running code could continuously append notes, or error due to body mutations (e.g. a numeric column becoming a string column)

Note that I removed the custom SourceNotes class, in favor of using a list of strings


machow:
addresses #23

rich-iannone (approved):
Looks great to me!

PR #25, opened 2023-11-09
docs: getting started

machow:
This PR fleshes out the intro page, and adds a sidebar so it's easy to navigate between our development example pages.


machow:
It's rough, but I think once we get spanners and row & group names rendering, we can spruce it up! I'm going to shift to getting deploying to pypi via CI and rendering docstring examples...!

machow:
It's weird, but I think the diff is unfortunately some of the changes from the fix-mutations PR :/

rich-iannone (approved):
This is super good. Thanks!

PR #26, opened 2023-11-09
ci: deploy to pypi on release

machow:
This PR deploys to pypi whenever we make a github release. It only runs if the tests pass. However, note it does not require the docs to build (since they are in a different workflow, and it's a bit tricky, but not impossible to do?!). There are ways to test that don't involve full blown attempting to deploy. But given how early it is, I vote for us merging, doing a release, and then seeing what happens 😈.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -33,3 +33,24 @@ jobs:
       - name: pytest unit tests
         run: |
           make test
+
+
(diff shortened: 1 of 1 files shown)


machow:
@rich-iannone whenever you're up for it, want to try merging and making a release? 😁

rich-iannone:
Let's do this!

rich-iannone (approved):

PR #31, opened 2023-11-20
Docs fix upgrade quartodoc

machow:
This PR upgrades to the latest quartodoc, which can correctly document the methods on the `GT` class.

Note that I had to remove the methods section from GT, which apparently quartodoc does not support  😓


rich-iannone (approved):
Great!

PR #36, opened 2023-11-29
fix: tab_header no longer mutates GTData

machow:
This PR changes `tab_header()`, so that it creates and returns a copy of GTData, and puts a new Heading object on it (vs mutating the existing heading).

This prevents heading mutations from creating headings in places where tab_heading isn't called.


rich-iannone (approved):
LGTM!

PR #37, opened 2023-11-30
fix: static tab completions for GT, and label rendering

machow:
This PR enables static tab completions off of GT. The challenge is that methods often enable this using `Self`. Since we define functions (which can't use Self), and then put them on a class, I had to basically define Self the long way.

(Self the long with is a TypeVar with a bound)


rich-iannone (approved):
LGTM!

PR #49, opened 2023-12-04
fix: spanners_print_matrix now excludes selected cols that are in the stub

machow:
This PR ensures that `spanners_print_matrix` does not include columns selected for a spanner, if those columns are in the stub.

For example, `currency` below was bumping other columns in header over:

```python

gt = GT(exibble, rowname_col="currency").tab_spanner("A", ["currency", "group", "row"])
```

image

Now it doesn't!:

image


rich-iannone (approved):
Excellent!

PR #50, opened 2023-12-04
refactor: freeze dataclasses

machow:
This PR freezes our dataclasses so we couldn't mutate them if we wanted to†. The `GTData._replace()` method can be used to return a copy of GTData, with specific attributes replaced.

I'm going to take out the API classes, so we can type their methods a bit better (currently, the type system doesn't know they can access a `._replace` method)

†: except by using `super().__setattr__("<attribute_name>", <value>)`, which should be used sparingly!

TODO:

* code paths it'd be helpful to tour Michael through:
  - locales
  - options


rich-iannone (approved):
LGTM!

PR #53, opened 2023-12-05
tests: add test of options

machow:
This PR adds a test file for the options api with 1 test.


machow:
It looks like I missed a line of code that tried accessing options in the old style, which caused one of the examples on the site to fail. I've added a fix here. @rich-iannone do you know why examples can fail without causing `quarto render` to fail? I'm not sure at which layer of the stack it's happening on (e.g. is a quarto thing, or great_tables thing?)

image

(Example should be fixed now!)

rich-iannone (approved):
LGTM!

PR #54, opened 2023-12-05
Test tab create modify

machow:
This PR adds a test to tab create modify that checks when we select multiple columns for styling


rich-iannone (approved):
LGTM!

PR #59, opened 2023-12-06
feat: shiny output and renderer for GT

machow:
This PR adds a `great_tables.shiny` module, with functions for rendering (`render_gt`) and ui output (`output_gt`).

This doesn't necessarily need to go into great_tables. It could also live in shiny. **Let's figure out the shiny's team preference, since it is early days on this type of extending in shiny.**

Here is a demo shiny app:

```python
from shiny import App, ui

import great_tables.shiny as gts


app_ui = ui.page_fluid(gts.output_gt("table"))


def server(input, output, session):
    @output
    @gts.render_gt
    def table():
        return GT(exibble)


app = App(app_ui, server)
```

Alternatively, it seems like shiny should have a mechanism to let classes like `GT` hook into output_ui? Something similar to the `._repr_html_()` method? Or maybe that exact method itself (this is exactly what we call in the shiny Transformer anyway)?

### DIFF
--- a/great_tables/shiny.py
+++ b/great_tables/shiny.py
@@ -1,0 +1,42 @@
+from __future__ import annotations
+
+__all__ = (
+    "output_gt",
+    "render_gt",
+)
+
+from great_tables import GT
+from shiny.render.transformer import (
+    output_transformer,
+    TransformerMetadata,
+    ValueFn,
+    resolve_value_fn,
+)
+from shiny.session._utils import RenderedDeps
+from shiny._namespaces import resolve_id
+
+from htmltools import Tag, div, HTML
+
+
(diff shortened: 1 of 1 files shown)


machow:
I tested this + #61 out on shinylive, and it works! (I just uploaded a built wheel to a public gcs bucket).

image

rich-iannone (approved):
Looks great!

machow:
I've added a friendly import error message for when user's don't have shiny installed. I had to change the import error we raise when users to try import data from the top-level (e.g. `from great_tables import airquality`), since apparently it needs to raise AttributeError, in order to not intercept the shiny submodule error!

PR #61, opened 2023-12-06
refactor!: replace libsass code with webcolors

machow:
This PR replaces libsass, by using python + the webcolors package to assemble the css used by GT.

The main changes are:

* use `string.Template` to fill in parameters in the css file
* replace the function `font-color` with variables like `font_color_table_background`
* add a python function `font_color()` for generating the font color values.

I added some basic tests of `font_color`, but it could be useful to try and test a more exhaustive set. I think the main weakness of this process like hinges on `font_color()` being as robust as `scss` functions like `color.red()`.

TODO: I haven't fully removed the libsass code, but can do it before merging, if this ends up seeming like the way to go!

### DIFF
--- a/great_tables/_scss.py
+++ b/great_tables/_scss.py
@@ -1,0 +1,126 @@
+from __future__ import annotations
+
+import pkg_resources
+import re
+import webcolors as wc
+
+from dataclasses import fields
+from functools import partial
+from typing import Optional
+from string import Template
+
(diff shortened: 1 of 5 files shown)


rich-iannone (approved):
LGTM!

machow:
Thanks! I'll remove the libsass code and then merge!

Issue #73, opened 2023-12-09
tab_source_note disappears when text is rendered as markdown


karangattu:
## Description

`tab_source_note` disappears when text is rendered as markdown

## Reproducible example
using
```python
    .tab_source_note(
        source_note = "Source: The World Almanac and Book of Facts, 1975, page 406."
    )
```
renders the text correctly
<img width="458" alt="great_tables_-_Examples" src="https://github.com/posit-dev/great-tables/assets/4220325/8963e44a-66c1-4d77-b38c-1a8fe05ee6e5">
but adding
```python
    .tab_source_note(
        source_note = md("Reference: McNeil, D. R. (1977) *Interactive Data Analysis*. Wiley.")
    )
```
Does not render the markdown and hides the text from the footnotes section

## source code
```python
from great_tables import GT, md, html
from great_tables.data import islands

islands_mini = islands.head(10)

(
    GT(islands_mini, rowname_col = "name")
    .tab_header(
        title = "Large Landmasses of the World",
        subtitle = "The top ten largest are presented"
    )
    .tab_source_note(
        source_note = "Source: The World Almanac and Book of Facts, 1975, page 406."
    )
    .tab_source_note(
        source_note = md("Reference: McNeil, D. R. (1977) *Interactive Data Analysis*. Wiley.")
    )
    . tab_stubhead(label = "landmass")
)
```

## Expected result

Should render footnote with markdown when used in conjunction with the `md` tag
## Development environment

- Operating System: macOS
- **great_tables** Version: 0.1.1

## Additional context

This code is also displayed on the examples page

rich-iannone:
Thanks for reporting. Will fix this soon!

PR #77, opened 2023-12-11
Ensure that `tab_source_note()` works when using `md()` or `html()`

rich-iannone:
This fixes an issue where using `md()` or `html()` on source note text (within `tab_source_note()`) would fail to produce the note. The solution is to apply `_process_text()` to incoming label text, handling the md- and html-classed text appropriately. After this fix is applied, the following example works as expected:

```
import great_tables as gt
from great_tables import md, html
from great_tables import exibble

(
    gt.GT(exibble)
    .tab_source_note(md("An **important** note."))
    .tab_source_note(md("Another *important* note."))
    .tab_source_note("A plain note.")
    .tab_source_note(html("An <strong>HTML heavy</strong> note."))
)
```

Fixes: #73

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -528,10 +528,12 @@ def create_source_notes_component_h(data: GTData) -> str:
         source_notes_tr: List[str] = []
 
         for note in source_notes:
+            note_str = _process_text(note)
+
             source_notes_tr.append(
                 f"""
   <tr>
-    <td class="gt_sourcenote" colspan="{n_cols_total}">{note}</td>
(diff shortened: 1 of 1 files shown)


machow (changes requested):
Can we add a test for this? I wonder if just doing a snapshot in test_formats (or a new file for test_utils_render_html?). We could use the [assert_repr_html](https://github.com/posit-dev/great-tables/blob/main/tests/test_formats.py#L24) helper

(feel free to merge and punt testing)

rich-iannone:
Good idea, just added basic snapshot testing based on the example above.

machow (approved):
This test is way better than what I suggested, thanks!

PR #78, opened 2023-12-11
ci: pre-commit checks

machow:
This PR addresses #4 by running pre-commit checks in CI.


machow:
@rich-iannone I fixed a few of the errors/warnings flagged by flake8, and then commented out the rest, so we can go over them when pairing! (I don't know if an exclamation is the right level for flake8 errors but maybe it is?!)

rich-iannone (approved):
This is really great!

PR #81, opened 2023-12-12
docs: single column examples on narrow screens

machow:
Currently, the example page shows two columns of examples, even when on narrow screens (e.g. on mobile). This PR changes examples to only display in two columns on large screens.

**before:**

image

**after:**

image


rich-iannone (approved):
Perfect! Ship it.

PR #83, opened 2023-12-13
feat: add style.from_column, implement for loc.body

machow:
This PR implements the `from_column()` behavior from gt, which allows for setting a style based on values in the data.

e.g.

```python
import pandas as pd
from great_tables import GT, exibble, loc, style

df = pd.DataFrame({"x": [1, 2], "color": ["red", "blue"]})

gt = GT(df)
gt.tab_style(
    style = style.text(color = style.from_column("color")),
    locations = loc.body(columns = ["x"])
)
```

image

NOTES:

* **scope**: currently only implemented for setting in `loc.body()`, raises if you try to use it with `loc.title()`
* **implementation**: the R gt library handles from_column inside each `fmt_*` function. This PR puts the handling inside `set_style`, by implementing a CellStyle._from_row() method. This gives each style the chance to return a new version of itself based on the data, just before setting it on the GT object.
* **importing**: I put it under `style.from_column()`, but another place it could go is with helpers like `md()` at the top-level of the package?

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -110,6 +110,7 @@ quartodoc:
         - style.fill
         - style.text
         - style.borders
+        - style.from_column
     - title: Helper functions
       desc: >
         An assortment of helper functions is available in the **Great Tables** package. The `md()`
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -110,7 +110,6 @@ quartodoc:
         - style.fill
         - style.text
         - style.borders
-        - style.from_column
     - title: Helper functions
       desc: >
         An assortment of helper functions is available in the **Great Tables** package. The `md()`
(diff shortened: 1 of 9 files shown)


machow:
Reference page!

image

rich-iannone (changes requested):
This is great! I think the main changes needed (as discussed) are: (1) having `from_column()` is not part of `style`, (2) enabling support for `pl.col()` from Polars.

machow:
Alright, now we have:

* support for polars expressions in addition to `from_colums()`
* `from_columns` now only exported at top-level (i.e. `from great_tables import from_column`)

Note that I added a new internal class called FromValues, to represent a materialized column of values (e.g. from the polars Series that results from executing the polars expression). I didn't move FromColumn out of `_styles.py`, but was mostly me being unsure of where to put it. I wonder if we might just leave it there until it gets used for other things? (e.g. `fmt_*`)

machow:
Here's the example from the from_column docs

image

rich-iannone (approved):
LGTM!!!

machow:
noyce

Issue #85, opened 2023-12-13
When multiple locations specified in single tab_style called, only the last is applied


machow:
It looks like when tab_style gets a list of locations, only the last is applied to the data

```python
from great_tables import GT
from great_tables.data import airquality

gt_air = GT(airquality.head())
gt_air.tab_style(
    style.fill("yellow"),
    [loc.body("Temp", [1, 2]), loc.body("Ozone", [0])]
)
```

<img width="804" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/afd43985-0bf7-430e-b2f2-97452ace1158">

rich-iannone:
Fixed by https://github.com/posit-dev/great-tables/pull/87.

Issue #84, opened 2023-12-13
when multiple styles specified in single tab_style call, only first is applied


machow:
It appears that when multiple styles are specified in a tab_style call, only the first appears.

```python
from great_tables import GT
from great_tables.data import airquality

gt_air = GT(airquality.head())

gt_air.tab_style(
    [style.fill("yellow"), style.borders(sides = "all")],
    loc.body("Temp", [1, 2])
)
```

<img width="858" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/052e6f3f-f71a-47f4-9356-449a1b8e505a">

Note that if you switch the styles in the list around, it seems like only the first appears.

rich-iannone:
Fixed by https://github.com/posit-dev/great-tables/pull/87.

PR #87, opened 2023-12-14
fix: allow lists of styles and lists of locations

machow:
This PR addresses #84 and #85 by fixing support for lists of styles and locations. Previously, only the first style in a StyleInfo.styles list was used, and the locations code needed a lil tweak to accumulate locations


rich-iannone (approved):
👍

Issue #80, opened 2023-12-12
Add a formatting method that transforms URLs to links


rich-iannone:
For any cells containing URLs, we should have a `fmt_url()` method to make them navigable links. We should be careful to employ the method only on column containing URL text (though it is probably the user's responsibility to ensure this and that there are, for example, no URLs as part of a larger block of text). 

We could expose options for how the links should be styled. They can be of the conventional style (with underlines and text coloring that sets it apart from other text), or, they can appear to be button-like (with a surrounding box that can be filled with a color).

URLs in data cells could be detected in two ways:

-  using the simple Markdown notation for URLs of the form: `[label](URL)`
- assumes that the text is simply the URL

For the second case the URL is also used as the label but there is the option to use the `label` argument to modify that text.

kmasiello:
Also include ability to format email addresses as [address](mailto:address)

PR #90, opened 2023-12-15
Add the `fmt_url()` method

rich-iannone:
This adds the `fmt_url()` method, which allows for the formatting of cells that contain URLs. This method can be used to transform the text into navigable links. This should ideally be used on columns that contain *only* URL text (i.e., no URLs as part of a larger block of text).

Fixes: https://github.com/posit-dev/great-tables/issues/80


great_tables/_formats.py
```diff
@@ -1771,6 +1771,209 @@ def fmt_time_fn(
...
+
+            if button_outline == "auto":
+                button_outline_color = "#BEBEBE"
+                button_outline_style = "none"
+
+        elif button_fill != "auto" and color == "auto":
+            if button_outline == "auto":
+                button_outline_color = "#DFDFDF"
+
+                if button_fill in [
```
machow, line 1891: I wonder if it'd be helpful to define lists like this at the top of the function, with an informative name like...

LIGHT_FILL_VALUES

great_tables/_formats.py
```diff
@@ -1771,6 +1771,209 @@ def fmt_time_fn(
...
+            button_width = None
+
+        button_outline_color = button_outline
+        button_outline_style = "solid"
+        button_outline_width = "2px"
+
+        # There are various combinations of "auto" or not with `button_fill` and
+        # `color` that need to be handled delicately so as to ensure contrast
+        # between foreground text and background fill is maximized
+        if button_fill == "auto" and color == "auto":
```
machow, line 1865: Does this `if` block not need to set button_outline_color? (which could be "auto")

machow:
pairing w/ Rich -- we'll revisit down the road

PR #92, opened 2023-12-15
Ensure that ID value in compiled CSS is applied to all rules

rich-iannone:
This fixes a `re.sub()` statement in the `compile_scss()` function so that the table ID is inserted before the rule for the `p` element. This corrects a regression where the compiled CSS no longer worked with one of the `re.sub()` statements. With this fix, the rule for the table's `<p>` element is specific to the table (and doesn't affect the page itself).

### DIFF
--- a/great_tables/_scss.py
+++ b/great_tables/_scss.py
@@ -138,7 +138,7 @@ def compile_scss(data: GTData, id: Optional[str], compress: bool = True) -> str:
     if has_id:
         compiled_css = re.sub(r"\.gt_", f"#{id} .gt_", compiled_css, 0, re.MULTILINE)
         compiled_css = re.sub(r"thead", f"#{id} thead", compiled_css, 0, re.MULTILINE)
-        compiled_css = re.sub(r"^p \{", f"#{id} p " + "{", compiled_css, 0, re.MULTILINE)
+        compiled_css = re.sub(r"^ p \{", f" #{id} p " + "{", compiled_css, 0, re.MULTILINE)
 
     finalized_css = f"{gt_table_class_str}\n\n{compiled_css}"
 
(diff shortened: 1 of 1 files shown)


great_tables/_scss.py
```diff
@@ -138,7 +138,10 @@ def compile_scss(data: GTData, id: Optional[str], compress: bool = True) -> str:
     if has_id:
         compiled_css = re.sub(r"\.gt_", f"#{id} .gt_", compiled_css, 0, re.MULTILINE)
         compiled_css = re.sub(r"thead", f"#{id} thead", compiled_css, 0, re.MULTILINE)
-        compiled_css = re.sub(r"^p \{", f"#{id} p " + "{", compiled_css, 0, re.MULTILINE)
+        if compress:
+            compiled_css = re.sub(r"^ p \{", f" #{id} p " + "{", compiled_css, 0, re.MULTILINE)
```
machow, line 142: I wonder if this could be condensed into a single regex?

```python
 re.sub(r"^( p|p) \{", f" #{id} p {{", compiled_css, 0, re.MULTILINE)
```

The main difference is using `^( p|p)` to capture that the difference is whether people has a space or not.

Here's how I tested:

```python
f = lambda x, id="ABC":  re.sub(r"^( p|p) \{", f" #{id} p {{", x, 0, re.MULTILINE)

# substitutes
f(" p {abc")
f("p {abc")

# no substitute (good!)
f("group {abc")

machow (changes requested):
I left a suggestion for condensing into a single regex, but feel free to merge if you'd rather keep as is!

machow:
I'm surprised this didn't fail / require regenerating any of the snapshot tests. (I guess maybe it's because the css one used condensed = False?).

great_tables/_scss.py
rich-iannone: Yep, that is totally good. I'll make the change!

rich-iannone:
^^^ that's right. The snapshot test is only for that case.

PR #109, opened 2024-01-09
Add basic implementation of the `data_color()` method

rich-iannone:
This adds in the `data_color()` method, and it allows for adding color to data cells based on their values. The `data_color()` method automatically computes colors based on the column type (allowing you to choose a specific methodology). Here is a basic example of how this method works in practice:

```py
from great_tables import GT

(
    GT(gtcars[["mfr", "model", "msrp"]], groupname_col="mfr", rowname_col="model")
    .data_color(columns="msrp", palette=["white", "lightgreen"])
    .fmt_currency(columns="msrp", use_subunits=False)
)
```

data_color_method


machow:
We can dig into this more while pairing in a sec--I wonder if it'd be helpful to do this with the _formats.py code:

* create a folder called `_formats`
* keep data_color in `_formats/base.py` along with the original `_formats.py` code
* move helpers into `_formats/utils.py`
* move literals into `_formats/constants.py` (or something similar). E.g. the hard-coded lists etc..

machow (changes requested):
This is looking great! It seems like the last pieces left are...

* tests of `data_color`
* a few tests of big utility functions (though this could already be covered by the existing ones?!)
* change the pd.na calls in base to use `great_tables._tbl_data.is_na`

Excited for this stuff to roll out!

rich-iannone:
Thank you for reviewing this again. Will definitely get these last things finished up!

machow:
Ah, I think we need to pin `griffe==0.38.1` in our [dev dependencies of pyproject.toml](https://github.com/posit-dev/great-tables/blob/data-color-method/pyproject.toml#L56) for now. A release from yesterday changed how it represents unrecognized sections. (It should be a quick fix to update quartodoc, and once that's released we can unpin it!)

rich-iannone:
Excellent investigative work! I’ll do what you said, thanks!

Issue #110, opened 2024-01-10
Column name typo on documentation site


atseewal:
## Description

On the [Get Started](https://posit-dev.github.io/great-tables/get-started/#some-beautiful-examples) page of the documentation site, in the Air Quality table example the `Solar_R` column name is spelled as `Solar.R`. This makes the spanner label not include that column in the output for that example.

## Reproducible example

```python
from great_tables import GT, md, html
from great_tables.data import airquality

airquality_mini = airquality.head(10).assign(Year = 1973)

(
    GT(airquality_mini)
    .tab_header(
        title = "New York Air Quality Measurements",
        subtitle = "Daily measurements in New York City (May 1-10, 1973)"
    )
    .tab_spanner(
        label = "Time",
        columns = ["Year", "Month", "Day"]
    )
    .tab_spanner(
        label = "Measurement",
        columns = ["Ozone", "Solar.R", "Wind", "Temp"]
    )
    .cols_move_to_start(columns = ["Year", "Month", "Day"])
    .cols_label(
        Ozone = html("Ozone,<br>ppbV"),
        Solar_R = html("Solar R.,<br>cal/m<sup>2</sup>"),
        Wind = html("Wind,<br>mph"),
        Temp = html("Temp,<br>&deg;F")
    )
)
```

### Reprex output

![image](https://github.com/posit-dev/great-tables/assets/39170343/b8d6ce75-213f-4607-b138-6444e729a812)

## Expected result

The label spanner should include the `Solar_R` column.

### Expected result

![image](https://github.com/posit-dev/great-tables/assets/39170343/5a7be15f-f2d9-42c8-8001-53eb7719032f)

## Development environment

- **great_tables** Version: 0.1.5

machow:
Thanks for the PR, table looks fixed!

<img width="510" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/a274b60d-6af0-4b9f-a1a3-7d42892655f3">

PR #111, opened 2024-01-10
Fix tab spanner column name

atseewal:
The example tab_spanner doesn't use the right name, so the table in the docs isn't created as intended.

# Summary

Fixes the air quality example so the output table includes the correct columns in the label spanner, as intended by the code.

- Ref: #110 

# Checklist


machow (approved):
LGTM, thanks! (I didn't view the docs, but double checked and verified the name uses an underscore). @rich-iannone I'll defer to you to merge!

Issue #117, opened 2024-01-12
Give better error message when pandas DataFrame has duplicate column names


KipBalkcom-USDA:
I am trying to concat two dataframes together and create a table with great_tables. I need the dataframes joined horizontally and not stacked on top of each other. I can make a table with each individual dataframe, but not combined. However, if I change axis=1 to axis=0 in order to stack the dataframes vertically and there is no problem making the table.

This made me suspect the issue was related to the column names were the same in each dataframe. I changed the column names in the second dataframe and there is no problem creating the table with the dataframes joined horizontally.

Although this solution works, ultimately I plan to create a spanner heading in the table to denote different variables, which will have the same column names under each spanner. If the column names are already correct when I import the data, is there a way to preserve the names so I can make the table and create the correct spanner heading for each variable?

The code below illustrates the problem with the airquality data set.

```
import pandas as pd
from great_tables import GT
from great_tables.data import airquality

airquality_73 = airquality.head(10).assign(Year=1973).iloc[:, 0:2]
airquality_74 = airquality.head(10).assign(Year=1974).iloc[:, 0:2]

GT(airquality_73)
GT(airquality_74)

df = pd.concat([airquality_73, airquality_74], axis=1)

GT(df)

# change the column names for the second dataframe

airquality_74_change = (
     airquality.head(10)
     .assign(Year=1974)
     .iloc[:, 0:2]
     .rename(columns={"Ozone": "col_1", "Solar_R": "col_2"})
 )

working_df = pd.concat([airquality_73, airquality_74_change], axis=1)

GT(working_df)

```

machow:
Hey, thanks for trying Great Tables and raising this issue. I think the issue is that `df` has two columns named `Ozone`, and Great Tables cannot handle duplicate column names.

I don't think Great Tables will be able to allow duplicate column names (this is also not allowed in DataFrame libs like polars). Does renaming one of the columns work?

Let's use this issue to track Great Tables giving a more informative error message when it encounters duplicate names.

machow:
If you want to have identical column names in your final table display, you can use `GT.cols_label` to label them the same thing. This way, you can still refer to the underlying columns using their unique names, but display identical names to users.

KipBalkcom-USDA:
ok, I will change the code upstream so there are different names and then correct them when I make the table.  Thanks,.

PR #118, opened 2024-01-12
Fix validate frame

machow:
Addresses #117 by giving a hopefully diagnostic error message, when a pandas DataFrame with duplicate column names is passed: `Column names must be unique. Detected duplicate columns:\n\n <specific list>`

~TODO: we should also fail DataFrames with multi-index columns.~ (done!)

Fixes: https://github.com/posit-dev/great-tables/issues/117


rich-iannone (approved):
LGTM!

Issue #125, opened 2024-01-18
Cannot render html: 'int' object has no attribute 'text'


jgutman:
## Prework

- [X] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [X] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

Didn't see this issue already but let me know if it already exists!

## Description

It appears that data frames with integer column names error when attempting to convert to html, and don't display nicely in notebook output

## Reproducible example
In a Jupyter notebook, run the following and then try to display the objects generated by GT
```
import pandas as pd
import numpy as np

from great_tables import GT

df1 = pd.DataFrame(np.random.randint(0,100, (3, 3)))
x = GT(df1)

df2 = pd.DataFrame(np.random.randn(3, 3))
y = GT(df2)

df3 = pd.DataFrame({"x": [1,2,3]})
z = GT(df3)

df1_colnames = df1 = pd.DataFrame(np.random.randint(0,100, (3, 3)), columns = ['x', 'y', 'z'])
x_colnames = GT(df1_colnames)
```
z displays as expected, as an html table. x and y give the following error, and do not render in html
If I rename the columns, then tables that previously failed are now able to render and display successfully

```
File ~/.pyenv/versions/3.10.11/lib/python3.10/site-packages/great_tables/_text.py:52, in _process_text(x)
     50     type = "plaintext"
     51 else:
---> 52     text = x.text
     53     type = x.type
     55 if type == "from_markdown":

AttributeError: 'int' object has no attribute 'text'
```

## Expected result

All of the above should be able to display as nice html tables in a notebook.

## Development environment

- Operating System: Mac OS 
- Python 3.10.11
- great_tables 0.1.5

machow:
Hmm... I suspect the error occurs when a DataFrame has non-string names for columns (this is possible in pandas, but not in polars).

It seems like we should either...

1. explicitly raise an error, when someone tries to use `GT()` on a frame with non-string names, OR
2. allow non-strings as names, but coerce them to strings, and give a warning.

jgutman:
[...]

PR #127, opened 2024-01-19
fix: validate_frame now coerces non string column names

machow:
This PR addresses #125 by...

* coercing non-string column names in pandas to strings.
* giving a UserWarning, with details on which columns were coerced

Note that polars requires column names to be strings, so shouldn't be affected by this particular case.


rich-iannone (approved):
LGTM!

PR #128, opened 2024-01-23
Use `to_list()` method as cross-df solution

rich-iannone:
This corrects the problem with Polars DataFrames not working with the `data_color()` method due to the problem of using Pandas' `tolist()` instead of our own `to_list()` method (which works with both).


machow (approved):
LGTM!

Issue #157, opened 2024-02-02
Add function to obtain 'font stacks': themed fonts that work well across systems


rich-iannone:
It would be good to add a standalone function that fetches different font families as a list. This could be used directly with `tab_options()`'s `table_font_names` parameter, which itself expects a list of font names. The idea of a system font stack is to have font themes where fonts in the list are expected to be part of a some computing platform. This obviates the need for accessing a font from an online service, but, the cost of working entirely offline means that there will be acceptable variations platform to platform.

PR #158, opened 2024-02-02
Add the `system_fonts()` helper function

rich-iannone:
This introduces the `system_fonts()` helper function, and this allows for obtaining a list of themed fonts, the collection of which is expected to work well across computing platforms. There are 15 font stacks in total, and the function is useful within `tab_options()` (for the `table_font_names` parameter). 

Fixes: https://github.com/posit-dev/great-tables/issues/157

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -146,6 +146,7 @@ quartodoc:
         - md
         - html
         - from_column
+        - system_fonts
     - title: Table options
       desc: >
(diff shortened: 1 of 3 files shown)


machow (commented):
I left a comment on improving font stacks code

rich-iannone:
Will merge this even in it's early state (agreement from pairing). But will revisit again.

PR #162, opened 2024-02-02
Remove unimplemented options from `tab_options()`

rich-iannone:
This removes arguments from `tab_options()` that refer to unimplemented features (resulting in confusing no-ops). Associated  documentation is removed from `tab_options()` as well. Tests are added here to ensure that the generated CSS stylesheet conforms to specific uses of the `tab_options()` method.


machow (approved):
Sitting next to Rich love it ship it

Issue #156, opened 2024-02-01
Allow display of images `fmt_image()`


machow:
It would be neat to recreate [this table from twitter](https://twitter.com/LisaHornung_/status/1752981867769266231?s=20):

![image](https://github.com/posit-dev/great-tables/assets/2574498/8481c24b-5c6b-48ad-afab-1dc39a090b67)

I wonder if one thing necessary for displaying the icons would be implementing `fmt_image()`? Alternatively, it seems like we could manually use html for each icon cell?

rich-iannone:
Yeah! Having a `fmt_image()` function would make this possible. We could also use `fmt_markdown()` which should pass through HTML/SVG.

PR #163, opened 2024-02-02
Feat fmt image

machow:
This PR adds basic support for the `.fmt_image()` method. Addresses #156 

Note that it adds the following:

* metro dataset images in `great_tables/data/metro_images`
* a FmtImage dataclass, with a `.to_html()` method. I'm not sure whether this is a good or bad idea, but it was helpful for thinking about how formatters could be approached (enclosed functions, vs partials, vs classes)!

Here's the example code, run in VS Code:

```python
from great_tables import GT
from great_tables.data import metro
from importlib_resources import files

img_paths = files("great_tables")  / "data/metro_images"
GT(metro).fmt_image("lines", path = img_paths, file_pattern="metro_{}.svg")
```

image

Fixes: https://github.com/posit-dev/great-tables/issues/156


machow:
Alright, I've put the images under a single span (to match the R package). In VS Code it cuts off the width, but I wonder whether that is specific to some css rule being set in VS Code?

image

rich-iannone (approved):
Looks great!

PR #184, opened 2024-02-08
docs: superbowl blog draft

machow:
This is a rough draft of calculating superbowl squares so that you can be the superbowl square champion of you office---all that glory can be yours---all with the simple use of Polars and Great Tables.


rich-iannone (approved):
LGTM!

PR #185, opened 2024-02-08
docs: blog superbowl, add source

machow:
This PR makes a small tweak to the Super Bowl blogpost, to add a source data reference to Lee Sharpe and the nflverse


machow:
cc @rich-iannone , going to merge, since it's a small change and we were sitting next to each other!

Issue #145, opened 2024-01-31
Allow for specification of preset palettes in `data_color()`


rich-iannone:
For the `data_color()` method, users need to specify a palette as a list of colors. It would be better to have a few named color palettes on hand to make palette specification more convenient. We could add palettes from ColorBrewer and viridis, and perhaps others.

PR #186, opened 2024-02-09
Enable use of ColorBrewer palettes in `data_color()`

rich-iannone:
This lets us use ColorBrewer palette names in the `palette=` arg of `data_color()`. In addition, we can use the viridis and related palettes: `viridis`, `inferno`, `plasma`, `magma`, and `cividis`.

Fixes: https://github.com/posit-dev/great-tables/issues/145


great_tables/_data_color/base.py
```diff
@@ -456,6 +515,20 @@ def _color_name_to_hex(colors: List[str]) -> List[str]:
     return colors
 
 
+def _colorbrewer_palette_to_hex(palette: str) -> List[str]:
```
machow, line 518: It looks like these are always used after a `in` check. WDYT of us removing these, and using the underlying dictionaries directly?

e.g.

```python
COLORBREWER_NAME_TO_PALETTE[palette]
```

If you want, you could combine both dictionaries (`COLORBREWER_NAME_TO_PALETTE` and `VIRIDIS_NAME_TO_PALETTE`) into one big dictionary at the bottom of constants:

```python
ALL_PALETTES = {
    **COLORBREWER_NAME_TO_PALETTE,
    **VIRIDIS_NAME_TO_PALETTE,
}
```

machow (changes requested):
Looking good! The one small thing I wanted to check with you, is whether we could replace the palette getting functions with using the dictionaries

great_tables/_data_color/base.py
rich-iannone: Yeah, the extra functions are really redundant especially since we have the checking beforehand. I’ll also combine the dicts into one. That’s a great idea.

machow (approved):
LGTM!

PR #189, opened 2024-02-10
Replace `with_row_count` by `with_row_index` for `Polars`

jrycw:
As `with_row_count` is [deprecated](https://docs.pola.rs/py-polars/html/reference/dataframe/api/polars.DataFrame.with_row_count.html), it's recommended to use `with_row_index` instead.

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -50,6 +50,7 @@ jobs:
           logs: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
 
       - name: Netlify docs preview
+        if: ${{ !github.event.pull_request.head.repo.fork }}
         run: |
           npm install -g netlify-cli
           # push main branch to production, others to preview --
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -41,6 +41,7 @@ jobs:
       - name: Create Github Deployment
         uses: bobheadxi/deployments@v1
         id: deployment
+        if: ${{ !github.event.pull_request.head.repo.fork }}
         with:
           step: start
           token: ${{ secrets.GITHUB_TOKEN }}
@@ -67,7 +68,7 @@ jobs:
 
       - name: Update Github Deployment
         uses: bobheadxi/deployments@v1
-        if: always()
+        if: ${{ !github.event.pull_request.head.repo.fork && always() }}
         with:
           step: finish
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


machow:
Thanks for submitting this! I wonder if we need a guard for backward compatibility earlier polars versions? WDYT of using something like `f = getattr(data._tbl_data, "with_row_index", None)`, and then if the method isn't there, using `with_row_count()`?

jrycw:
> Thanks for submitting this! I wonder if we need a guard for backward compatibility with earlier polars versions? WDYT of using something like `f = getattr(data._tbl_data, "with_row_index", None)`, and then if the method isn't there, using `with_row_count()`?

It seems that the developers of Polars are diligently working towards releasing version v1.0 in the near future. Considering this, we may no longer need to maintain backward compatibility for earlier versions. As `with_row_index` is expected to be the stable API for Polars v1.0, we can proceed with this assumption.

machow:
Thanks for weighing in--it looks like `with_row_index` was added within the last month. We need to support polars from more than a month ago, just in case. If it's okay with you, I'm happy to make the change and then merge.

Thanks again for submitting the PR, it's really helpful to see this deprecation in polars, and get ahead of it

jrycw:
> Thanks for weighing in--it looks like `with_row_index` was added within the last month. We need to support versions of polars from more than a month ago, just in case. If it's okay with you, I'm happy to make the change and then merge.
> 
> Thanks again for submitting the PR, it's really helpful to see this deprecation in polars, and get ahead of it

Sounds good to me. Please proceed.

jrycw:
> Thanks for submitting this! I wonder if we need a guard for backward compatibility with earlier polars versions? WDYT of using something like `f = getattr(data._tbl_data, "with_row_index", None)`, and then if the method isn't there, using `with_row_count()`?

Perhaps using `f = getattr(data._tbl_data, "with_row_index", "with_row_count")` would be neater.

rich-iannone (approved):

machow:
@rich-iannone just a heads up, I did a tiny bit more CI tweaking to get the docs job to skip trying to push to netlify and deployment steps!

rich-iannone:
> @rich-iannone just a heads up, I did a tiny bit more CI tweaking to get the docs job to skip trying to push to netlify and deployment steps, when running on a fork!

Thanks for doing that! Should help *a lot* for community PRs that focus on improving the docs.

jrycw:
> > Thanks for submitting this! I wonder if we need a guard for backward compatibility with earlier polars versions? WDYT of using something like `f = getattr(data._tbl_data, "with_row_index", None)`, and then if the method isn't there, using `with_row_count()`?
> 
> Perhaps using `f = getattr(data._tbl_data, "with_row_index", "with_row_count")` would be neater.

The updated code looks good. However, I just realized that what I actually meant was something along the lines of `f = getattr(data, "with_row_index", data.with_row_count)` or `f = getattr(data, "with_row_index", getattr(data, "with_row_count"))`.

Issue #182, opened 2024-02-06
epic: Implement `sub_*()` functions


machow:
Functions like `sub_missing()` and `sub_zero()` allow you to replace specific values in the table. Replacing missing is especially useful, so it isn't left blank (or rendered as None).

machow:
Related to #134, we should decide on the default representation of missing values (e.g. in pandas it's currently blank, in polars it shows None).

rich-iannone:
I'm kinda leaning toward having something visible so that users really know that there are missing values that require some finalizing action. Currently I think `"None"` is better than `""`, so we could improve things by making the Pandas missing be `"None"`. We could go further and choose something like `"NA"` (like R) or `"N/A"` (like Excel) but I'm totally good with just harmonizing on `"None"`.

machow:
That makes sense! I could even see leaving the string representation of w/e the null value we received for pandas was (e.g. pd.NA, None, np.nan), so people know what they put in.

One trick for us to consider is that pandas considers np.nan a null, while [polars considers it an nan](https://docs.pola.rs/user-guide/expressions/null/#null-and-nan-values).

```python
import pandas as pd
import polars as pl
import numpy as np

pl_ser = pl.Series([np.nan])
pd_ser = pd.Series([np.nan])

pl_ser.is_null()       # False
pd_ser.isnull()        # True

pl_ser.is_nan()        # True
```

I wonder if we should add a `sub_nan` that...

* For polars, uses `is_nan()`
* For pandas, uses `is_null()` (so that sub_nan and sub_missing do the same thing for pandas)

WDYT?

ChristopherRussell:
I would vote on using the string rep of whatever nan/null value is present as a default: 'None', 'nan' 'NaT'. Avoids any surprises. 

Other than that, definitely want to be able to choose whether to substitute different values for nan and null in float columns.

rich-iannone:
I think we ought to add that `sub_nan` method to the list. That way, users can make much more targeted replacements where the distinction is important.

PR #190, opened 2024-02-11
format: add nan_rep and null_rep kwargs

ChristopherRussell:
# Summary

Add arguments to fmt_X methods to control the string representation of nan and null values. I have written test cases but I'm not quite sure how to test the formatting result and looking for advice there.

# Checklist

- [] I have added **pytest** unit tests for any new functionality.

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -121,6 +121,8 @@ def fmt_number(
     sep_mark: str = ",",
     dec_mark: str = ".",
     force_sign: bool = False,
+    null_rep: str = "None",
+    nan_rep: str | None = None,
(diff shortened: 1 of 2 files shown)


machow:
Hey, thanks for opening this PR! We should be able to support the nan and null representation, via the WIP `sub_missing` function (https://github.com/posit-dev/great-tables/issues/182). 

This is really helpful to see--since it's clear we need to account for nan handling too

machow:
Thanks for working on this PR -- I'm going to close for now, as the sub_missing method can handle replacing missing values. Note that it replaces both null and nan, but we should be able to support a custom substitutions by adding the more generic `sub()` method to GT down the road

PR #191, opened 2024-02-13
Clean up examples in API docs

rich-iannone:
This modifies the examples slightly so that the coding style is more in line with the doc site's top-level examples page (and getting started guides).


rich-iannone:
Yeah, I did a manual pass through the re-rendered pages to make sure nothing was amiss. Not much else you could really do!

PR #192, opened 2024-02-13
feat: draft allow data_color to take a palette function

machow:
This is a very rough draft of what `data_color` might look like, if the palette argument accepted an arbitrary function, mapping values -> hex colors.

Example:

```python

small_exibble = exibble[["date", "currency"]]
GT(small_exibble).data_color("currency", palette=lambda vals: ["#FFFFFF" if x > 0 else "#000000" for x in vals])
```

image

Note too important features of the current implementation:

* the function takes a list of values, and returns a list of strings
* the function must return hex strings (e.g. `#FFFFFF`)

I wonder if there's a nice way to work in polars expressions? The challenge is there's no `this` expression in polars. We could add a surrogate columns, like `_this_` etc.. This would allow people to use `pl.when(pl.col("_this_") ...).then(...)`. However, it doesn't seem ideal.....

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -173,24 +173,8 @@ def data_color(
     else:
         na_color = _html_color(colors=[na_color], alpha=alpha)[0]
 
-    # If palette is not provided, use a default palette
-    if palette is None:
-        palette = DEFAULT_PALETTE
-    elif isinstance(palette, str):
-        # Check if the `palette` value refers to a ColorBrewer or viridis palette
-        # and, if it is, then convert it to a list of hexadecimal color values; otherwise,
-        # convert it to a list (this assumes that the value is a single color)
-        if palette in ALL_PALETTES:
-            palette = ALL_PALETTES[palette]
-        else:
-            palette = [palette]
-
-    # Reverse the palette if `reverse` is set to `True`
-    if reverse:
-        palette = palette[::-1]
-
-    # Standardize values in `palette` to hexadecimal color values
-    palette = _html_color(colors=palette, alpha=alpha)
+    if isinstance(palette, (list, str)):
+        palette = _process_palette(palette, reverse, alpha)
 
     # Set a flag to indicate whether or not the domain should be calculated automatically
     if domain is None:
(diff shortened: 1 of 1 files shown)


ChristopherRussell:
Re: what else is out there

colorcet (seems to also be available via matplotlib. claims to provide better best colormaps for continuous data) 
https://colorcet.holoviz.org/user_guide/index.html

ChristopherRussell:
I think it would be really great to be able to do things like this:

https://matplotlib.org/stable/users/explain/colors/colormapnorms.html

One could do that by passing a callable that handles both normalization (mapping values to [0,1]) and the color map. Maybe we want to allow that, but in general it is nice if these things are separate, so you can outsource one or the other to pre-existing collections of maps / normalization tools!

ChristopherRussell:
This is how I would use matplotlib normalizers + a cmap to get hex values:

from typing import Callable

import matplotlib.colors as mcolors
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd

```

def _color_series(
    values: pd.Series,
    colormap: plt.cm,
    norm: mcolors.Normalize | Callable | None = None,
    na_color: str | None = None,
) -> pd.Series:
    """
    Color a Series of numeric values using a Matplotlib colormap and normalization.

    Parameters:
    values : pd.Series
        The Series to color.
    colormap : plt.cm
        The Matplotlib colormap to use.
    norm : mcolors.Normalize | Callable | None
        A matplotlib normalization object (e.g., Normalize, LogNorm) or a function that returns
        normalized valeus within the range [0, 1]. If None, defaults to linear normalization
        between the minimum and maximum values.
    na_color : str, optional
        The color to use for NaN values.

    Returns:
    pd.Series
        A new Series with hex color strings.
    """
    # Normalize the data
    if norm is None:  # Defaults to linear normalization between min and max
        norm = mcolors.Normalize(values.min(), values.max())
    normalized = norm(values.values)
    colors = colormap(normalized)

    # Convert RGBA colors to hex, handling NaN values
    hex_colors = [mcolors.to_hex(color) if not np.isnan(color[0]) else na_color for color in colors]

    return pd.Series(hex_colors, index=values.index)

```

I have also attached some example use cases to show off where they can be useful.

[color_series.pdf](https://github.com/posit-dev/great-tables/files/14546251/color_series.pdf)

PR #197, opened 2024-02-14
Docs table themes

machow:
This PR adds a page on table theming options to the Get Started guide. It's still very rough!

https://pr-197--gt-python.netlify.app/get-started/table-theme-options


docs/get-started/table-theme-options.qmd
```diff
@@ -0,0 +1,203 @@
...
+* column_labels, row_group, stub, stub_row
+* table_body
+:::
+
+## Styling borders
+
+Many table parts support customizing border colors and style.
+This is shown below for column labels.
+
+```{python}
```
machow, line 85: I couldn't see the result of setting these options. Is there something else I need to set?!

docs/get-started/table-theme-options.qmd
```diff
@@ -0,0 +1,203 @@
...
+)
+```
+
+## A basic theme
+
+Based on the sections above, we can design an overall theme for a table.
+
+This requires setting a decent number of options, but makes a big difference when presenting a table! Below is a table with a simple, blue theme. (The code is hidden by default, but can be expanded to see all the options set).
+
+```{python}
```
machow, line 155: Note that I essentially derived this from the `opt_stylize` R code.

docs/get-started/table-theme-options.qmd
```diff
@@ -0,0 +1,203 @@
...
+        # misc ----
+        #row_striping_background_color="#F4F4F4"
+    )
+
+)
+```
+
+## Composite options
+
+TODO: which options are handled in a special way? (e.g. `row_group_as_column = True`)
```
machow, line 203: Any idea which options do a set of things, rather than a single thing? (Or which of those options would be good to illustrate---either here or in their own section).

docs/get-started/table-theme-options.qmd
rich-iannone: Use this instead:

```py
gt_ex.tab_options(
    heading_border_bottom_color="blue",
    heading_border_bottom_style="solid"
)
```

The reason the other settings may not work is because the shared border (bottom of header, and top of column labels) has a rule where the larger sized part wins.

rich-iannone: Like, if you use:

```python
gt_ex.tab_options(
    column_labels_border_top_color="blue",
    column_labels_border_top_style="solid",
    column_labels_border_top_width="5px",
)
```

Then you'll see the blue border because the `5px` size is beating out the size of the `heading_border_bottom...` border.

docs/get-started/table-theme-options.qmd
rich-iannone: container_width and container_height (and container_overflow_x|y) do more complex things which are worth illustrating (especially if space is at a premium for the table on a page).

docs/get-started/table-theme-options.qmd
machow: Ah, thanks! I might use this, and then mention the interaction of borders, if that's okay!

rich-iannone (approved):
So far this is looking good!

rich-iannone (approved):
Looks great!

Issue #165, opened 2024-02-06
Implement table themes


machow:
https://gt.rstudio.com/reference/opt_stylize.html

We could put themes in a submodule (e.g. `great_tables.theme`)

PR #198, opened 2024-02-14
Add the `opt_stylize()` method

rich-iannone:
This adds in the `opt_stylize()` method. With it you can quickly style a table with a carefully curated set of background colors, line colors, and line styles. There are six styles to choose from and they largely vary in the extent of coloring applied to different table locations. Some have table borders applied, some apply darker colors to the table stub and summary sections, and, some even have vertical lines. In addition to choosing a `style` preset, there are six `color` variations that each use a range of five color tints. Each of the color tints have been fine-tuned to maximize the contrast between text and its background. There are 36 combinations of `style` and `color` to choose from.

Fixes: https://github.com/posit-dev/great-tables/issues/165


great_tables/_options.py
```diff
@@ -969,3 +969,673 @@ def opt_all_caps(
...
+    largely vary in the extent of coloring applied to different table locations. Some have table
+    borders applied, some apply darker colors to the table stub and summary sections, and, some even
+    have vertical lines. In addition to choosing a `style` preset, there are six `color` variations
+    that each use a range of five color tints. Each of the color tints have been fine-tuned to
+    maximize the contrast between text and its background. There are 36 combinations of `style` and
+    `color` to choose from.
+
+    Parameters
+    ----------
+    style : int
```
machow, line 989: We don't need to write annotations in the parameters section of a docstring. griffe + quartodoc will add the annotations from the function signature.

great_tables/_options.py
```diff
@@ -969,3 +969,673 @@ def opt_all_caps(
...
+    are applied to the table. The default style and color can be modified by providing the `style=`
+    and `color=` arguments.
+
+    ```{python}
+    gt_tbl.opt_stylize(style=2, color="green")
+    ```
+    """
+
+    # Validate the `style` and `color` arguments
+    if style not in [1, 2, 3, 4, 5, 6]:
```
machow, line 1041: Pairing, I asked why we use numbers and not names for styles. Rich and I copied out all the 6 styles side-by-side, and I just want to +1 that it's very hard to think of names that would distinguish these things.

great_tables/_options.py
```diff
@@ -969,3 +969,673 @@ def opt_all_caps(
...
+
+    # Get the style parameters based on the `style` and `color` arguments
+    params = _dict_styles_colors_params[f"{color}-{style}"]
+
+    # Apply the style parameters to the table using the `tab_options()` method
+    # TODO: the commented out lines are for future use when:
+    #   (1) row striping is implemented
+    #   (2) summary rows are implemented
+    #   (3) grand summary rows are implemented
+    res = tab_options(
```
machow, line 1053: I wonder if it would be useful to create a mapper, that takes the style parameters in its init, and can output the tab_options values as a dict.

e.g.

```
from dataclasses import dataclass, fields
from typing import ClassVar

@dataclass
class StyleMapper:
    table_hlines_color: str
    column_labels_background_color: str

    mappings: ClassVar[dict[str, list[str]]] = {
        "table_hlines_color": ["table_border_top_color", "table_border_bottom_color"],
        "column_labels_background_color": ["column_labels_background_color"]
    }

    def map_entry(self, name: str):
        return {k: getattr(self, name) for k in self.mappings[name]}

    def map_all(self):
        items = {}
        for field in fields(self):
            items.update(self.map_entry(field.name) )
        return items

StyleMapper("ABC", "DEF").map_all()
```

Then you could pass the result straight to tab_options, e.g.

```python
# inside opt_stylize
self.tab_options(**StyleMapper(...).map_all())
```

What is helpful about this for me is that it signals that:

* StyleMapper is a very simple one entry thing to multiple entries in a dict
* opt_stylize is a thin wrapper around tab_options

machow: Feel free to punt this to an issue and just link here

machow (approved):
I left a suggestion on potentially creating a class to do the mapping, but feel free to merge and just create an issue if we want to punt!

great_tables/_options.py
rich-iannone: Was able to integrate this nicely. Thanks so much for working through this with me!

Issue #194, opened 2024-02-14
Standardizing code style for checking container emptiness


jrycw:
To determine whether a container-like structure is empty in Python, we typically use `if not container` or `if len(container)==0`. Both approaches are acceptable in my view. However, it seems redundant to use `if not len(container)` in the code, as seen [here](https://github.com/posit-dev/great-tables/blob/bfc9a5cf99fc0ff98f7aeb6aefa5863bdaa8e7db/great_tables/_spanners.py#L257). It would be better to simplify the code using one of the common patterns I described.

I suggest standardizing the code style for this type of check. We can use both `if not container` and `if len(container)==0`, but let's aim to replace `if not len(container)` with one of these two options.

If we can agree on this approach, I'm ready to submit a PR to address it.

machow:
Hey, thanks for surfacing. I feel okay with how it's written, but since the next elif clause uses `len(sel_after) > 1`, I could see us changing the original check to `if len(sel_after) == 0`.

If you want to submit a PR for that, happy to merge!

jrycw:
Thanks for the response. It seems like this issue pertains more to a preference in code style. Allow me to delve a bit deeper into my perspective.

When I encounter the code `if not container`, it's immediately clear to me that the if-statement is evaluating the container's truthiness. Conversely, when I come across if `len(container) == 0`, I understand that the if-statement is checking the container's length using the `len()` function, which relies on the return value of the `__len__()` dunder method of the container's class.

However, when I encounter `if not len(container)`, I find myself needing to first consider the length of the container and then deduce the truth value based on that number. This additional step feels a bit like a detour to me.

PR #199, opened 2024-02-15
Standardizing code style for checking container emptiness

jrycw:
This PR aims to standardize the code style for checking container emptiness. Relevant discussion can be found in issue #194.

Additionally, I noticed the pattern `if len(container)`, and initially planned to replace it with `if container`, which feels more intuitive to me than `if len(container) != 0`.  However, it appears that this modification would result in breaking too many tests, so I will leave it unchanged for now.

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -728,7 +728,7 @@ class Spanners(_Sequence[SpannerInfo]):
         Spanners whose columns do not overlap are put on the same level.
         """
 
-        if not len(self):
+        if len(self) == 0:
             return 0
 
         overlapping_levels = [
--- a/great_tables/_locations.py
+++ b/great_tables/_locations.py
@@ -214,7 +214,7 @@ def resolve_cols_i(
 
         # TODO: special handling of "stub()"
         if isinstance(expr, list) and "stub()" in expr:
-            if stub_var:
+            if len(stub_var):
                 return [(stub_var[0], 1)]
 
             return []
@@ -248,7 +248,7 @@ def resolve_cols_i(
             # always excluded but in certain cases (i.e., `rows_add()`)
             # we may want to include this column
             _group_vars = data._boxhead.vars_from_type(ColInfoTypeEnum.row_group)
-            group_var = _group_vars[0] if _group_vars else None
+            group_var = _group_vars[0] if len(_group_vars) else None
         else:
             group_var = None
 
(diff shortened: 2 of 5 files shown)


machow:
Thanks for submitting! It's helpful for me to have to think explicitly about these choices!

> Additionally, I noticed the pattern if len(container), and initially planned to replace it with if container, which feels more intuitive to me than if len(container) != 0. However, it appears that this modification would result in breaking too many tests, so I will leave it unchanged for now.

Do you mind pointing me to a place where this happened? I'm guessing it was for a custom class?

jrycw:
> Thanks for submitting! It's helpful for me to have to think explicitly about these choices!
> 
> > Additionally, I noticed the pattern if len(container), and initially planned to replace it with if container, which feels more intuitive to me than if len(container) != 0. However, it appears that this modification would result in breaking too many tests, so I will leave it unchanged for now.
> 
> Do you mind pointing me to a place where this happened? I'm guessing it was for a custom class?

There should be four places where the pattern `if len(container)` is used. You can review the final revert commit [here](https://github.com/posit-dev/great-tables/pull/199/commits/a5398228d5e9f280edbdeb26fe05b92774818f71). 

Meanwhile, you can check the test fail log, which should be available somewhere like [here](https://github.com/posit-dev/great-tables/actions/runs/7914142787/job/21603274713). It should be something like:
``` python
ERROR tests/test_gt.py::test_gt_replace - ValueError: The truth value of a Index is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
ERROR tests/test_gt.py::test_gt_object_prerender - ValueError: The truth value of a Index is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
ERROR tests/test_gt.py::test_gt_table_render - ValueError: The truth value of a Index is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

machow:
Ah thanks for pointing to the code! I think this is the issue discussed in this [SO post](https://stackoverflow.com/a/9381545/1144523) (they focus on numpy, but same concept; arrays in data libraries raise errors on `__bool__`).

jrycw:
Thanks for sharing the details. Do you think we should stick with the current approach or switch to `if len(container) != 0`? Or do you have any other ideas?

machow:
I've been thinking through these kinds of checks, and how they work with data tools. If it's okay with you, I feel okay with leaving the original code as is. Thanks for flagging the need for consistency, and digging into this code. It was really helpful to have to read about these checks, PEP 8, and how this works with DataFrames/Series, etc..

It seems okay to leave as is, since we're a fairly data centric library, and I suspect I adopted these practices in part because they work with data arrays etc..

Issue #201, opened 2024-02-16
Issue of reordering


jrycw:
I'm currently following the examples provided in [countrypops](https://github.com/posit-dev/great-tables/blob/main/docs/examples/index.qmd#L73-L107), but I'm encountering some unexpected results. One immediate observation is that "New Zealand" should be classified under the region of "Australasia" according to the `countries dict`.
![fig_orig](https://github.com/posit-dev/great-tables/assets/67060418/111776d9-7dfc-40b1-87a0-c6dc07270e24)

I attempted to address this issue using `pandas`, and the results appeared to be more accurate this time.
``` python
from great_tables import GT
from great_tables.data import countrypops


countries = {
    "Australasia": ["AU", "NZ"],
    "Melanesia": ["NC", "PG", "SB", "VU"],
    "Micronesia": ["FM", "GU", "KI", "MH", "MP", "NR", "PW"],
    "Polynesia": ["PF", "WS", "TO", "TV"],
}

region_to_country = {
    region: country for country, regions in countries.items() for region in regions
}


def rename_cols(df_):
    df_.columns.name = ""
    df_.columns = df_.columns.astype(str)
    return df_


wide_pops_pd = (
    countrypops.query(
        "country_code_2.isin(@region_to_country.keys()) and year.isin([2000, 2010, 2020])"
    )
    .assign(region=lambda df_: df_.country_code_2.replace(region_to_country))
    .pivot(
        index=[
            "region",
            "country_name",
        ],
        columns=["year"],
        values=["population"],
    )
    .droplevel(level=0, axis=1)
    .pipe(rename_cols)
    .reset_index()
    .sort_values(["region", "2020"], ascending=[True, False])
)

(
    GT(wide_pops_pd, rowname_col="country_name", groupname_col="region")
    .tab_header(title="Populations of Oceania's Countries in 2000, 2010, and 2020")
    .tab_spanner(label="Total Population", columns=wide_pops_pd.columns.tolist())
    .fmt_integer()
)
```
![fig_modified](https://github.com/posit-dev/great-tables/assets/67060418/1e32f285-de2f-42a6-aead-8f04d0940d37)

It appears that the problem may stem from the code that manipulates `Polars`, but I'm having difficulty tracing the exact source of the issue.

I suspect there may be some flawed logic in the codebase when recording the rows from `Polars`.

jrycw:
[...]

PR #202, opened 2024-02-16
fix: iterate over sorted rows for rendering

machow:
This PR addresses #201, by ensuring we iterate over rows sorted by grouping when rendering.

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -5,6 +5,7 @@ from typing import List, Any, cast
 from htmltools import tags, HTML, css, TagList
 from itertools import groupby, chain
 from ._text import StringBuilder, _process_text, _process_text_id
+from .utils_render_common import get_row_reorder_df
 
 
 def create_heading_component_h(data: GTData) -> StringBuilder:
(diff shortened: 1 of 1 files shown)


machow:
I can pick this one back up! (changed the assignee)

machow:
There is an extra entry in group B in our test. From pairing w/ Rich it sounds like maybe an off by 1 error

image

machow:
Okay, I've gotten the snapshot to produce the correct output. I'm not sure how `.indices_map()` works in the original R program, but I tweaked it to return 1 entry per row, so we can iterate over it directly when rendering. In general, I think it'd be useful for us to trace / sequence diagram when things get reordered. The internal logic is clear, but it seems a bit tricky to know when/where all reorderings have happened!

rich-iannone:
> Okay, I've gotten the snapshot to produce the correct output. I'm not sure how `.indices_map()` works in the original R program, but I tweaked it to return 1 entry per row, so we can iterate over it directly when rendering. In general, I think it'd be useful for us to trace / sequence diagram when things get reordered. The internal logic is clear, but it seems a bit tricky to know when/where all reorderings have happened!

It *is* tricky. Was that way in the R implementation as well. But I found peace there after lots of patches and extensive testing (it became a solved problem). Definitely would be good for us to diagram/explain the internals a bit!

rich-iannone (approved):
Looks great, thank you!

jrycw:
Nice work!

Issue #204, opened 2024-02-18
`data_color()` is not compatible with mizani 0.9.3 when missing values are present


mattheaphy:
The `data_color()` method relies on `mizani.palettes.gradient_n_pal()` to create color palette functions: https://github.com/posit-dev/great-tables/blob/c288893ccfc6a4c44c42354f43e2043eeef234fd/great_tables/_data_color/base.py#L252

In older versions of mizani, a function produced by `gradient_n_pal()` returns `np.nan` when fed missing values. The current version of mizani returns `None` whenever it sees a missing value. 

`data_color()` handles the latter situation, but not the former: https://github.com/posit-dev/great-tables/blob/c288893ccfc6a4c44c42354f43e2043eeef234fd/great_tables/_data_color/base.py#L258

It would be helpful to adjust the prior line so that `na_color` also replaces `np.nan`. For example:

`color_vals = [na_color if x is None or x != x else x for x in color_vals]`

Alternatively, the minimum required version of mizani could be increased to dodge this issue.

mattheaphy:
Thanks for the update!

PR #205, opened 2024-02-19
Ensure that np.nan values are replaced with 'na_color=` vals

rich-iannone:
The `data_color()` method relies on `mizani.palettes.gradient_n_pal()` to create color palette functions. However, In older versions of mizani, a function produced by `gradient_n_pal()` returns `np.nan` when given missing values. The current version of mizani returns `None` on missing values.

This PR provides wider compatibility with mizani by ensuring both behaviors are handled. The only change required is to the list comprehension here

```python
color_vals = [na_color if x is None else x for x in color_vals] 
```

becoming:

```python
color_vals = [na_color if x is None or x != x else x for x in color_vals]
```

Fixes: https://github.com/posit-dev/great-tables/issues/204

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -255,7 +255,7 @@ def data_color(
         color_vals = color_scale_fn(scaled_vals)
 
         # Replace 'None' and 'np.nan' values in `color_vals` with the `na_color=` color
-        color_vals = [na_color if x is None or x != x else x for x in color_vals]
+        color_vals = [na_color if is_na(data_table, x) else x for x in color_vals]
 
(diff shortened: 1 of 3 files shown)


machow:
From pairing, I can pick this one up and run with it

machow:
@rich-iannone I can't assign you as a reviewer, but just wanted to double check that this looks okay to merge!

rich-iannone:
That's weird. Will review now!

rich-iannone:
Oh, I think it's because I opened the PR and GH doesn't allow 'self-review' :\

rich-iannone:
So, you're good to merge!

PR #207, opened 2024-02-22
Remove type annotations from docstrings

rich-iannone:
This removes all type annotations from docstrings. They aren't needed since griffe generates them automatically.


machow:
Yo, thanks for all this cleanup work!

machow (approved):

Issue #203, opened 2024-02-17
ability to save table to a png


bkxndata:
Amazing work on this project -- I'm highly appreciative

I'm curious if there is any way to save a table to a .png (or any sort of image).  I'm looking for something similar to the gtsave package in gt.  I know there's a way to render the html code, but curious if I'm missing a way to export to image.

Thank you in advance

rich-iannone:
Use this: https://github.com/vgalin/html2image

rich-iannone:
From Discord user `dersu` in great_tables server, this code can be adapted:

```python
from html2image import Html2Image
from IPython.core.display import HTML

html_file_path = r"C:\Scripts\test.html"

# Get the HTML content from the displayed output
html_content = HTML(gt._repr_html_()).data

# Save the HTML content to a file
with open(html_file_path, "w") as f:
    f.write(html_content)

# Instantiate Html2Image
hti = Html2Image()

# Screenshot an HTML file
hti.screenshot(
    html_file=html_file_path, save_as="test.png"
)
```

rich-iannone:
^^^ tried this approach. Was a bit disappointed with the result. You can’t target the table element, the image is at a low resolution, and it’s uncropped. 

Another way is to use Selenium, which provides more access to the Chrome Webdriver. And you can target the table element too.

Yasin197:
> Amazing work on this project -- I'm highly appreciative
> 
> I'm curious if there is any way to save a table to a .png (or any sort of image).  I'm looking for something similar to the gtsave package in gt.  I know there's a way to render the html code, but curious if I'm missing a way to export to image.
> 
> Thank you in advance
> 
> 

Hi, do you have an example where your concert the GT table into HTML?

Issue #177, opened 2024-02-06
Implement `gtsave()`


machow:
Could be `GT.save()` or `GT.to_image()`, etc..

PR #208, opened 2024-02-22
Add the `.save()` and `.as_raw_html()` methods

rich-iannone:
This adds the `.save()` and `.as_raw_html()` methods. The first allows for image export of a GT table. It uses Selenium and the Chrome webdriver to take the screenshot, leaving cropping of the table to the PIL (via pillow) library. The user is able to control the detail of the resulting image via the `zoom=` argument; the `expand=` argument allows for a looser cropping of the table image for a better presentation. It's possible to save to a PNG, PDF, TIFF, BMP, or GIF file; this is detected from the file extension of the provided output filename.

The `.as_raw_html()` method facilitates the export of an HTML fragment for the table. This is useful for inspecting the table HTML and also for embedding the self-contained table into other environments.

Fixes: https://github.com/posit-dev/great-tables/issues/177
Fixes: https://github.com/posit-dev/great-tables/issues/203


great_tables/gt.py
```diff
@@ -403,17 +404,150 @@ def _get_column_of_values(gt: GT, column_name: str, context: str) -> List[str]:
...
+    -------
+    str
+        An HTML fragment containing a table.
     """
     gt_built = gt._build_data(context="html")
     html_table = gt_built._render_as_html()
     return html_table
+
+
+def gtsave(
```
machow, line 429: Can we move this logic out of `gt.py`? It seems helpful to restrict this file to the creation of the GT class (which is mostly a wrapper around other things)

pyproject.toml
```diff
@@ -42,10 +42,12 @@ dependencies = [
     "typing_extensions>=3.10.0.0",
     "packaging>=20.9",
     "pandas>=1.4.2",
+    "Pillow>=10.2.0",
     "numpy>=1.22.4",
     "Babel>=2.13.1",
     "mizani>=0.9.3",
-    "webcolors",
+    "selenium>=4.18.1",
```
machow, line 49: I wonder if it's worth making this (and Pillow) an optional dependency? We could make a set of optional dependencies called `web` or similar, and queue users to `pip install 'great_tables[html]'` or `pip install 'great_tables[web]'`

(I'm def open to us keeping it as required!)

machow (changes requested):
This is looking great! I left some questions about whether `as_raw_html` might be duplicating render, and whether we might make `gtsave()` a method.

(I did try running the code, and wanted to check on these things first!)

great_tables/gt.py
```diff
@@ -403,17 +404,150 @@ def _get_column_of_values(gt: GT, column_name: str, context: str) -> List[str]:
     return cell_values
 
 
-def _as_raw_html(gt: GT) -> str:
+def as_raw_html(gt: GT) -> str:
```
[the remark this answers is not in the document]
rich-iannone: It is, for now. I think we'd want to have a scoped method just for HTML so we could add useful arguments for how the HTML should be extracted/processed. For instance, an `inline_css=` argument is something we do want (for HTML email, and other HTML embedding elsewhere). And there are other possibilities for HTML table composition that would work better in `as_raw_html()` compared to the more general `render()` method.

great_tables/gt.py
rich-iannone: Putting all of this into `_export.py`.

great_tables/gt.py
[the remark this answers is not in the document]
machow: Ah, that makes sense! Looking at the R reference, it looks like render tries to produce a nice output for display, but as you mentioned, the `as_*()` functions have some extra options. I can definitely see the value of having more granular `as_*()` funcs.

It seems like...

* They'd be useful as functions (e.g. `GT.as_html()`). But we could attach them similar to formatters?
* It makes a lot of sense why there is some duplication there, thanks for walking me through it!

pyproject.toml
rich-iannone: I moved those deps into `extra`. Will document in README about the different pip install options.

great_tables/_gt_data.py
```diff
@@ -89,6 +90,16 @@ def from_data(
             _options=Options(),
         )
 
+        if id is not None:
```
machow, line 93: Could this happen directly above (on line 90), when the Options object is initially created?

great_tables/_gt_data.py
```diff
@@ -1136,6 +1144,9 @@ class Options:
     quarto_disable_processing: OptionsInfo = OptionsInfo(False, "quarto", "logical", False)
     quarto_use_bootstrap: OptionsInfo = OptionsInfo(False, "quarto", "logical", False)
 
+    def _replace(self, **kwargs: OptionsInfo) -> Self:
```
machow, line 1147: (See comment above.) I think we can remove this, since the dataclasses.replace function works okay here.

machow (changes requested):

great_tables/_gt_data.py
rich-iannone: Now done, thanks for pairing on this.

great_tables/_export.py
```diff
@@ -0,0 +1,246 @@
...
+        path.
+
+    Details
+    -------
+    We create the output file based on the HTML version of the table. With the filename extension
+    .png, we get a PNG image file. Using .pdf will yield a PDF document. The other file extensions
+    supported are .gif, .bmp, .tif, and .tiff. The file extension is used to determine the format of
+    the output file.
+
+    This process is facilitated by two libraries:
```
machow, line 90:
```suggestion
    Details
    -------
    We create the output file based on the HTML version of the table.

    This process is facilitated by two libraries:
```


I think this is okay to remove, since the file extension determining the format is mentioned in the initial description: 

> Finally, and the resulting image is saved to the specified file path in the format specified (via the file extension).

great_tables/_export.py
```diff
@@ -0,0 +1,246 @@
...
+        "gif": "gif",
+        "bmp": "bmp",
+        "tif": "tiff",
+        "tiff": "tiff",
+        "pdf": "pdf",
+    }
+
+    # Get the PIL format from the file extension
+    pil_format = pil_formats[file_extension]
+
```
machow, line 162:
```suggestion
    pil_format = file_extension
```

WDYT of us just passing the file_extension directly into PIL, and telling users that's what happens. This way, they can use any of the PIL output types, and take advantage of the pillow docs.

For example, we could say it gets passed straight to the `PIL.Image.save()` method.

great_tables/_utils.py
```diff
@@ -1,9 +1,24 @@
...
 
+def _try_import(name: str, pip_install_line: Optional[str] = None) -> ModuleType:
+    try:
+        return importlib.import_module(name)
+    except ImportError:
+        if pip_install_line is not None:
+            raise ImportError(
+                f"""Module {name} not found
+                - run `{pip_install_line}` to install."""
+            )
```
machow, line 17:
```suggestion
            raise ImportError(
                f"Module {name} not found. Run the following to install."
                f"\n\n`{pip_install_line}`"
            ) from None
```

The spacing of this multiline string deeply indents the `- run...` part.

image

I added a `from None`, so the stack trace isn't so deep!

great_tables/_export.py
```diff
@@ -0,0 +1,246 @@
...
+    str
+        An HTML fragment containing a table.
+    """
+    return self._build_data(context="html")._render_as_html()
+
+
+def save(
+    self: GTData,
+    filename: str,
+    path: Optional[str] = None,
```
machow, line 32: Can we match tools like polars and pandas, and...

* remove the path paramter
* (polars uses the name "file" for filename, and slashes are interpreted as directories; pandas uses the name path_or_buf, which seems a bit gratuitous!)

It seems like the full file path can just be passed straight into PIL. See PIL's [.save() method docs](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save)

great_tables/_export.py
```diff
@@ -0,0 +1,246 @@
...
+    pil_format = pil_formats[file_extension]
+
+    # Get the HTML content from the displayed output
+    html_content = as_raw_html(self=self)
+
+    # Create a temp directory to store the HTML file
+    temp_dir = tempfile.mkdtemp()
+
+    # Create a temp file to store the HTML file; use the .html file extension
+    temp_file = tempfile.mkstemp(dir=temp_dir, suffix=".html")
```
machow, line 170: We have to manually delete this file when we're done with it (unfortunately this detail is buried at the bottom of `tempfile.mkstemp()` 😭 )

You can use a context handler to do it automatically. This is important, because if the code errors, the context handler will still make sure to delete the file...

```
with tempfile.TemporaryFile() as fp:
    fp.write(b'Hello world!')
    fp.seek(0)
    fp.read()
```

Note that the file gets deleted after the `with` block, so everything using the file needs to go in there.

machow (changes requested):
This is looking great! I pulled the PR locally and tested out saving images. I've left some comments based on:

* **small docstring improvements** based on where I looked for info (e.g. it took me a sec to find what I should write for the extension, so I left a suggestion to move it up into parameters)
* **technical pieces** (tempfile needs to be deleted after use; shortening try_import stack trace)
* **parameters:** a suggestion to remove the path argument, using pandas, polars, and PIL for inspiration (they just use a single name, which can include folders; no need for extra handling internally.)

great_tables/_export.py
rich-iannone: I like it! Was actually unaware PIL could do this.

machow (approved):
This looks great--thanks for knocking out all the changes!

Since many of the commits are simply "Update file.py", WDYT of use doing a squash merge?

PR #209, opened 2024-02-22
Add the `opt_table_outline()` method

rich-iannone:
This PR adds another `opt_*()` method, this time the `opt_table_outline()` method. It lets you easily draw a line around the entire table, setting several options inside `tab_options()` all at once.


machow (approved):
LGTM!

Issue #212, opened 2024-02-22
data_color() raises division by zero if single value in column


machow:
Here's a reproducible example for three cases where errors get raised:

```python
import pandas as pd
from great_tables import GT

# single value -- division by zero
df1 = pd.DataFrame({"x": [1], "y": [3]})
df2 = pd.DataFrame({"x": [1, None], "y": [3, 4]})

# all missing -- min error
df3 = pd.DataFrame({"x": [None], "y": [3]})

GT(df1).data_color("x")
GT(df2).data_color("x")
GT(df3).data_color("x")
```

PR #213, opened 2024-02-23
Handle `.data_color()` edge cases with single val columns / all missing values

rich-iannone:
As discovered in https://github.com/posit-dev/great-tables/issues/212, there are a few pernicious edge cases of data frame columns being incompatible with the color-scaling methodology of `.data_color()`. We shouldn't error or disallow these types of cases from working with that method, we should handle them with some logic.

Here are the cases and how they will be dealt with.

(1) Single value in column --> use first color from palette for the single row

```python
df = pd.DataFrame({"x": [1], "y": [3]})
GT(df).data_color("x", palette=["green", "blue"], na_color="red")
```

(2) Single value , multiple rows (rest of the rows are missing) --> use first color from palette and apply `na_color=` to the missing rows

```python
df = pd.DataFrame({"x": [1, None], "y": [3, 4]})
GT(df).data_color("x", palette=["green", "blue"], na_color="red")
```

(3) Multiple rows in a column but all are missing --> apply `na_color=` to all rows

```python
df = pd.DataFrame({"x": [None, None], "y": [3, 6]})
GT(df).data_color("x", palette=["green", "blue"], na_color="red")
```

(4) Single missing value from a single row --> apply `na_color=` to the row

```python
df = pd.DataFrame({"x": [None], "y": [3]})
GT(df).data_color("x", palette=["green", "blue"], na_color="red")
```

Fixes: https://github.com/posit-dev/great-tables/issues/212

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -212,6 +212,8 @@ def data_color(
     else:
         columns_resolved = columns
 
+    gt_obj = self
+
     # For each column targeted, get the data values as a new list object
     for col in columns_resolved:
         column_vals = data_table[col].to_list()
@@ -219,7 +221,7 @@ def data_color(
         # If there only a single value in `column_vals`, and it is not missing,
         # then use the first value from the palette
         if len(column_vals) == 1 and not is_na(data_table, column_vals[0]):
-            self = self.tab_style(
+            gt_obj = gt_obj.tab_style(
                 style=fill(color=palette[0]), locations=body(columns=col, rows=[0])
             )
 
@@ -237,7 +239,7 @@ def data_color(
             )
 
             # Apply the first color from the palette to the non-missing value
-            self = self.tab_style(
+            gt_obj = gt_obj.tab_style(
                 style=fill(color=palette[0]), locations=body(columns=col, rows=[non_missing_index])
             )
 
@@ -245,7 +247,7 @@ def data_color(
             missing_indices = [i for i, x in enumerate(column_vals) if is_na(data_table, x)]
 
             # Apply the `na_color=` color to all missing values
-            self = self.tab_style(
+            gt_obj = gt_obj.tab_style(
                 style=fill(color=na_color), locations=body(columns=col, rows=missing_indices)
             )
 
@@ -254,7 +256,7 @@ def data_color(
         # If the entire column contains NA values, then apply the `na_color=` color to the
         # entire column and then move on to the next column
         if all(is_na(data_table, x) for x in column_vals):
-            self = self.tab_style(style=fill(color=na_color), locations=body(columns=col))
+            gt_obj = gt_obj.tab_style(style=fill(color=na_color), locations=body(columns=col))
             continue
 
         # Filter out NA values from `column_vals`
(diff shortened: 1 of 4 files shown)


great_tables/_data_color/base.py
```diff
@@ -216,9 +216,57 @@ def data_color(
     for col in columns_resolved:
         column_vals = data_table[col].to_list()
 
+        # If there only a single value in `column_vals`, and it is not missing,
+        # then use the first value from the palette
+        if len(column_vals) == 1 and not is_na(data_table, column_vals[0]):
+            self = self.tab_style(
```
machow, line 222: `self` is sort of viewed in a special way, since it stands mentally for the object being operated on. It feels a bit funky to re-assign it. Maybe we could call it gt_obj or something?

I think a common pattern is this:

```python

if some_thing:
    gt_obj = self.some_action()
else:
    gt_obj = self
```

This way, people have a strong queue that a new object might be returned

machow (changes requested):
This PR hits some key cases, but I think we might need to go into some of the deeper calculations to get at the root cause?

```python
df1 = pd.DataFrame({"x": [1, 1], "y": [2, 3]})
GT(df1).data_color("x")
```

It seems like...

* the divide by zero error is a range calculation issue. Maybe we could use a try/except to catch division by zero, or handle the logic inside that function so division by zero doesn't occur?
* the min error seems similar (as it occurs as part of the calculation of the range).

**edit:** What if---when the range is 0 (all same value)---you scaled the values to some arbitrary value? e.g. 0, 1, or .5. It seems like a core issue is that what the scale value should be is ambiguous. Similarly, when all the values are None, it doesn't matter what you set the range to, because everything will remain None.

great_tables/_data_color/base.py
rich-iannone: Now done. Thanks!

rich-iannone:
> This PR hits some key cases, but I think we might need to go into some of the deeper calculations to get at the root cause?
> 
> ```python
> df1 = pd.DataFrame({"x": [1, 1], "y": [2, 3]})
> GT(df1).data_color("x")
> ```
> 
> It seems like...
> 
> * the divide by zero error is a range calculation issue. Maybe we could use a try/except to catch division by zero, or handle the logic inside that function so division by zero doesn't occur?
> * the min error seems similar (as it occurs as part of the calculation of the range).
> 
> **edit:** What if---when the range is 0 (all same value)---you scaled the values to some arbitrary value? e.g. 0, 1, or .5. It seems like a core issue is that what the scale value should be is ambiguous. Similarly, when all the values are None, it doesn't matter what you set the range to, because everything will remain None.

This is now handled where the scaling occurs. If the range is zero, then the scaling factor is automatically set to `0.0` which chooses the first color in the palette (mirroring the other edge case behavior).

great_tables/_data_color/base.py
```diff
@@ -212,13 +212,65 @@ def data_color(
...
 
+        # If there are only NA values in `column_vals`, regardless of the length of
+        # `column_vals`, then apply the `na_color=` color to the entire column and
+        # then move on to the next column
+        if all(is_na(data_table, x) for x in filtered_column_vals):
+            gt_obj = gt_obj.tab_style(
+                style=fill(color=na_color), locations=body(columns=col, rows=[0])
+            )
+            continue
+
```
machow, line 273: It seems like all the code added here is unneeded now

machow (changes requested):
Do you mind double checking whether the initial catches are needed anymore? It seems like it should be okay now.

The one piece you might need to add is on this line...

https://github.com/posit-dev/great-tables/pull/213/files#diff-85b631a9e1bb79b5f833b150410a89fc3ef04401e7cfd07e8eee0a6d3f33624eR277

In python, all over an empty iterator evaluates to true:

```
all(ii for ii in [])
```

So you might need to do:

```python
if len(filtered_column_vals) and all(isinstance(x, (int, float)) for x in filtered_column_vals):
```

great_tables/_data_color/base.py
rich-iannone: I removed the first and last blocks. Failing test cases occurred from removing any more.

machow (approved):
Let's goooooooo!!!!

Issue #214, opened 2024-02-26
Allow opt_horizontal/vertical_padding to accept the value 3 (i.e. bounds inclusive)


machow:
Currently opt_horizontal_padding and its vertical analog accepts between 0 and 3, but does not include 3 itself. Let's allow passing 3, since it's more convenient to say 3 for the max, rather than 2.99, etc..

```python
from great_tables import GT, exibble

# raises error
GT(exibble).opt_horizontal_padding(3)
```

PR #215, opened 2024-02-27
Fix `opt_horizontal_padding()` example

rich-iannone:
This slightly modifies the `opt_horizontal_padding()` example in the documentation. This one is a bit odd because during manual testing, `GT.opt_horizontal_padding(scale=3)` did not seem to work (which was chalked up to a bounds issue in an argument check) but the code shows it does accept a `scale` value of 3---and the API docs also has an example of this working in https://posit-dev.github.io/great-tables/reference/GT.opt_horizontal_padding.html#great_tables.GT.opt_horizontal_padding.

Fixes: https://github.com/posit-dev/great-tables/issues/214

### DIFF
--- a/docs/get-started/table-theme-premade.qmd
+++ b/docs/get-started/table-theme-premade.qmd
@@ -60,7 +60,7 @@ This section shows the different `GT.opt_*()` methods available. They serve as c
 ### Align table header
 
 ```{python}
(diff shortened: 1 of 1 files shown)


machow:
Oh strange, thanks for looking into it :o

Issue #98, opened 2023-12-19
Using row ID values in `rows` is inconsistent when the table has row groups


rich-iannone:
When the table has row groups, rows tend to lose their original ordering due to migration of rows to the different groups. This makes using a row index value in a `rows` argument less recommended (you have to know the pre-shuffled index). Using the row ID is the recommended way, however, this fails and it seems that the row ID maps to the wrong row. Here is an example of that using `tab_style()`:

```py
from great_tables import GT, md, style, loc
from great_tables.data import towny

towny_mini = towny[["name", "website", "land_area_km2", "csd_type"]].tail(4)

(
    GT(towny_mini, rowname_col="name", groupname_col="csd_type")
    .tab_style(
        style=style.fill(color="Beige"),
        locations=loc.body(columns="land_area_km2", rows=["Woodstock"]),
    )
)
```

machow:
From pairing, it sounds like the row ID should always be the original position (as passed into GT)

rich-iannone:
This was partially fixed by @machow  in https://github.com/posit-dev/great-tables/pull/202. However, I'm seeing that the Pandas table behaves incorrectly compared to a Polars one. What follows is a pair of examples that involves a GT table having a stub and row groups, and, some basic styling.

Let's take a portion of the `towny` dataset as a Pandas DataFrame. From the output we see that 'Woodstock' is the sole city under the `csd_type` column: 

```python
from great_tables import GT, md, style, loc
from great_tables.data import towny
import polars as pl

towny_mini = towny[["name", "website", "land_area_km2", "csd_type"]].tail(4)

towny_mini
```

<img width="667" alt="pandas-output" src="https://github.com/posit-dev/great-tables/assets/5612024/22283722-8c83-463a-8f3b-b2dc37586b66">

Let's write the GT statement for this Pandas DF; notice that 'Woolwich' is in the 'city' group.

```python
# With Pandas -- incorrectly classifies Woolwich as a city; styling the 'Woodstock' row is also incorrect
(
    GT(towny_mini, rowname_col="name", groupname_col="csd_type")
    .tab_style(
        style=style.fill(color="Beige"),
        locations=loc.body(columns="land_area_km2", rows="Woodstock"),
    )
)
```

<img width="775" alt="pandas-GT" src="https://github.com/posit-dev/great-tables/assets/5612024/d73a8fc7-cb9a-4d3b-ba8a-38b186ee2689">

With a Polars DF, everything is correct:

```python
[...]

PR #218, opened 2024-02-27
Fix render reorder causing incorrect groupings

machow:
Currently, Great Tables keeps a GT._body.body dataframe, ordered to match the final output (e.g. sorted by group). This means that what happens currently looks like this:

1. formatters run, put results on GT._body.body
2. GT._build_data() returns a new GT object, with `._body.body` reordered to match final output
3. GT.render_as_html() -> create_body_component_h() : attempts to set any missing values in ._body.body to a string version of the original data. This is to fill in any unformatted values.

However, there are two issues with this:

* We attempt to grab rows from `_body.body` when rendering html, as if the dataframe hasn't been sorted. (but it has).
* With polars, we are accidentally filling in missing values in step (3) with the original data, in its original order.

In order to fix this, we need to decide whether `_body.body` should be sorted, and make sure it's filled consistently across pandas and polars.

This PR simply removes the sorting of `_body.body`. Since we index based on the original data (e.g. passing `row=1` to a formatter refers to that row position in the original, unsorted data), it seems like we can avoid sorting `._body.body`, and just grab the correctly ordered rows when rendering.

(Note that the snapshot test I added when addressing #201 was wrong 😭, and I fixed it here)

Fixes: https://github.com/posit-dev/great-tables/issues/98


machow:
Shoot, I checked out this branch from the formatter PR one. Once that's merged I'll just get this one caught up, so it's less chaotic!

rich-iannone:
> Shoot, I checked out this branch from the formatter PR one. Once that's merged I'll just get this one caught up, so it's less chaotic!

The formatter PR has been reviewed (great work!) so it can be merged.

machow:
Alright, it should now just show the diff against main (no fmt method changes)! Note that this PR makes a fairly substantial change of keeping `_body.body` in the original input order. This means nothing gets reordered, only that the renderer func grabs rows in final display order.

Definitely let me know if this will cause issues in the future, based on any gt for R behaviors!

rich-iannone (approved):
Looks great! Thank you!

PR #219, opened 2024-02-27
Add initial implementation for nanoplots

rich-iannone:
This adds a minimal, working version of nanoplots functionality to the package. The focus initially is to get utility functions to generate SVG line plots and then hook this up to a new `.fmt_nanoplot()` method.


machow (approved):
From pairing, we're going to merge this for now (and not release), without review. I'll look through it (or the diff) later and leave comments and suggestions (and also just get up to speed).

Issue #230, opened 2024-03-06
Allow special color keywords `"transparent"` and `"currentColor"`


rich-iannone:
Right now we accept named colors and hex colors but not these CSS color keywords.

rich-iannone:
Relevant discussion: https://github.com/posit-dev/great-tables/discussions/151#discussioncomment-8681542

PR #232, opened 2024-03-06
Allow the 'transparent' color (and others) to be used in `tab_options(table_background_color == <color>)`

rich-iannone:
This PR makes it possible for a GT table to have a transparent background by way of using the `"transparent"` color keyword. Previously, there would be an error because color values in `_font_color()` are routed through a **webcolors** function that only accepts CSS color names.

Fixes: https://github.com/posit-dev/great-tables/issues/230


machow:
I don't understand why this PR also refactors away the `data_color` submodule. Can you provide brief descriptions of what big things you did in this PR and your thinking behind each one?

tests/__snapshots__/test_options.ambr
```diff
@@ -31,7 +31,7 @@
     line-height: normal;
     margin-left: 10px;
     margin-right: 10px;
-    color: lightred;
+    color: rebeccapurple;
```
machow, line 34: I don't have enough context in the PR description to understand what this means or why it is important

machow (changes requested):
I am having difficulty reviewing this PR. It seems like a number of refactoring things were done, in addition to the feature. I can't quickly separate what is refactoring and what is core to allowing the "transparent" color in tab_options.

Can you clarify what I should be looking for?

great_tables/_scss.py
```diff
@@ -39,23 +39,30 @@
...
-def font_color(color: str, table_font_color: str, table_font_color_light: str):
-    if color.startswith("#"):
-        rgb = wc.hex_to_rgb(color)
-    elif color.startswith("rgb") and "%" in color:
-        # TODO: rgb_percent_to_rgb() expects a tuple
-        raise NotImplementedError()
-        rgb = wc.rgb_percent_to_rgb(color)
-    else:
-        rgb = wc.name_to_rgb(color)
+def _font_color(color: str, table_font_color: str, table_font_color_light: str) -> str:
```
rich-iannone, line 42: This `_font_color()` function was reworked for two reasons:

1.  we already had some existing logic for determining which opposing color (choice between a predetermined `dark` and `light` color) should be used for maximal contrast (that was used in `data_color()`)
2. we documented that all CSS3 and even X11 color names would be accepted as named colors. We didn't include other CSS color keywords like `"transparent"` and `"currentColor"`. The webcolors package doesn't support those and we already have `_html_color()` which accepts them handles the normalization to hex color as well.

great_tables/_utils_color.py
```diff
@@ -0,0 +1,314 @@
+# Utility functions for color handling
```
rich-iannone, line 1: I moved this out of the `data_color()` sub dir because (and this may be not the best reason) at least one of the related functions was being used in `_scss.py`. We can totally move this all back if you're thinking that's not worth it right now.

great_tables/_scss.py
```diff
@@ -39,23 +39,30 @@
...
+        bgnd_color=color_normalized[0],
+        light=table_font_color,
+        dark=table_font_color_light,
+    )
 
-    return table_font_color_light
+    return ideal_font_color
 
 
-def css_add(value: str | int, amount: int):
```
rich-iannone, line 58: I'm still not entirely clear on whether to consistently use leading underscores for 'private' functions/methods. Maybe we could talk about this briefly so I could better understand whether it's needed virtually everywhere.

tests/__snapshots__/test_options.ambr
rich-iannone: The `"lightred"` color is not a valid CSS color name (and not a web safe color). So using it results in an error. I instead opted for `"rebeccapurple"` since it's a valid CSS3 color name (and not a valid X11 color name).

tests/test_scss.py
```diff
@@ -2,19 +2,19 @@
...
 
 from great_tables import GT
-from great_tables._scss import font_color, css_add, compile_scss
+from great_tables._scss import _font_color, _css_add, _compile_scss
 
 
 @pytest.mark.parametrize(
     "src,dst",
     [
-        ("#FFFFFF", "dark"),
```
rich-iannone, line 11: The changed `_font_color()` util function requires a valid color name in it's latter two args. So, I changed the param values of  `"light"` and `"dark"` to light and dark color hex values.

great_tables/_scss.py
machow: Let's discuss. These types of changes aren't really related to the PR, so I would push them out of scope (and then confirm in words the refactor changes when the PR is reviewed. That way the relevant fix can go through review quickly, and refactor changes can be applied at the end).

Making these kinds of changes brings a lot of extra files into the PR review

machow (changes requested):
Thanks for adding comments, explaining what you did. I've tried to look through the PR again, and am still having trouble making heads or tails of it. The refactor changes, though each one is small, together are too extensive.

I've gone back through individual commits to try and see if I can isolate just the fix changes, but the refactor seems to cut through the commit history.

Do you mind if we pair on re-doing this PR? I think it should be pretty quick. Sorry to hold things up, but I want to make sure to fully appreciate what this fix is doing.

rich-iannone:
Closing this in favor of https://github.com/posit-dev/great-tables/pull/242.

Issue #228, opened 2024-03-05
Wrap rendered html table in html or body tags


machow:
from pairing w/ @isabelizimm, it would be useful to have the output wrapped in html or body tags.

PR #233, opened 2024-03-07
Improve HTML representation of tables

rich-iannone:
This PR adds the option to generate a complete HTML page around the emitted table fragment. And this vastly improves the printing of a table in the vscode jupyter notebook environment by raising the specificity of the tables CSS rules (in the `<style>` block). We also detect whether rendering is occurring in a Quarto document and, in that case, the compiled CSS is kept as is.

Fixes: https://github.com/posit-dev/great-tables/issues/228


great_tables/gt.py
```diff
@@ -258,8 +255,21 @@ def __init__(
 
     # -----
 
+    def _repr_html_(self):
+        import os
+
+        # Check if we are rendering in the Quarto environment
+        quarto_enabled = os.environ.get("QUARTO_BIN_PATH") is not None
+
```
isabelizimm, line 263: It might be nice to do some handling to make a guess on if you should make a page or not. Eg, make_page should be false if in an IPython terminal or Jupyter Notebook, but true in other occasions?

You can do checks for these situations with something like:

```python
options = ["TerminalInteractiveShell", "ZMQInteractiveShell"]
make_page = ipython.get_ipython().__class__.__name__ not in options
```

where
 Jupyter Notebooks ➡️ 'ZMQInteractiveShell'
IPython in a terminal  ➡️ 'TerminalInteractiveShell

rich-iannone: The latest in this PR includes code to determine whether a full page should be returned. Thanks for the example code!

great_tables/gt.py
```diff
@@ -291,14 +319,37 @@ def _build_data(self, context: str) -> Self:
 
         return built._replace(_body=final_body, _stub=final_stub)
 
-    def render(self, context: str) -> str:
-        html_table = self._build_data(context=context)._render_as_html()
+    def render(
+        self,
+        context: str,
+        make_page: bool = False,
```
machow, line 325: If we want to give people the option to set this, but not require it (e.g. render feels like it can run "automatically", similar to `_repr_html_()`), we could set arguments like `make_page=None` by default. In the case where `make_page` and `all_important` are both None, this might essentially produce the same behavior as `_repr_html_()`.

See comment in `_repr_html_()`. If we separate environment detection and options fetching (e.g. get `make_page` and `all_important` for environment X) then we also should be able to use that code here.

great_tables/gt.py
machow: Just a quick note, ZMQInteractiveShell is a subclass of TerminalInteractiveShell, so we can pick up either in a isinstance(..., TerminalInteractiveShell) check

machow: No wait, ZMQ subclass InteractiveShell, which is a lil bit different 😅

machow: For posterity, here is all the handy checks plotly does on environments:

https://github.com/plotly/plotly.py/blob/95cc3d0dc6b3379d181e28ba90a0b708fad3bc40/packages/python/plotly/plotly/io/_renderers.py#L514

machow: @isabelizimm I've consolidated logic for detecting environments here:

https://github.com/posit-dev/great-tables/pull/233/files#diff-4725a5793cb0f0916ca5679673d3a72a4f2896f336e87af59f5cacc157b04dfeR15

We can tack things onto there pretty easily!

great_tables/gt.py
machow: From pairing, it sounds like we might want to use the following strategy:

* in the future, `.render()` might take a dataclass as context. The dataclass will hold rendering options (e.g.  LatexContext)
* in the now, `.as_raw_html()` is essentially the context specific version. The parameters in `.as_raw_html()` are what the options in the HtmlContext would be, etc..

machow:
From pairing w/ Rich, I've consolidated the code for inferring the IDE environment into the `_render.py` submodule, and hard-coded the IDE environment specific flags there.

great_tables/gt.py
```diff
@@ -291,14 +319,37 @@ def _build_data(self, context: str) -> Self:
...
+        page_background_color: str = "white",
```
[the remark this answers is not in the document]
rich-iannone: Now taken out.

great_tables/gt.py
```diff
@@ -258,8 +254,40 @@ def __init__(
...
+            make_page = False
+        else:
+            ipython_options = ["TerminalInteractiveShell", "ZMQInteractiveShell"]
+            make_page = get_ipython().__class__.__name__ not in ipython_options
+
+        rendered = self.render(context="html", make_page=make_page, all_important=all_important)
+
+        return rendered
+
     def _get_has_built(self: GT) -> bool:
```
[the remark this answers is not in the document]
rich-iannone: Removed this function.

tests/test_repr.py
```diff
@@ -0,0 +1,49 @@
...
+    return GT(pd.DataFrame({"x": [1, 2], "y": [4, 5]}), id="test")
+
+
+def assert_rendered_html_repr(snapshot, gt):
+
+    html_repr_output = gt._repr_html_()
+    assert snapshot == html_repr_output
+
+
+@mock.patch.dict(os.environ, {"QUARTO_BIN_PATH": "1"}, clear=True)
```
machow, line 21: This is neat!

great_tables/gt.py
```diff
@@ -363,6 +385,21 @@ def _render_as_html(self) -> str:
...
         """
 
+        if make_page:
+
+            # Create an HTML page and place the table within it
+            finalized_table = f"""<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="utf-8"/>
+<style>body {{ background-color:white; }}
```
machow, line 395: This styles every body element on the page to white (including the overall body element for the page). We need a more specifically scoped css style. If we want to set only the body element defined here to white, then we need to either...

* specify the style inline
* give the body element a unique id and set the style on that

machow (changes requested):
This is looking good -- the one big note I left is that the css style setting body background to white is too broad--it can affect the entire page (similar to the screenshot I posted that turned the page blue).

great_tables/gt.py
rich-iannone: Sorry, didn't go all the way with this. I'm going to remove it entirely.

rich-iannone:
@isabelizimm @machow thank you for your help with this PR! Going to merge now but there should be many more improvements in this particular area down the road.

Issue #224, opened 2024-03-04
Table header (by way of `.tab_header()`) generates malformed and invalid HTML


rich-iannone:
Using the `.tab_header()` method generates one or two `<th>` elements. However, the tags are not properly closed. A browser will render this, but passing this HTML to anything else that consumes it as input will be problematic since it doesn't pass HTML validation. Furthermore, there should only be a single `<th>` element.

rich-iannone:
Related discussion: https://github.com/quarto-dev/quarto-cli/discussions/8971

PR #235, opened 2024-03-08
Ensure that header component HTML tags pass HTML validation

rich-iannone:
There were unclosed tags when generating a header. This generally causes problems when the output HTML is consumed by another process (expectation is that the input HTML should be valid; browsers can handle invalid HTML to great extent, other programs cannot really).

This PR fixes this issue and also performs a few HTML formatting cleanups at the same time. The output HTML was verified in the Nu HTML checker tool (at https://validator.w3.org/nu) to have been improved from the changes here. 

Fixes: https://github.com/posit-dev/great-tables/issues/224

And this should fix a user issue in https://github.com/quarto-dev/quarto-cli/discussions/8971.

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -34,19 +34,23 @@ def create_heading_component_h(data: GTData) -> StringBuilder:
         stub=data._stub, row_groups=data._row_groups, options=data._options
     )
 
-    result.append(
-        f"""  <tr>
-    <th colspan="{n_cols_total}" class="gt_heading gt_title gt_font_normal">{title}
-  </tr>"""
-    )
-
     if has_subtitle:
-        subtitle_row = f"""  <tr>
-    <th colspan="{n_cols_total}" class="gt_heading gt_subtitle gt_font_normal gt_bottom_border">{subtitle}
+        heading = f"""
+  <tr>
+    <th colspan="{n_cols_total}" class="gt_heading gt_title gt_font_normal">{title}</th>
+  </tr>
+  <tr>
+    <th colspan="{n_cols_total}" class="gt_heading gt_subtitle gt_font_normal gt_bottom_border">{subtitle}</th>
   </tr>"""
-        result.append(f"\n{subtitle_row}")
+    else:
+        heading = f"""
+  <tr>
+    <th colspan="{n_cols_total}" class="gt_heading gt_title gt_font_normal">{title}</th>
+  </tr>"""
 
-    return StringBuilder('<thead class="gt_header">', result, "</thead>")
+    result.append(heading)
+
+    return StringBuilder('<thead class="gt_header">', result, "\n</thead>")
 
 
 def create_columns_component_h(data: GTData) -> str:
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -21,6 +21,10 @@ def create_heading_component_h(data: GTData) -> StringBuilder:
     if not has_title and not has_subtitle:
         return result
 
+    # Raise an error if there is a subtitle but no title
+    if not has_title and has_subtitle:
+        raise ValueError("A subtitle was provided without a title.")
+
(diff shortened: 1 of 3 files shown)


machow:
Can you paste in what you saw that indicated something was fixed? Either the output of a tiny reprex, or a screenshot of what you looked at that indicated things were fixed?

great_tables/_utils_render_html.py
```diff
@@ -21,6 +21,10 @@ def create_heading_component_h(data: GTData) -> StringBuilder:
     if not has_title and not has_subtitle:
         return result
 
+    # Raise an error if there is a subtitle but no title
```
rich-iannone, line 24: Added an additional check here to raise in the case that subtitle text is available in the absence of title text. While this may be also caught within the `.tab_header()` method, I think it's good to have a second check here in case there is manual addition/removal of title/subtitle (manipulations done outside the prescribed API).

great_tables/_utils_render_html.py
```diff
@@ -154,8 +162,8 @@ def create_columns_component_h(data: GTData) -> str:
                 )
             )
 
-        # Join the <th> cells into a string and separate each with a newline
-        th_cells = "\n".join([str(tag) for tag in table_col_headings])
+        # Join the <th> cells into a string and begin each with a newline
```
rich-iannone, line 165: An opportunity was taken here to beautify the HTML output of the column labels portion since snapshots had to be regenerated anyway.

rich-iannone:
There are generally quite a few HTML validation problems with the generated output but this PR does address the closing of `<th>` values in the title block. Here are before and after screenshots of the changes.

Before PR changes:

html-before

After:

html-after

I was going to attach a screenshot of the HTML checker results but since there are quite a few problems to be resolved (beyond the scope of this PR) the failure of not having closed `<th>` is masked by preceding validation failures.

tests/__snapshots__/test_export.ambr
```diff
@@ -49,16 +49,20 @@
    
   </style>
   <table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">
-  <thead class="gt_header">  <tr>
-      <th colspan="4" class="gt_heading gt_title gt_font_normal">Data listing from <strong>exibble</strong>
+  <thead class="gt_header">
+    <tr>
+      <th colspan="4" class="gt_heading gt_title gt_font_normal">Data listing from <strong>exibble</strong></th>
```
machow, line 54: Here is the test that changed, based on this fix. There is a closing `</th>` element now.

tests/__snapshots__/test_export.ambr
```diff
@@ -49,16 +49,20 @@
...
-      <th colspan="4" class="gt_heading gt_subtitle gt_font_normal gt_bottom_border"><code>exibble</code> is a <strong>Great Tables</strong> dataset.
-    </tr></thead>
-  <tr class="gt_col_headings"><th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col" id=""></th>
-  <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="num">num</th>
-  <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col" id="char">char</th>
-  <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="currency">currency</th></tr>
+      <th colspan="4" class="gt_heading gt_subtitle gt_font_normal gt_bottom_border"><code>exibble</code> is a <strong>Great Tables</strong> dataset.</th>
+    </tr>
+  </thead>
+  <tr class="gt_col_headings">
```
machow, line 60: This and the rest of snapshot test changes are just stylistic

machow:
Thanks, this is super helpful! Based on these notes, I was able to find the snapshot that updated with the added `</th>`.

PR #238, opened 2024-03-10
draft: coloring using matplotlib normalizations

ChristopherRussell:
# Summary

Thank you for contributing to **great_tables**... this is great! To make this process easier for everyone, please explain the context and purpose of your contribution. Also, list the changes made to the existing code or documentation.

- Ref: #

# Checklist

### DIFF
--- a/great_tables/_data_color/__init__.py
+++ b/great_tables/_data_color/__init__.py
@@ -1,3 +1,4 @@
 from .base import data_color
+from .matplotlib import data_color_mpl
 
-__all__ = ["data_color"]
+__all__ = ["data_color", "data_color_mpl"]
--- a/great_tables/_data_color/matplotlib.py
+++ b/great_tables/_data_color/matplotlib.py
@@ -28,6 +28,7 @@ def is_numeric_or_none(x: Any) -> bool:
 def data_color_mpl(
     self: GTSelf,
     columns: Union[str, List[str], None] = None,
+    rows: Union[int, list[int], None] = None,
     cmap: Colormap | str | list[str] | None = None,
     norm: Normalize | Callable[[float], float] | None = None,
     na_color: Optional[str] = None,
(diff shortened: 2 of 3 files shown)


ChristopherRussell:
This is a draft / to share ideas referenced in #192

machow:
Thanks so much for sharing this draft, and I'm so sorry it took so long to get to it 😓. We really want to put focus on `data_color` at some point in the future, and this will be a useful reference. I'm going to close for now, since we likely will not get to it in the near term (and will need to figure out how / whether to depend on matplotlib).

ChristopherRussell:
I understand the hesitation with an unnecessary matplotlib dependency. I picked it up initially as a time-saver because it was familiar but it's definitely avoidable and I'd agree is good to not be tied to.

I've continued developing and using my own data_color features locally, including removing matplotlib dependencies. I'd be happy to help implement something that aligns with your vision while providing the enhanced functionality I'm lookng for. Would you be interested in discussing this further?

ChristopherRussell:
I understand the hesitation with an unnecessary matplotlib dependency. I initially used it as a familiar quick solution, but it's definitely avoidable and I agree it's good to avoid external dependencies.

I've continued developing and using my own data_color features, including removing matplotlib dependencies. I'd be happy to help implement something that aligns with great_tables' design principles while providing the functionality I'm looking for.

To recap, I have two main goals:
(i) Allow norm functions that remap cell values before coloring (e.g., log scaling for better visualization of wide-ranging data)
(ii) Allow coloring domain to be inferred across rows, columns, or entire selection rather than column-by-column. E.g. when you don't specify a domain it's inferred to be `[min, max]` for the min and max in the column but it should also be possible to do row-wise, or entire selection-wise.

@machow @rich-iannone would you be interested in discussing this further? I'll share a couple of examples of these features on a made-up dataset. While possible with current GT, I had to do a lot of work to make the domain argument right that could be automatic, and had to essentially use a dataframe with the values used for coloring rather than the values to be show, and use `.fmt` with a dictionary to replace the former with the latter - very hacky.

1) Log-scaling norm can be useful if you want to visually distinguish with colors values that span a couple orders of magnitude. Also, inferred domains are very natural for pivots. And pivots with a row of column totals naturally wants a row-wise inferred domain.
Screenshot 2025-06-10 at 15 02 20

 2) Symmetric norms (in this case symmetric linear centered at 0) are nice for datasets where you want to use a diverging color map to distinguish something like positive and negative values (e.g. profits/losses as green/red) when the min and max value are on different scales.
Screenshot 2025-06-10 at 15 05 58

Issue #211, opened 2024-02-22
Add the `fmt_units()` method


rich-iannone:
Having the method `fmt_units()` available would allow for formatting of text in 'units notation' in the table body to nicely formatted measurement units.

Issue #169, opened 2024-02-06
epic: Implement `fmt_units()` and `cols_units()`


machow:
These functions use a Great Tables specific mini language to specify units in labels. Note that functions like `md()` can do a lot of unit stuff already (like creating superscripts), but it's more limited (e.g. some combos of subscripts and superscripts) and this allows Great Tables to render to things like latex, etc..

https://gt.rstudio.com/reference/cols_units.html

machow:
Closing, since we have `fmt_units()` now -- let's open an issue for `cols_units()` once it seems important

PR #240, opened 2024-03-12
Add the `fmt_units()` method

rich-iannone:
This PR adds the `fmt_units()` method. This performs a conversion of units in the *units notation* syntax (e.g., `"x10^9 / L"` or `"x10^9 L^-1"`, etc.) to HTML. The method, like others of the `fmt_*()`, only concerns itself with text transformations in the table body. Other methods will also gain the ability to convert text in units notation to nicely formatted HTML in later PRs.

Here's an example that uses `fmt_units()` with the `illness` dataset. It so happens that the `units` column has strings in units notation, so, we just need to point this method to that column:

```python
from great_tables import GT, style, loc
from great_tables.data import illness

(
    GT(illness, rowname_col="test")
    .fmt_units(columns="units")
    .fmt_number(columns=lambda x: x.startswith("day"), decimals=2, drop_trailing_zeros=True)
    .tab_header(title="Laboratory Findings for the YF Patient")
    .tab_spanner(label="Day", columns=lambda x: x.startswith("day"))
    .tab_spanner(label="Normal Range", columns=lambda x: x.startswith("norm"))
    .cols_label(
      norm_l="Lower",
      norm_u="Upper",
      units="Units"
    )
    .opt_vertical_padding(scale=0.4)
    .opt_align_table_header(align="left")
    .tab_options(heading_padding="10px")
    .tab_style(
        locations=loc.body(columns="norm_l"),
        style=style.borders(sides="left")
    )
    .opt_vertical_padding(scale=0.5)
)
```

illness

Fixes: https://github.com/posit-dev/great-tables/issues/211
Partially addresses the `.epic` issue: https://github.com/posit-dev/great-tables/issues/169


machow (changes requested):
From pairing a while ago, let's pause this until bug fixes, testing, and other work is closed out.

machow:
Let's...

* put an explanation of the rules under `define_units()`. e.g. it looks like there are 5 major modes.
* create a notebook (or quarto doc) show individual input -> output examples (could be a table, but we need to be able to see the raw input string)
  - it should be clear what new piece of logic each added example explains
  - e.g. we should not have examples that feel like they're repeating an earlier rule, unless they reinforce it
  - if we have a table, we should have a column for "rule" or something, articulating the rule the example shows

tests/test_formats.py
```diff
@@ -1489,6 +1489,415 @@ def test_fmt_image_path():
     assert 'src="/a/b/c"' in strip_windows_drive(res)
 
 
+def test_fmt_units():
```
machow, line 1492: It's not clear what about fmt units is being tested here. Please...

* test values to be formatted one by one
* only test values that test / make clear something new about fmt_units

tests/test_helpers.py
```diff
@@ -81,3 +96,171 @@ def test_get_font_stack_add_emoji_true(font_stack_names):
...
+            "-20 kg^2 m_12 m[_0^2] g/L %C6H12O6% x10^-3",
+            ["-20", "kg^2", "m_12", "m[_0^2]", "g/L", "%C6H12O6%", "x10^-3"],
+        ),
+        (
+            "  -20  _kg_^2      *m*_12  m[_0^2] g/L%C6H12O6% x10^-3",
+            ["-20", "_kg_^2", "*m*_12", "m[_0^2]", "g/L%C6H12O6%", "x10^-3"],
+        ),
+        (
+            "m^-3 :cdot: kg :cdot: m[_0^2] *per:space:mille*",
+            ["m^-3", ":cdot:", "kg", ":cdot:", "m[_0^2]", "*per:space:mille*"],
```
machow, line 115: I don't understand what different things are being tested across these cases. The 3rd seems to be testing tokens splitting multiple spaces, but that doesn't seem to be all it tests. Please break the tests into...

* tests of small and specific bits of logic
* new tests should test / clarify new pieces of behavior

It's okay to have some bigger inputs tested, but it should be clear / quickly explainable what specifically is being tested / what's new compared to prev tests.

tests/test_helpers.py
```diff
@@ -81,3 +96,171 @@ def test_get_font_stack_add_emoji_true(font_stack_names):
...
+@pytest.mark.parametrize(
+    "content, x_out",
+    [
+        ("2", '<span style="white-space:nowrap;"><sub>2</sub></span>'),
+        ("2*e*5", '<span style="white-space:nowrap;"><sub>2*e*5</sub></span>'),
+        ("", '<span style="white-space:nowrap;"><sub></sub></span>'),
+    ],
+)
+def assert_units_to_subscript(content: str, x_out: str):
+    x = _units_to_subscript(content=content)
```
machow, line 136: `_units_to_subscript()` is a single line function that produces a string. I don't understand what new thing is being tested by e.g. using both `"2"` and `"2*e*5"` in the tests.

tests/test_helpers.py
```diff
@@ -81,3 +96,171 @@ def test_get_font_stack_add_emoji_true(font_stack_names):
...
+@pytest.mark.parametrize(
+    "content, x_out",
+    [
+        ("34.3", '<span style="white-space:nowrap;"><sup>34.3</sup></span>'),
+        ("4*d*2", '<span style="white-space:nowrap;"><sup>4*d*2</sup></span>'),
+        ("", '<span style="white-space:nowrap;"><sup></sup></span>'),
+    ],
+)
+def assert_units_to_superscript(content: str, x_out: str):
+    x = _units_to_superscript(content=content)
```
machow, line 149: similar to above

tests/test_helpers.py
```diff
@@ -81,3 +96,171 @@ def test_get_font_stack_add_emoji_true(font_stack_names):
...
+    [
+        (
+            "*2i*",
+            "3.23517",
+            '<span style="display:inline-block;line-height:1em;text-align:left;font-size:60%;vertical-align:-0.25em;margin-left:0.1em;">3.23517<br>*2i*</span>',
+        ),
+        (
+            "0",
+            "e2.7",
+            '<span style="display:inline-block;line-height:1em;text-align:left;font-size:60%;vertical-align:-0.25em;margin-left:0.1em;">e2.7<br>0</span>',
```
machow, line 164: Do you mind moving the manually repeated pieces in these x_out pieces into the test function itself (e.g. vs manually coding `<span...>{value_we_care_about}</span>` twice, just use `value_we_care_about` for `x_out`, and add it to the "<span..." stuff inside the test)

machow (changes requested):
Shoot sorry---I thought this PR had re-requested review, so made some comments 😅 . I'm going to submit the review, in case they're helpful, but sorry for jumping early on this :o

tests/test_helpers.py
rich-iannone: Just did this, thanks!

tests/test_helpers.py
rich-iannone: I removed the latter since it was redundant.

tests/test_helpers.py
rich-iannone: I removed the redundant test.

tests/test_helpers.py
rich-iannone: This is now much simpler and much improved!

great_tables/_helpers.py
```diff
@@ -518,6 +521,458 @@ def _get_font_stack(name: FontStackName = "system-ui", add_emoji: bool = True) -
...
+
+    def to_html(self) -> str:
+
+        from great_tables._text import _md_html
+
+        for i in range(len(self)):
+
+            units_str_i = ""
+
+            units_object_i = self[i]
```
machow, line 573: Currently, this code does the following:

* create a new variable for each attribute
* redefine that variable in the clauses below
* check and potentially redefine that variable again in later clauses

Could you do this instead?

* Do not create a new variable as a short-cut for an attribute
* Create new variables only for final, potentially transformed attributes
* Group logic for defining a new variable like `exponent` together (e.g. in a if/elif/else block)
* Do not edit the new variables in later clauses

This should get rid of re-assignments showing up much after things are defined.

For example, exponent is..

* defined on 576
* potentially redefined on 596
* potentially redefined on 603
* potentially redefined on 626
* potentially redefined on 628

But it seems like you could put all exponent stuff in on tight if/else block:

```python
# since .to_html() should be moved onto the unit, assume unit is unit_object[i]
# create exponent part ----

if unit.exponent is None:
    exponent = None
elif len(unit.exponent) > 2 and "*" in unit.exponent:
    exponent = _md_html(_units_symbol_replacements(text=exponent))
else:
    exponent = _units_to_superscript(content=exponent)

# I could see a small redefinition here, if it is close to initial definition and clear
# it won't happen later one. But I also don't know if this is right, since in this
# original code this doesn't occur when chemical_forumla is True.
# But that seems like a bug in the original code?
if exponent is not None:
    exponent = exponent.replace("-", "&minus;")


# create unit subscript part ----
unit_subscript = ...


# after everything that can be defined early is defined ===========
# create the final output string

unit_str = ""

# append sub super part ----

if unit.sub_super_overstrike and unit_subscript is not None and exponent is not None:
    # no redefining earlier parts here
    # unit_str is okay, since it is like a final result list being appended to
    unit_str += _units_html_sub_super(
        content_sub=unit_subscript, content_sup=exponent
    )
```

great_tables/_helpers.py
```diff
@@ -518,6 +521,458 @@ def _get_font_stack(name: FontStackName = "system-ui", add_emoji: bool = True) -
...
+
+    for i in range(len(tokens_list)):
+
+        tokens_list_i = tokens_list[i]
+
+        unit_subscript = None
+        sub_super_overstrike = False
+        chemical_formula = False
+        exponent = None
+
```
machow, line 893: Do you mind providing an example for these high-level cases in comments, so to queue people on what to look for in the regex code?

For example...

```python
        # Case: chemical formula
        #   * e.g. "%C6H12O6%"
        if re.match(r"^%.*%$", tokens_list_i) and len(tokens_list_i) > 2:
            chemical_formula = True

        # Case: subscript and superscript present, which requires an overstrike
        #   * e.g. "h[_0^3]"
        elif re.search(r".+?\[_.+?\^.+?\]", tokens_list_i):
            ...

        # Case: simple subscript and exponent
        #   * e.g. "h_0^3"
        elif re.search(r".+?_.+?\^.+?", tokens_list_i):
```

Having the case comments outside the blocks seems like it may help with reading a bit here, since regexes by themselves are a bit terse (so in this case the comment will be more prominent and hopefully give some hints to help with reading the regex).

great_tables/_helpers.py
```diff
@@ -518,6 +521,458 @@ def _get_font_stack(name: FontStackName = "system-ui", add_emoji: bool = True) -
...
+    # Get a list of raw tokens
+    tokens_list = _generate_tokens_list(units_notation=units_notation)
+
+    # Initialize a list to store the units
+    units_list = []
+
+    if len(tokens_list) == 0:
+        return UnitDefinitionList(units_list=[])
+
+    for i in range(len(tokens_list)):
```
machow, line 885: Because...

* this loop creates a UnitDefinition object for each token
* UnitDefinition will likely never be instantiated directly (because you have to input token, along with everything else this loop calculates)

It might be good as a constructor on `UnitDefinition`?

e.g.

```python
@dataclass
UnitDefinition:

    ...


    @classmethod
    def from_token(cls, token: str) -> UnitDefinition:
        # logic from the loop here ----

        unit_subscript = None
        sub_super_overstrike = False
        chemical_formula = False
        exponent = None

        ...
        
        return cls(token, ...)
```

machow (changes requested):
Thanks for putting all this work in the PR, especially into documenting define units. The tests are looking really great. I left some comments suggesting some ways we might be able to clean up the code a little.

These mostly center on...

* grouping the logic for a variable definition together (into tighter if/elif/else blocks)
* creating a `UnitDefinition.from_token()` method, since the class doesn't seem like it can be instantiated directly.

great_tables/_helpers.py
rich-iannone: Now implemented.

great_tables/_helpers.py
rich-iannone: The code here has now been refactored to avoid multiple redefinitions and to focus on one component at a time. Thanks for the detailed advice here!

machow (changes requested):
TODO:

- need to escape `>` and `<`
- let's rewrite the specification section of define units to cover the rules of the DSL

## Unit DSL Rules

From pairing w/ @rich-iannone, here are what seems like the rules of define_units:

```python
# Within unit rules ----
# 1. ^ creates a superscript
# 2. _ creates a subscript
# 3. subscripts and superscripts may be combined
#   - however, _ inside a superscript does not create a superscript

# 4. use [_subscript^superscript] to create an overstrike

# 5. / at the beginning adds the superscript -1
# 6. hyphen is transformed to minus sign
# 7. x at the beginning transformed to ×
# 8. ascii terms from biology/chemistry turned into TERM FORM (TODO: enumerate via code)

# 9. can create italics with * or _, and can create bold with ** or __
#   - can italicize AND bold together
#   - issue: because we use commonmark, a broader set of behaviors occur
#   - e.g. **m^2**, "a<marquee>123</marquee>b"

# ISSUE: < and > are unescaped

# Special notations ----
# 10. special symbol set surrounded by colons (e.g. :angstrom:)
# 11. chemistry notation: %C6%
```

machow (approved):
This looks great, thanks for taking the time to make all the changes! One quick thing---it might be helpful to add a reference to `define_units()` in the docstring of  `fmt_units()` (but we can always punt to another PR)

rich-iannone:
@machow I added a reference to `define_units()` in the *See Also* section of the `fmt_units()` docstring.

PR #242, opened 2024-03-14
Allow the 'transparent' color (and others) to be used in tab_options(table_background_color == <color>) (supersedes previous PR)

rich-iannone:
This PR makes it possible for a GT table to have a transparent background by way of using the `"transparent"` color keyword. Previously, there would be an error because color values in `_font_color()` are routed through a **webcolors** function that only accepts CSS color names.

This PR is a more focused version of an earlier version: https://github.com/posit-dev/great-tables/pull/232

Fixes: https://github.com/posit-dev/great-tables/issues/230


machow (approved):
LGTM! Thanks for pushing this through!

PR #243, opened 2024-03-15
Add warning callout for experimental status

rich-iannone:
Since nanoplots is a very new feature (with less testing than the usual), we are placing a warning noting that this feature is currently experimental. This will be removed once adequate tests are in place.


machow (approved):
Look great, thanks!

PR #244, opened 2024-03-15
feat: basic sub_missing and sub_zero

machow:
This PR adds support for substitution functions. It address #182, by adding the following methods:

* `sub_missing`: substitutes any missing values. This includes both null and nan.
  * polars: should be equivalent to polars using `.is_null() | .is_nan()`, since distinguishes between the two.
  * pandas: should be equivalent to `.isna()`, which flags both kinds of missingness.
* `sub_zero`

Currently, it is just using the formatter machinery. As I understand, substitutions should always go after formatters. (e.g. you should be able to `.sub_zero()` right away in a chain, and expect later `fmt_*()` calls to not override that.

Note these important pieces:

* Introduced new class `FormatterSkipElement`, which when returned from a format call, indicates that no change should be made.
* Rich, in his HTML wizardry, had to address this situation: when a row of a table has all empty cells, then its height collapses (because there is no content inside). Recommended practice is to add a single `<br>`. 
* Fixed `is_na` to properly detect float("nan")
* Added the parameter `is_substitution=` to `fmt()`. This deviates from gt. Happy to change to be more similar.

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -64,6 +64,7 @@ def fmt(
     fns: Union[FormatFn, FormatFns],
     columns: SelectExpr = None,
     rows: Union[int, List[int], None] = None,
+    is_substitution=False,
 ) -> GTSelf:
     """
     Set a column format with a formatter function.
@@ -86,6 +87,8 @@ def fmt(
         In conjunction with `columns`, we can specify which of their rows should undergo
         formatting. The default is all rows, resulting in all rows in `columns` being formatted.
         Alternatively, we can supply a list of row indices.
+    is_substitution
+        Whether the formatter is a substitution. Substitutions are run last, after other formatters.
 
     Returns
     -------
@@ -105,6 +108,10 @@ def fmt(
     col_res = resolve_cols_c(self, columns)
 
     formatter = FormatInfo(fns, col_res, row_pos)
+
+    if is_substitution:
+        return self._replace(_substitutions=[*self._substitutions, formatter])
+
     return self._replace(_formats=[*self._formats, formatter])
 
 
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -491,9 +491,9 @@ def _(df: PlDataFrame, x: Any) -> bool:
     import polars as pl
     import numpy as np
 
-    from math import nan
+    from math import isnan, nan
 
-    return isinstance(x, (pl.Null, type(None))) or x is np.nan or x is nan
+    return isinstance(x, (pl.Null, type(None))) or (isinstance(x, float) and isnan(x))
 
 
 @singledispatch
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -113,6 +113,8 @@ quartodoc:
         - GT.fmt_nanoplot
         - GT.fmt
         - GT.data_color
+        - GT.sub_missing
+        - GT.sub_zero
     - title: Modifying columns
       desc: >
         The `cols_*()` methods allow for modifications that act on entire columns. This includes
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -9,6 +9,8 @@ from dataclasses import dataclass, field, replace
 from ._utils import _str_detect
 from ._tbl_data import create_empty_frame, to_list, validate_frame
 
+# TODO: move this class somewhere else (even gt_data could work)
+
 from ._styles import CellStyle
 
 # Note that we replace with with collections.abc after python 3.8
@@ -155,6 +157,9 @@ class Body:
                 raise Exception("Internal Error")
             for col, row in fmt.cells.resolve():
                 result = eval_func(_get_cell(data_tbl, row, col))
+                if isinstance(result, FormatterSkipElement):
+                    continue
+
                 # TODO: I think that this is very inefficient with polars, so
                 # we could either accumulate results and set them per column, or
                 # could always use a pandas DataFrame inside Body?
@@ -842,7 +847,12 @@ from typing import Any, Callable, TypeVar, Union, List, Optional, Tuple
 
 from ._tbl_data import n_rows
 
-FormatFn = Callable[[Any], str]
+
+class FormatterSkipElement:
+    """Represent that nothing should be saved for a formatted value."""
+
+
+FormatFn = Callable[[Any], "str | FormatterSkipElement"]
 
 
 class FormatFns:
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -7,7 +7,7 @@ from typing import overload, TypeVar, Dict, Optional
 from typing_extensions import Self, TypeAlias
 from dataclasses import dataclass, field, replace
 from ._utils import _str_detect
-from ._tbl_data import create_empty_frame, to_list, validate_frame
+from ._tbl_data import create_empty_frame, to_list, validate_frame, copy_data
 
 # TODO: move this class somewhere else (even gt_data could work)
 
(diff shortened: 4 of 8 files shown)


great_tables/_tbl_data.py
```diff
@@ -491,9 +491,9 @@ def _(df: PlDataFrame, x: Any) -> bool:
     import polars as pl
     import numpy as np
 
-    from math import nan
+    from math import isnan, nan
 
-    return isinstance(x, (pl.Null, type(None))) or x is np.nan or x is nan
+    return isinstance(x, (pl.Null, type(None))) or (isinstance(x, float) and isnan(x))
```
machow, line 496: I noticed that in python `float("nan") is float("nan")` evaluates to false. Using `math.isnan()` seems to work though for detecting either numpy or builtin nans.

great_tables/_formats.py
```diff
@@ -64,6 +64,7 @@ def fmt(
     fns: Union[FormatFn, FormatFns],
     columns: SelectExpr = None,
     rows: Union[int, List[int], None] = None,
+    is_substitution=False,
```
machow, line 67: I added this flag mostly because it's very fast, but am down for creating a dedicated `sub()` function, if it seems better!

rich-iannone: What you have makes a lot of sense! By way of comparison, the R version has the analogous `subst()` function but it's not exported.

rich-iannone:
From a quick test of `sub_missing()`, something worth fixing is allowing the `missing_text=` to work with the `md()` and `html()` helper functions. This currently fails:

```python
from great_tables import GT, md, exibble

GT(exibble).sub_missing(missing_text=md("*MISSING*"))
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File [~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:344](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:344), in BaseFormatter.__call__(self, obj)
    [342](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:342)     method = get_real_method(obj, self.print_method)
    [343](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:343)     if method is not None:
--> [344](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:344)         return method()
    [345](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:345)     return None
    [346](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:346) else:

File [~/py_projects/great-tables/great_tables/gt.py:195](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:195), in GT._repr_html_(self)
    [194](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:194) def _repr_html_(self):
--> [195](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:195)     return self.render(context="html")

File [~/py_projects/great-tables/great_tables/gt.py:298](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:298), in GT.render(self, context)
    [297](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:297) def render(self, context: str) -> str:
--> [298](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:298)     html_table = self._build_data(context=context)._render_as_html()
    [299](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:299)     return html_table

File [~/py_projects/great-tables/great_tables/gt.py:277](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:277), in GT._build_data(self, context)
    [274](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:274) def _build_data(self, context: str) -> Self:
    [275](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:275)     # Build the body of the table by generating a dictionary
    [276](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:276)     # of lists with cells initially set to nan values
--> [277](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:277)     built = self._render_formats(context)
    [278](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:278)     # built._body = _migrate_unformatted_to_output(body)
    [279](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:279) 
...
--> [421](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/pandas/core/arrays/string_.py:421) if len(value) and not lib.is_string_array(value, skipna=True):
    [422](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/pandas/core/arrays/string_.py:422)     raise TypeError("Must provide strings.")
    [424](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/pandas/core/arrays/string_.py:424) mask = isna(value)

TypeError: len() of unsized object
```

rich-iannone:
Having `"---"` as the default isn't ideal. What do you think of having `None` here as a default, then we'll use an em-dash in each output type (right now just `"&mdash"` for HTML) for that default case? Then we don't have to worry about an `AsIs()`-type helper function at all (like we discussed before).

machow:
Using None as the flag for using some default, such as an emdash, seems reasonable! I wonder if we could keep the surface area of Great Tables down, by avoiding automatic conversions of plaintext to other characters (e.g. `"---"` to emdash)?

In the future, some kind of enum might make it easy to separate things like raw `"---"` from emdash?

E.g.

```
from enum import Enum

class Chars(Enum):
    emdash = "emdash"
    endash = "endash"
```

(But I'm really spitballing here! May be some better way to represent? Or could people use unicode?!)

rich-iannone:
> Using None as the flag for using some default, such as an emdash, seems reasonable! I wonder if we could keep the surface area of Great Tables down, by avoiding automatic conversions of plaintext to other characters (e.g. `"---"` to emdash)?

I’m proposing skipping that conversion altogether. The default emdash is enough for this purpose. You’re right in that users could just supply Unicode characters for virtually anything here!

PR #245, opened 2024-03-15
Add examples for `fmt_nanoplot()`

rich-iannone:
This adds some basic examples for the `fmt_nanoplot()` method.


machow (approved):
LGTM! One quick thought is that the description above the examples is quite long, and verbally expresses information that is quickly communicated through examples. I wonder if it's worth shortening the description? This would let people to get to examples quickly.

(In either event feel free to merge!)

rich-iannone:
Thanks for reviewing this so quickly! I've reduced the length of the Details section. You're right it was way too much (plus we have that dedicated spot in the Get Started guide for all this).

PR #247, opened 2024-03-18
Add the v0.4.0 release post

rich-iannone:
This adds the blog post for the `v0.4.0` release of the package.

### DIFF
--- a/docs/blog/introduction-0.4.0/index.qmd
+++ b/docs/blog/introduction-0.4.0/index.qmd
@@ -9,11 +9,11 @@ jupyter: python3
 
 The recent `v0.4.0` release of **Great Tables** contains nanoplots as a major new feature. So, in this post I'll concentrate on showing you all the things you can do with nanoplots. While the emphasis here will be on the aforementioned nanoplots feature, the last two releases (`v0.3.1` and `v0.4.0`) overall gave us a nice collection of improvements which includes:
 
-- the `fmt_nanoplot()` method for adding nanoplots to your table
+- the [`fmt_nanoplot()`](`great_tables.GT.fmt_nanoplot`) method for adding nanoplots to your table
 - improved HTML table representations in different code environments
 - integration of Polars selectors in the `columns=` arg of all formatting (`fmt_*()`) methods
-- the `save()` method for saving a **GT** table as an image file
-- rendering a **GT** table as an HTML string though `as_raw_html()`
+- the [`save()`](`great_tables.GT.save`) method for saving a **GT** table as an image file
+- rendering a **GT** table as an HTML string though [`as_raw_html()`](`great_tables.GT.as_raw_html`)
 
 Now let's dive into the wonderful world of nanoplots!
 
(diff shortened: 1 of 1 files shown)


machow:
RE this funky quarto behavior:

image

I wonder if updating quarto could make it go away? Otherwise, @cscheid might have an idea (here's the [page with guts out](https://pr-247--gt-python.netlify.app/blog/introduction-0.4.0/#nanoplots-small-interactive-plots-in-your-table))

rich-iannone:
I see that, but I wonder what's making it happen? I've never had that happen before. Will update my local `quarto-cli` install and test it there.

rich-iannone:
Found the problem, the code block was accidentally indented!

cscheid:
> Found the problem, the code block was accidentally indented!

(just a reminder that in Markdown, 4 spaces is the same as triple ticks)

cscheid:
(Also, wow does the great table website look great or what?)

docs/blog/introduction-0.4.0/index.qmd
```diff
@@ -0,0 +1,137 @@
...
+random_numbers_df = pl.DataFrame(
+    {
+        "i": range(1, 5),
+        "lines": [
+            { "val": [20.0, 23.0, 6.0, 7.0, 37.0, 23.0, 21.0, 4.0, 7.0, 16.0] },
+            { "val": [2.3, 6.8, 9.2, 2.42, 3.5, 12.1, 5.3, 3.6, 7.2, 3.74] },
+            { "val": [-12.0, -5.0, 6.0, 3.7, 0.0, 8.0, -7.4] },
+            { "val": [2.0, 0.0, 15.0, 7.0, 8.0, 10.0, 1.0, 24.0, 17.0, 13.0, 6.0] },
+        ],
+    }
```
machow, line 61: Having this data created above in pandas and then recreated in polars is a lot of reading. What about creating it once and using it across the two examples?

Whichever you choose, you might mention briefly how the input is being used (e.g. a string is being split into a sequence of data; or the first entry of the struct "val" is used), etc..

docs/blog/introduction-0.4.0/index.qmd
```diff
@@ -0,0 +1,137 @@
...
+
+The recent `v0.4.0` release of **Great Tables** contains nanoplots as a major new feature. So, in this post I'll concentrate on showing you all the things you can do with nanoplots. While the emphasis here will be on the aforementioned nanoplots feature, the last two releases (`v0.3.1` and `v0.4.0`) overall gave us a nice collection of improvements which includes:
+
+- the [`fmt_nanoplot()`](`great_tables.GT.fmt_nanoplot`) method for adding nanoplots to your table
+- improved HTML table representations in different code environments
+- integration of Polars selectors in the `columns=` arg of all formatting (`fmt_*()`) methods
+- the [`save()`](`great_tables.GT.save`) method for saving a **GT** table as an image file
+- rendering a **GT** table as an HTML string though [`as_raw_html()`](`great_tables.GT.as_raw_html`)
+
+Now let's dive into the wonderful world of nanoplots!
```
machow, line 18: It would be nice to show a nanoplot here without code (or expandable code). That way, people can see the result right away, before putting effort into reading code.

Is there a good nanoplot example we can use to make a splash (e.g. something that might also live on the examples page)?

machow (approved):
Feel free to take or punt the feedback :p. I think my biggest suggestion is a holistic example (with code hidden) up top

PR #252, opened 2024-03-21
Remove pandas dependencies

deanm0000:
This is a WIP of removing pandas as hard dependency. I started off thinking I just needed to remove it from nanoplots and then I kept looking for more places.

One annoying one, in particular, is the mizani import which itself requires pandas and is only used for a gradient calc. For this draft I just made it a try/except import.

I eventually lost steam on this so it's incomplete. My "workflow" on this is to

1. build it
2. install it
3. uninstall pandas
4. try to import it 
5. see where it errors for lack of pandas.
6. make it tolerant of polars and pandas
7. Rinse and Repeat

### DIFF
--- a/great_tables/_boxhead.py
+++ b/great_tables/_boxhead.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import pandas as pd
+
 from typing import TYPE_CHECKING, Optional, Any
 from typing_extensions import Self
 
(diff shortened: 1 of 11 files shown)


machow:
Hey, thanks for working on this. Like you flagged, taking out pandas ended up needing quite a bit of work across the code base. I've opened a PR to remove it from everything except the data submodule and `data_color()` (and will open a PR for `data_color()` next!).

I'm going to close this, in favor of that PR. Thanks again for looking into the issue!

https://github.com/posit-dev/great-tables/pull/261

PR #254, opened 2024-03-23
Add docs site and codecov badges

rich-iannone:
This adds documentation site and codecov badges to the README.md file.


machow (approved):
Nice!

PR #258, opened 2024-03-31
Fix error from incorrectly passing nanoplot options to args in `_generate_nanoplot()` 

rich-iannone:
In a previous PR, a call to `_generate_nanoplot()` was simplified to use `**` to pass through args: 

https://github.com/posit-dev/great-tables/blob/cd618a4111b0f5206e412b522e2c50c06b2c9e8d/great_tables/_formats.py#L3832

However, the argument names in three cases don't match. These are:

- `line_type` != `data_line_type`
- `show_ref_line` != `show_reference_line`
- `show_ref_area` != `show_reference_area`

The solution here is to expand the shorter arg names to their longer counterparts. This doesn't break the GT API since the longer names are what the user has access to within `nanoplot_options()`.

With this change, examples in https://posit-dev.github.io/great-tables/blog/introduction-0.4.0/ should no longer be broken.

### DIFF
--- a/great_tables/_utils_nanoplots.py
+++ b/great_tables/_utils_nanoplots.py
@@ -514,8 +514,8 @@ def _construct_nanoplot_svg(
     show_data_points: bool,
     show_data_line: bool,
     show_data_area: bool,
-    show_ref_line: bool,
-    show_ref_area: bool,
+    show_reference_line: bool,
+    show_reference_area: bool,
     show_vertical_guides: bool,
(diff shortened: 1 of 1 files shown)


machow (approved):
Shoot--thanks for fixing, and sorry for sacking all nanoplots with my clean up 😓. I'll merge now, since it's easy to confirm in the generated docs it's working, but let's be sure to add some basic fmt_nanoplot tests that, so the tests flag these situations

PR #260, opened 2024-04-02
Add several tests for different formatting methods

rich-iannone:
Quite a few of the formatting methods had very low test coverage so this PR provides some testing for the following:

- [x] `fmt_roman()`
- [x] `fmt_bytes()`
- [x] `fmt_time()`
- [x] `fmt_datetime()`
- [x] `fmt_markdown()`
- [x] `fmt_nanoplot()`


tests/test_formats.py
```diff
@@ -1060,6 +1062,511 @@ def test_format_number_with_sep_dec_marks():
...
+    assert x == x_out
+
+
+# ------------------------------------------------------------------------------
+# Test `fmt_time()`
+# ------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "time_style,x_out",
```
machow, line 1156: Since each case is run against the same 7 input times, it's not clear which of these inputs tests something new per timestyle, and which is redundant.

Rather than each case involving a time_style and a the same list of 7 inputs, can you make each case a time_style, 1 input, and the output expected? A case should only be added if it tests some new bit underlying logic.

tests/test_formats.py
```diff
@@ -1060,6 +1062,511 @@ def test_format_number_with_sep_dec_marks():
...
+    x = _get_column_of_values(gt, column_name="x", context="html")
+    assert x == x_out
+
+
+# ------------------------------------------------------------------------------
+# Test `fmt_datetime()`
+# ------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
```
machow, line 1207: Same feedback as above, on splitting into single input cases. It appears the underlying logic is largely babel's format_datetime, so you don't need to re-test that (but it makes sense there's a lot of stitching of arguments to fmt_datetime that need to get tested).

tests/test_formats.py
```diff
@@ -1060,6 +1062,511 @@ def test_format_number_with_sep_dec_marks():
...
+    assert x == x_out
+
+
+# ------------------------------------------------------------------------------
+# Test `fmt_bytes()`
+# ------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "standard,decimals,use_seps,force_sign,incl_space,x_out",
```
machow, line 1266: See feedback on avoiding using many parameters, and resetting things to their default arguments.

See also feedback on using 1 long input across all cases. I can't tell which values in the output tell us something new. For example all the outputs contain "-342 B", so it's not clear what I should learn there across the 5 parameter set cases.

(Sometimes tests do need to validate that some result is unchanged, but it seems likely the testing here is maybe doing that repeatedly for the same underlying logic)

tests/test_formats.py
```diff
@@ -1060,6 +1062,511 @@ def test_format_number_with_sep_dec_marks():
...
+                "+2.2GiB",
+                "+769.1TiB",
+                "+225.2PiB",
+                "+60YiB",
+                "+746519.9YiB",
+            ],
+        ),
+    ],
+)
+def test_fmt_bytes(
```
machow, line 1395: See feedback on parametrizing and not using the same inputs for everything

tests/test_formats.py
```diff
@@ -1060,6 +1062,511 @@ def test_format_number_with_sep_dec_marks():
...
+    x = _get_column_of_values(gt, column_name="x", context="html")
+    assert x == x_out
+
+
+# ------------------------------------------------------------------------------
+# Test `fmt_roman()`
+# ------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
```
machow, line 1438: I strongly suspect you don't need to test that upper works on many variations of inputs values. (one is probably fine).

tests/test_formats.py
```diff
@@ -1179,3 +1686,243 @@ def test_fmt_image_path():
...
+# ------------------------------------------------------------------------------
+
+
+def _all_vals_nanoplot_output(nanoplot_str: List[str]):
+    import re
+
+    return all(bool(re.match("^<div><svg.*</svg></div>$", x)) for x in nanoplot_str)
+
+
+@pytest.mark.parametrize(
```
machow, line 1702: It looks like many of the tests below...

* test across a wide range of inputs (10 rows in a DataFrame)
* only assert that the nanoplot exists with a div and svg tag

Please re-write so that it clear what each cases is testing, and their is a more specific assertion. For example, if reference line is "mean", does that mean the line is the average across rows? If so, in order to write these tests, you may need to reduce the number of input rows (e.g. to two), and test the exact value of the reference line in resulting nanoplots.

tests/test_formats.py
```diff
@@ -1179,3 +1686,243 @@ def test_fmt_image_path():
...
+    gt = GT(df).fmt_nanoplot(
+        columns="vals",
+        **params,
+    )
+    res = _get_column_of_values(gt, column_name="vals", context="html")
+
+    assert _all_vals_nanoplot_output(res)
+
+
+@pytest.mark.parametrize(
```
machow, line 1771: See above feedback. It's not clear what specific behaviors are being tested here. There are likely too many inputs, and the assertions are not specific enough.

It's okay to include the inputs in the parameters, if each case requires it's own (minimal) set of inputs to be meaningful. If the specification of arguments, inputs, and results checks are two cumbersome for `@pytest.mark.parametrize`, then it may be a signal to put each in its own function.

tests/test_formats.py
```diff
@@ -1179,3 +1686,243 @@ def test_fmt_image_path():
...
+    gt_2 = GT(df_2).fmt_nanoplot(
+        columns="vals",
+        **params,
+    )
+    res_2 = _get_column_of_values(gt_2, column_name="vals", context="html")
+
+    assert _all_vals_nanoplot_output(res_2)
+
+
+@pytest.mark.parametrize(
```
machow, line 1868: See above feedback

tests/test_formats.py
machow: If some of these parameters end up being tested more thoroughly in another place (e.g. `_generate_nanoplots`), then we should think about what parameters need to be tested at this level and how thoroughly (based on the underlying logic in the fmt_nanoplot function).

machow (changes requested):
Thanks for working on upping our test game. I left some feedback on...

* reducing the number of parameters set in `@pytest.mark.parametrize` (which were resulting in things having to be constantly reset back to defaults)
* leaning away from using the same long list of inputs across many case, in favor of individual inputs which test new & distinct bits of underlying logic
* providing more informative assertions for fmt_nanoplot, and using fewer inputs (or clarifying what we're testing their above and beyond what's tested for `_generate_nanoplot`).

tests/test_formats.py
```diff
@@ -1060,6 +1062,511 @@ def test_format_number_with_sep_dec_marks():
     assert x == ["12.345.678,12346", "1,00000", "0,00000", "\u2212" + "12.345.678,12346"]
 
 
+# ------------------------------------------------------------------------------
+# Test `fmt_currency()`
+# ------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "currency,use_subunits,decimals,drop_trailing_dec_mark,use_seps,placement,incl_space,x_out",
```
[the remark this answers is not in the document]
rich-iannone: Now done. Thanks!

rich-iannone:
I believe all the comments have been addressed through the changes we made during code pairing. I handled a merge conflict and reformatted to satisfy the pre-commit error.

machow (approved):
LGTM, thanks for spending time to up the coverage here :)

PR #261, opened 2024-04-02
Refactor rm pandas

machow:
This PR removes pandas as a dependency, making it optional. (Great Tables will still support pandas DataFrames as an input!)

Notes:

* It will leave pandas as a dependency for import `great_tables.data` for now.
* `great_tables.exibble` is a pandas DataFrame. We'll use a dynamic getattr module method to error only when `great_tables.exibble` is accessed.

Details:

* Replaced pkg_resources with importlib_resources (see #270)
* Locale data now loaded using builtin csv library
* babel used to validate locales we don't recognize
* added a CI job that runs `tests/test_dependencies.py` w/o pandas installed
* I left in numpy for now, since it's used in a couple places
  - nanoplot aggregation functions
  - in some places to mark missing values
* Added some tests of format utility functions (to make sure I didn't break anything!)
* There's some dynamic magic going on in our `great_tables/__init__.py` to ensure people can still import exibble.
* Set mizani as an optional dependency (in order to use data_color). Per [discussion here](https://github.com/posit-dev/great-tables/pull/261#issuecomment-2040297010), I'll open a new PR to take it out.


Yasin197:
Is this PR related to removing pandas as a hard dependency?

machow:
> Is this PR related to removing pandas as a hard dependency?

Yes! I'm going to have this ready for review by EOW 🤞🏼

Yasin197:
The mizani package dependency forces pandas in great-tables to be >=2.1.0?

Would this mean even though you are making pandas not a hard dependency, the mizani package will still exhibit this?

machow:
> The mizani package dependency forces pandas in great-tables to be >=2.1.0?

Shoot, I forgot about this piece. I'll restrict this PR to removing direct pandas imports from our codebase, and adding safeguards against people accidentally adding them in the future. We've talked through our use of mizani's `gradient_n_pal()`. We should be able to roll a stripped down version of that logic into Great Tables for now (but will do that in a separate PR).

Yasin197:
Thanks, this new PR will still be apart of your upcoming release right? It would make great-tables more pleasant to work with.

.github/scripts/no_pandas.py
```diff
@@ -0,0 +1,72 @@
+# %%
```
machow, line 1: This script runs in the code-check CI and does a static check that pandas imports aren't used.

see https://github.com/posit-dev/great-tables/pull/261/files#diff-38696fb127ded399e74cd12e7529270058aa86df58256e55b02c5b382adf908aR31

.github/workflows/code-checks.yaml
```diff
@@ -18,3 +18,14 @@ jobs:
...
+  optional-deps:
+    name: "Static check optional deps"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - run: |
+          python .github/scripts/no_pandas.py great_tables _tbl_data.py data/__init__.py
```
machow, line 31: We exempt..

* _tbl_data because it holds all pandas logic
* `data/__init__.py` so we can punt on it for now

.github/workflows/ci-tests.yaml
```diff
@@ -45,10 +40,25 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          pip install -e '.[dev]'
```
machow, line 53: This job is essentially the same as the one above, but we only install `dev` dependencies, rather than `all`

Makefile
```diff
@@ -3,6 +3,9 @@
 test:
 	pytest --cov=great_tables --cov-report=xml
 
+test-no-pandas:
+	pytest tests/test_dependencies.py -m "no_pandas"
```
machow, line 7: we only test this one file, because other tests do top-level imports of pandas. So if you don't have pandas installed, the tests would fail on collection.

great_tables/_boxhead.py
```diff
@@ -166,14 +165,3 @@ def cols_align(self: GTSelf, align: str = "left", columns: Optional[str] = None)
 
     # Set the alignment for each column
     return self._replace(_boxhead=self._boxhead._set_column_aligns(columns, align=align))
-
```
machow, line 169: unused code

great_tables/_data_color/base.py
```diff
@@ -167,6 +166,13 @@ def data_color(
     ```
     """
 
+    try:
```
machow, line 169: mizani moved inside function, so if you don't have it, it doesn't fail on import

great_tables/_formats.py
```diff
@@ -456,7 +459,7 @@ def fmt_integer_fn(
         scale_by: float = scale_by,
     ):
         # If the `x` value is a Pandas 'NA', then return the same value
-        if pd.isna(x):
+        if is_na(self._tbl_data, x):
```
machow, line 462: Replaced a bunch of pd.isna checks

great_tables/_formats.py
```diff
@@ -2568,17 +2571,22 @@ def _replace_minus(string: str, minus_mark: str) -> str:
     return _str_replace(string, "-", minus_mark)
 
 
-def _filter_pd_df_to_row(pd_df: pd.DataFrame, column: str, filter_expr: str) -> pd.DataFrame:
-    filtered_pd_df = pd_df[pd_df[column] == filter_expr]
+T_dict = TypeVar("T_dict", bound=TypedDict)
+
+
+# TODO: remove pandas
+def _filter_pd_df_to_row(pd_df: "list[T_dict]", column: str, filter_expr: str) -> T_dict:
```
machow, line 2578: All locale info went from being a DataFrame, to a list of dictionaries (i.e. list of rows). This function just got the row where some column was equal to something.

great_tables/_formats.py
```diff
@@ -2593,7 +2601,7 @@ def _get_locale_sep_mark(default: str, use_seps: bool, locale: Union[str, None]
     # the column named 'group'; this could potentially be of any type but we expect
     # it to be a string (and we'll check for that here)
     sep_mark: Any
-    sep_mark = pd_df_row.iloc[0]["group"]
+    sep_mark = pd_df_row["group"]
```
machow, line 2604: After filtering DataFrames to be a 1 row DataFrame, a specific value was peeled off. Now it's just pulled out of the row dictionary.

great_tables/_formats.py
```diff
@@ -2632,39 +2642,18 @@ def _get_locales_list() -> List[str]:
     """
 
     # Get the 'locales' dataset and obtain from that a list of locales
+    # TODO: remove pandas
     locales = _get_locales_data()
-    locale_list = locales["locale"].tolist()
+    locale_list = [entry["locale"] for entry in locales]
```
machow, line 2647: Removed misc pandas Series methods (mostly iloc and tolist())

great_tables/_formats.py
```diff
@@ -2682,14 +2671,16 @@ def _validate_locale(locale: Union[str, None] = None) -> None:
...
+    default_locales_list = [entry["default_locale"] for entry in _get_default_locales_data()]
 
     # Replace any underscores with hyphens
     supplied_locale = _str_replace(locale, "_", "-")
 
     # Stop if the `locale` provided isn't a valid one
     if supplied_locale not in locales_list and supplied_locale not in default_locales_list:
-        raise ValueError("The supplied `locale` is not available in the list of supported locales.")
+        raise ValueError(
+            f"The normalized locale name `{supplied_locale}` is not in the list of locales."
```
machow, line 2682: showing locale value

great_tables/_formats.py
```diff
@@ -2716,22 +2707,28 @@ def _normalize_locale(locale: Union[str, None] = None) -> Union[str, None]:
...
+    matches = [
+        entry["base_locale"]
+        for entry in default_locales
+        if entry["default_locale"] == supplied_locale
+    ]
+
+    if matches:
+        return matches[0]
+
+    try:
```
machow, line 2722: This is a new behavior. Before it allowed unrecognized to pass through. I added some babel validation, to try and only let the subset of things through that:

* aren't in our locale data
* will end up working with babel

(but could have gotten it wrong!)

great_tables/_formats.py
```diff
@@ -3713,7 +3707,7 @@ def fmt_nanoplot(
         # Why are we grabbing the first element of a tuple? (Note this also happens again below.)
         all_single_y_vals = list(
             data_tbl[columns].apply(
```
machow, line 3709: We should maybe take this `.apply` out at some point, but because pandas and polars implement it in a similar what it works okay 😁

great_tables/_formats_vals.py
```diff
@@ -1,14 +1,16 @@
...
 from great_tables import GT
 from typing import List, Any, Union, Optional, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from ._formats import DateStyle, TimeStyle
+    from ._tbl_data import SeriesLike
 
 
-def _make_one_col_table(vals: Union[Any, List[Any]]) -> GTData:
+def _make_one_col_table(vals: Union[Any, List[Any], SeriesLike]) -> GT:
```
machow, line 13: One challenge here is that, if a user passes a list, then we have to use a pandas Series on it.

By adding SeriesLike, it lets users pass their own series (pandas or polars), so we don't have to depend on a specific one. 

See #266

great_tables/_locale.py
```diff
@@ -1,7 +1,14 @@
 from __future__ import annotations
-from typing import Optional
-import pandas as pd
-import pkg_resources
+
+from csv import DictReader
```
machow, line 3: Replaced all pandas DataFrames with the built in python DictReader. This returns a list of dictionaries (where each dictionary is a row)

great_tables/_tbl_data.py
```diff
@@ -545,3 +556,38 @@ def _(df: PdDataFrame) -> PdDataFrame:
 @validate_frame.register
 def _(df: PlDataFrame) -> PlDataFrame:
     return df
+
+
+# to_frame ----
+
+
+@singledispatch
+def to_frame(ser: "list[Any] | SeriesLike", name: Optional[str] = None) -> DataFrameLike:
```
machow, line 565: This supports the `vals.fmt_*` functions, by converting a list or series into a DataFrame (which can be fed into GT)

great_tables/_utils_nanoplots.py
```diff
@@ -1,11 +1,22 @@
...
-import pandas as pd
 import numpy as np
+
 from great_tables._utils import _match_arg
+from great_tables._tbl_data import Agnostic, is_na
 
 REFERENCE_LINE_KEYWORDS = ["mean", "median", "min", "max", "q1", "q3"]
 
 
+def _is_na(x: Any) -> bool:
```
machow, line 10: This runs `is_na()` in a DataFrame agnostic way throughout the module

great_tables/data/__init__.py
```diff
@@ -1,9 +1,17 @@
-import pandas as pd
-import pkg_resources
+from importlib_resources import files
 
-DATA_MOD = "great_tables.data"
+try:
```
machow, line 3: Give a friendly message when users don't have pandas, so raise an error trying to import data.

rich-iannone (approved):
This is all really good! I checked the rendered Quarto site and no tables were affected.

PR #262, opened 2024-04-02
Include choice of webdrivers in the `.save()` method

rich-iannone:
Previously, using `.save()` to capture a table image as a file required a local install of Chrome/Chromium on the host system. While this is a common browser to have, it can be constraining if Chrome is not available (but other common browsers are). In this PR, we increase this to four browsers:

- chrome
- safari
- firefox
- edge

This is exposed as the `wd=` argument and the default is `"chrome"` (as before). In my testing on Mac, all of them work very well. There is only minimal differences in rendering and quality is basically the same across the different scaling factors. The cropping works equally well across the four webdrivers tested.

Safari does require that you 'Allow remote automation' in the Develop settings (a warning *will* be emitted in the console to this effect). Safari doesn't seem to have headless mode, set that command isn't set. Firefox will appear on the screen during capture (but it does shut down gracefully). Safari does something similar but it steals focus on my system.


Yasin197:
Will this work in a Linux environment?

rich-iannone:
> Will this work in a Linux environment?

It should but we'd need to test it. We won't merge until it is tested with Chrome, Firefox, and Edge in some Linuxes.

rich-iannone:
In testing on Mac, here is the code used (to create individual files, changing `driver` to modify the webdriver used):

```python

import great_tables as gt
from great_tables.data import sp500

start_date = "2010-06-07"
end_date = "2010-06-14"

sp500_mini = sp500[(sp500["date"] >= start_date) & (sp500["date"] <= end_date)]

driver="chrome"

(
    gt.GT(data=sp500_mini)
    .tab_header(title="S&P 500", subtitle=f"{start_date} to {end_date}")
    .fmt_currency(columns=["open", "high", "low", "close"])
    .fmt_date(columns="date", date_style="wd_m_day_year")
    .fmt_number(columns="volume", compact=True)
    .cols_hide(columns="adj_close")
    .save(f"test-{driver}.png", scale=1, wd=driver)
)
```

Here are the results on Mac.

chrome:
test-chrome

safari:
test-safari

firefox:
test-firefox

edge:
test-edge

machow (approved):
WDYT of naming the `wd` argument to `web_driver`? or `driver`? The name is a bit shorter than `window_size`, so it might be nice to expand for consistency.

(I'm not wed to any particular name here, so feel free to keep as is and merge!)

rich-iannone:
I like `"driver"` and totally agree that the current name is too short and non-descriptive. Will make that change soon!

Issue #7, opened 2023-10-10
epic: make pandas an optional dependency


machow:
**Edit:** polars support is complete! This issue now tracks only the removing of pandas as a mandatory dependency. This is largely a code cleanup activity, since the machinery to make pandas optional is already in great tables.

~This may take a bit of work, but using the approach in https://github.com/machow/databackend, we should be able to make pandas and polars optional dependencies.~

~There is a data decorator pattern used in `_tbl_data.py` that wraps a dataframe to provide standard methods (e.g. for getting column names). We should be able to use generic functions, so gt can call things like `get_column_names` directly on any DataFrame.~

johnros:
In the meanwhile, worth updating the DT() documentation clarifying that `dataframe` refers to `pd.Dataframe()`.

machow:
Polars is supported (though pandas is still an optional dependency)! I'll update this issue to reflect that the remaining thing is making pandas optional.

PR #271, opened 2024-04-09
feat: remove mizani as dependency, re-implement logic internally

machow:
This PR removes mizani as a dependency, so that we don't transitively depend on scipy and pandas. Note that our internal implementation does not depend on numpy, so that we can drop it as a dependency in a later PR.

This implementation is not as elegant as mizani's, or quite as optimized, but should do well for our purposes.

## Speed

Our default implementation was about 50% slower than mizani in a simple test. But both are very fast for simple table displays (1.52 ms for 1,000 points for ours, 1ms for mizani.)

```python
from great_tables._data_color.palettes import GradientPalette, CoeffSequence
from mizani.palettes import gradient_n_pal

palette = GradientPalette(["red", "orange", "blue", "grey", "yellow", "red", "orange"])
palette2 = GradientPalette(["red", "orange", "blue", "grey", "yellow", "red", "orange"], cls_coeff_sequence=CoeffSequence)
palette3 = gradient_n_pal(["red", "orange", "blue", "grey", "yellow", "red", "orange"])


%%timeit
# internal + bisect lookup: 1.52 ms ± 5.85 µs per loop
palette([x / 1000. for x in range(1000)])

%%timeit
# 1.76 ms ± 6.27 µs per loop
palette2([x / 1000. for x in range(1000)])

%%timeit
# mizani: 1.03 ms ± 9.82 µs per loop
palette3([x / 1000. for x in range(1000)])
```

The main difference is that we're using a simple bisect lookup to find cutoff corresponding to a value (in order to get coefficients for transforming within a cutoff band). Mizani uses a table lookup, which cuts the input/response space into 256 bins.

Fixes: https://github.com/posit-dev/great-tables/issues/7

### DIFF
--- a/tests/data_color/__snapshots__/test_data_color.ambr
+++ b/tests/data_color/__snapshots__/test_data_color.ambr
@@ -137,8 +137,8 @@
     <td class="gt_row gt_left">two</td>
   </tr>
   <tr>
-    <td style="color: #000000; background-color: #73c476;" class="gt_row gt_right">3</td>
-    <td style="color: #000000; background-color: #73c476;" class="gt_row gt_right">8</td>
+    <td style="color: #000000; background-color: #74c476;" class="gt_row gt_right">3</td>
+    <td style="color: #000000; background-color: #74c476;" class="gt_row gt_right">8</td>
     <td class="gt_row gt_left">three</td>
   </tr>
   <tr>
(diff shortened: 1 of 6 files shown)


tests/data_color/__snapshots__/test_data_color.ambr
machow, line 1: AFAICT changes to snapshots are due to differences in rounding (and/or the resolution of mizani's table lookup).

rich-iannone:
Thank you for the huge amount of work you put into this and the related PR! Will review shortly.

rich-iannone (approved):
LGTM!

machow:
@abstractqqq in case it's useful for polars_ds, this PR should remove pandas as a dependency (except for importing from `great_tables.data`, which we'll tackle later!). We should be able to release soon. Definitely let us know if you run into any issues with it.

Issue #175, opened 2024-02-06
Add `opt_table_font()`


machow:
Should be a quick one

https://gt.rstudio.com/reference/opt_table_font.html

machow:
If it's okay, I'm moving this back to the backlog. Please finish out existing issues that have been in TODO for a while, like https://github.com/posit-dev/great-tables/issues/142, https://github.com/posit-dev/great-tables/issues/227, #196

PR #272, opened 2024-04-10
Add the `opt_table_font()` method

rich-iannone:
This adds the `opt_table_font()` method, which can be used for the easy addition of fonts (or a font stack) to the table. Here's an example where a font stack was used:

```python

gt_ex = (
    GT(exibble.head(5), rowname_col="row", groupname_col="group")
    .tab_header("THE HEADING", "(a subtitle)")
    .tab_stubhead("THE STUBHEAD")
    .tab_source_note("THE SOURCE NOTE")
    .opt_table_font(stack="industrial")
)

gt_ex
```
opt_table_font_font_stack

Here's another example where a single font name is prepended to the list of default fonts:

```python

gt_ex = (
    GT(exibble.head(5), rowname_col="row", groupname_col="group")
    .tab_header("THE HEADING", "(a subtitle)")
    .tab_stubhead("THE STUBHEAD")
    .tab_source_note("THE SOURCE NOTE")
    .opt_table_font(font="Comic Sans MS")
)

gt_ex
```

opt_table_font_prepend_font

Fixes: https://github.com/posit-dev/great-tables/issues/175


machow (changes requested):
Hey, thanks for opening. What's the status of #262 and #260? I think we should pause this, in order to finish out prior work.

rich-iannone:
Sounds good. PR #260 is ready for a re-review.

great_tables/_options.py
```diff
@@ -1054,6 +1055,196 @@ def opt_table_outline(
...
+    else:
+
+        # Case where only `font=` is provided and that value is a list of fonts which may be
+        # prepended to the existing fonts or replaced entirely (depending on the `add=` value)
+        if add:
+            combined_fonts = font + existing_fonts
+            res = tab_options(res, table_font_names=combined_fonts)
+
+        else:
+            res = tab_options(res, table_font_names=font)
```
machow, line 1231: It looks like..

* all 3 branches are about setting table_font_names
* 2 of the 3 branches are focused on creating a `combined_fonts` object

Could we try shifting the focus to creating `combined_fonts`, and then setting the table_font_names option after (vs focusing on setting the option)? Something like this...

```python
   if stack is not None:
    
        # Case where value is given to `stack=` and this is a keyword that returns a
        # list of fonts (i.e., the font stack); in this case we combine with `font=` values
        # (if provided) and we *always* replace the existing fonts (`add=` is ignored)
        from great_tables._helpers import system_fonts

        font_stack = system_fonts(name=stack)
        combined_fonts = font + font_stack
    elif add:
        # Case where `font=` is prepended to existing fonts
        combined_fonts = font + existing_fonts
    else:
        # Case where  `font=` replacing existing fonts
        combined_fonts = font
            
    res = tab_options(res, table_font_names=font)
```

Note two potentially useful pieces:

* the big if/elif/else supports the idea in the comment that if stack is set, then add is ignored.
* The comment in the previous else had two conditions ("depending on..."), and here is split into two (i.e. this is split up: `"may be prepended to the existing fonts or replaced entirely (depending on the add= value)"`)

machow (changes requested):
This is looking good! I added two quick comments on code tweaks, but feel free to push back on anything

great_tables/_options.py
```diff
@@ -1054,6 +1055,196 @@ def opt_table_outline(
...
+
+    if weight is not None:
+
+        if isinstance(weight, int) or isinstance(weight, float):
```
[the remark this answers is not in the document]
rich-iannone: Updated so typing of any `*weight=` arg is that of `int`, `float`, or `str`. Now raising a TypeError for other types passed in.

great_tables/_options.py
rich-iannone: Thank you for the recommendation and the detailed explanations here (really helps!). I've made the change so that the conditionals are as above.

great_tables/_options.py
```diff
@@ -1054,6 +1055,202 @@ def opt_table_outline(
...
+        raise ValueError("The value from `_options.table_font_names` must be a list.")
+
+    res = self
+
+    # If `stack` is provided, we ignore the `add` argument and always replace the existing fonts;
+    # This is because users are likely to want to use the font stack as a whole the set is safe
+    # enough to use across different systems (so keeping fallbacks usually won't be seen as
+    # necessary)
+    if stack is not None:
+        add = False
```
machow, line 1199: afaict this is uneccessary (because of the if/elif)

great_tables/_options.py
```diff
@@ -1054,6 +1055,202 @@ def opt_table_outline(
...
+        # Case where  `font=` replacing existing fonts
+        combined_fonts = font
+
+    res = tab_options(res, table_font_names=combined_fonts)
+
+    if weight is not None:
+
+        if isinstance(weight, int) or isinstance(weight, float):
+
+            if weight < 1 or weight > 1000:
```
machow, line 1232: I would remove this validation, since it's only validating weight when it's an int or float. If a user passes a str weight "1001" it will be accepted.

machow (approved):
Thanks for this beautiful PR -- I added a couple small suggestions. If they sound okay, feel free to merge after changing (or punt the suggestions and merge :)

rich-iannone:
I've incorporated your suggestions in two commits. If CI passes I'll merge this PR. Thanks!

Issue #241, opened 2024-03-12
Remove mentions of 'Accounting' format in docs


Yasin197:
In the docs it mentions numbers will be formatted if accounting=True.

So I would expect negative numbers would be represented in accounting terms e.g. a cell with value of 2 would be represented as (2)

I don't see a clean example on how to use this API?

rich-iannone:
Thanks for reporting this. I had a look at the documentation and this appears to be a mistake. There is the repeated line:

`"This option is disregarded when using notation with 'accounting = True'"`

We don't yet have accounting notation implemented in the relevant numeric formatting methods.

I think the best course of action right now is to remove all mention of this in the docs and create an issue that tracks the implementation of accounting notation.

PR #276, opened 2024-04-11
fix: issues with save function

rich-iannone:
This adds a workflow job that tests whether the `.save()` method is properly producing a table image when using the `"firefox"` webdriver.

Also fixes: https://github.com/posit-dev/great-tables/issues/241

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -89,6 +89,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -e '.[extra]'
+          pip install pandas
       - name: pytest unit tests
         run: |
           make save-browser-table
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -74,3 +74,21 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
+
+  browser-save-image-test:
+    name: "Test for saving table image with browser"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: browser-actions/setup-firefox@v1
+      - run: firefox --version
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          pip install -e '.[extra]'
+      - name: pytest unit tests
+        run: |
+          make test-no-pandas
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -91,4 +91,10 @@ jobs:
           pip install -e '.[extra]'
       - name: pytest unit tests
         run: |
-          make test-no-pandas
+          make save-browser-table
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: browser-save-image
+          path: exibble_firefox.png
+          if-no-files-found: error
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-tests.yaml
```diff
@@ -74,3 +74,27 @@ jobs:
...
+      - uses: browser-actions/setup-firefox@v1
+      - run: firefox --version
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          pip install -e '.[extra]'
```
machow, line 91: I would just run `pip install pandas` for now after this, since we need it for the datasets

rich-iannone:
This fails with:

```
python .github/scripts/save_browser_table.py
Traceback (most recent call last):
  File "/home/runner/work/great-tables/great-tables/.github/scripts/save_browser_table.py", line 3, in <module>
    GT(exibble).save("exibble_firefox.png", web_driver="firefox")
  File "/home/runner/work/great-tables/great-tables/great_tables/_export.py", line 181, in save
    wdriver(options=wd_options) as headless_browser,
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/selenium/webdriver/firefox/webdriver.py", line 69, in __init__
    super().__init__(command_executor=executor, options=options)
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/selenium/webdriver/remote/webdriver.py", line 208, in __init__
    self.start_session(capabilities)
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/selenium/webdriver/remote/webdriver.py", line 292, in start_session
    response = self.execute(Command.NEW_SESSION, caps)["value"]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/selenium/webdriver/remote/webdriver.py", line 347, in execute
    self.error_handler.check_response(response)
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/selenium/webdriver/remote/errorhandler.py", line 229, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: Process unexpectedly closed with status 1

make: *** [Makefile:53: save-browser-table] Error 1
```

Probably because Firefox is run in non-headless mode: https://stackoverflow.com/questions/46809135/webdriver-exceptionprocess-unexpectedly-closed-with-status-1

Now, technically it runs in headless mode on my system but the resulting image is incorrect (only gives correct image output when *not* run in headless mode).

rich-iannone:
I suppose we could instead run with chrome using: https://github.com/browser-actions/setup-chrome.

machow:
Ah, dang--what does the image look like? It looks there are options to run FF headless in that SO post and [this one](https://stackoverflow.com/a/46768243/1144523), so I'm curious if there's a way to get it to work?!

rich-iannone:
The firefox screenshot looks like this in headless mode:

test-firefox

and otherwise just fine if we run w/o headless:

test-firefox_not_headless

I'll try out some of the things in that SO post to see if there's a way to work around this!

rich-iannone:
Seems like `"---headless=new"` solved the problem! See https://www.selenium.dev/blog/2023/headless-is-going-away/. I'll test this with the rest of the web drivers to ensure everything works with that.

rich-iannone:
With `"---headless=new"` all four browsers now work well in headless mode. Pushed the change.

machow:
- [ ]

PR #277, opened 2024-04-11
fix: remove uses of DataFrame.apply and dtype methods

machow:
This PR removes some lingering uses of DataFrame methods outside of our DataFrame specific `_tbl_data.py` submodule. It also restricts the `columns=` argument of `.fmt_nanoplot()` to be a string for now, as this seemed to be the only supported path.


rich-iannone (approved):
Looks good! And I verified that the warning no longer appears on the documentation site.

PR #281, opened 2024-04-12
fix: close some opened files

machow:
This PR fixes some small cases, where files were being opened outside of a context manager.


rich-iannone (approved):
LGTM!

PR #285, opened 2024-04-13
Resolve issue with `table_font_color` not accepting named colors

jrycw:
It appears we're having an issue with the [Table Theme Options -> Set options across table parts](https://posit-dev.github.io/great-tables/get-started/table-theme-options.html#set-options-across-table-parts) section. 

image

The problem seems to arise when color values are input using common names like `yellow` or `white`, causing the `font_color` not to transform them into the appropriate hexadecimal format.

To address this issue, I've implemented a conversion pattern similar to the one used in `font_color`, which should hopefully resolve the issue.

Additionally, I noticed quick returns in the first few lines of `font_color`. I'm uncertain whether these also need conversion. Could the team please review this aspect?

### DIFF
--- a/great_tables/_scss.py
+++ b/great_tables/_scss.py
@@ -39,13 +39,20 @@ FONT_COLOR_VARS = (
 )
 
 
-def font_color(color: str, table_font_color: str, table_font_color_light: str) -> str:
-    """Return either table_font_color or table_font_color_light, whichever is higher contrast."""
+def font_color(color: str, dark_option: str, light_option: str) -> str:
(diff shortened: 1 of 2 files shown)


machow:
Thanks so much for this work! I'm pairing right now with @rich-iannone, and he mentioned this function should always return hex (which it wasn't even before this PR). I've tweaked it a bit, so that when things like "transparent" are passed, hex colors are output.

jrycw:
The refactored code looks better. In fact, the variable names definitely represent their meaning clearly.

PR #288, opened 2024-04-14
Code refactoring in `_tbl_data`

jrycw:
Hello team,

I wanted to express my gratitude for creating this fantastic library. However, I believe there's always room for improvement. I've taken a look around the codebase and identified some areas where we could refactor or simplify certain logics. I'm curious to know if you would welcome this kind of PR?

### DIFF
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -249,17 +249,20 @@ def _(data: PdDataFrame, group_key: str) -> Dict[Any, List[int]]:
 
 @group_splits.register
 def _(data: PlDataFrame, group_key: str) -> Dict[Any, List[int]]:
-    # TODO: should ensure row count name isn't already in data
     import polars as pl
 
+    _row_count_name = "__row_count__"
+    if _row_count_name in data.columns:
+        raise ValueError(f"{_row_count_name} is already in data.columns")
+
     # with_row_index supercedes with_row_count
-    meth_row_number = getattr(data, "with_row_index", None)
-    if not meth_row_number:
-        meth_row_number = data.with_row_count
+    meth_row_number = getattr(data, "with_row_index", "with_row_count")
 
-    groups = meth_row_number("__row_count__").group_by(group_key).agg(pl.col("__row_count__"))
+    groups = (
+        meth_row_number(_row_count_name).group_by(group_key).agg(pl.col(_row_count_name)).to_dict()
+    )
 
-    res = dict(zip(groups[group_key].to_list(), groups["__row_count__"].to_list()))
+    res = dict(zip(groups[group_key], groups[_row_count_name]))
     return res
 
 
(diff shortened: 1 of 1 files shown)


great_tables/_tbl_data.py
```diff
@@ -249,17 +249,20 @@ def _(data: PdDataFrame, group_key: str) -> Dict[Any, List[int]]:
...
 
+    _row_count_name = "__row_count__"
+    if _row_count_name in data.columns:
+        raise ValueError(f"{_row_count_name} is already in data.columns")
+
     # with_row_index supercedes with_row_count
-    meth_row_number = getattr(data, "with_row_index", None)
-    if not meth_row_number:
-        meth_row_number = data.with_row_count
+    meth_row_number = getattr(data, "with_row_index", "with_row_count")
```
machow, line 259: The code will return the string "with_row_count", when it should be returning a .with_row_count method.

great_tables/_tbl_data.py
```diff
@@ -249,17 +249,20 @@ def _(data: PdDataFrame, group_key: str) -> Dict[Any, List[int]]:
...
-    # TODO: should ensure row count name isn't already in data
     import polars as pl
 
+    _row_count_name = "__row_count__"
+    if _row_count_name in data.columns:
+        raise ValueError(f"{_row_count_name} is already in data.columns")
+
     # with_row_index supercedes with_row_count
-    meth_row_number = getattr(data, "with_row_index", None)
-    if not meth_row_number:
```
machow, line 257: This is here because if the getattr fails, then we want to grab data.with_row_count. But if it succeeds, we do not want to try that (since the first succeeding could mean the second would fail, etc..)

machow:
Hey, thanks so much for all these PRs. This particular one is a bit tricky, since it needs to maintain backwards compatibility with earlier polars versions. I've left some comments, on how the code changes could hit issues (compared to the original code).

We're super excited for all your contributions, and appreciate you looking through. I'm going to close this specific PR, but just know that we really appreciate all these contributions you're making

great_tables/_tbl_data.py
```diff
@@ -249,17 +249,20 @@ def _(data: PdDataFrame, group_key: str) -> Dict[Any, List[int]]:
...
+    meth_row_number = getattr(data, "with_row_index", "with_row_count")
 
-    groups = meth_row_number("__row_count__").group_by(group_key).agg(pl.col("__row_count__"))
+    groups = (
+        meth_row_number(_row_count_name).group_by(group_key).agg(pl.col(_row_count_name)).to_dict()
+    )
 
-    res = dict(zip(groups[group_key].to_list(), groups["__row_count__"].to_list()))
+    res = dict(zip(groups[group_key], groups[_row_count_name]))
     return res
```
jrycw, line 266: @machow Do you think this approach is better since it only requires one `to_dict()` call instead of two `to_list()` calls?

great_tables/_tbl_data.py
```diff
@@ -295,8 +298,7 @@ def _(
     elif callable(expr):
         # TODO: currently, we call on each string, but we could be calling on
         # pd.DataFrame.columns instead (which would let us use pandas .str methods)
-        col_pos = {k: ii for ii, k in enumerate(list(data.columns))}
-        return [(col, col_pos[col]) for col in data.columns if expr(col)]
+        return [(col, i) for i, col in enumerate(data.columns) if expr(col)]
```
jrycw, line 301: @machow It seems we can directly utilize the index from `enumerate` without constructing another `dict` for lookup.

jrycw:
@machow Thanks for your detailed explanation. I think that's the reason the code looks so familiar to me. It seems that I've made a similar error a few months ago. I've left two comments that might be beneficial for this code in the PR. Would you mind taking a look?

PR #289, opened 2024-04-14
Refactor `letters` and `Letters` functions in `_helpers`

jrycw:
I recommend utilizing the built-in `string` module to define the `letters` and `Letters` functions.


machow (approved):
LGTM! Thanks for submitting, this was really helpful to see.

Issue #222, opened 2024-02-28
ISO-formatted date strings don't work with `.fmt_datetime()`


rich-iannone:
Datetime strings work well within the `.fmt_datetime()` method, so long as they are ISO-8601 formatted. However, date strings in the same format do not work (though they ought to).

Here's an MRE that shows the error/stacktrace:

```python
from great_tables import GT, exibble

exibble_mini = exibble[["date", "time"]]

(
    GT(exibble_mini)
    .fmt_datetime(
        columns="date",
        date_style="month_day_year",
        time_style="h_m_s_p"
    )
)
```

```console
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File [~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:344](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:344), in BaseFormatter.__call__(self, obj)
    342     method = get_real_method(obj, self.print_method)
    343     if method is not None:
--> 344         return method()
    345     return None
    346 else:

File [~/py_projects/great-tables/great_tables/gt.py:194](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:194), in GT._repr_html_(self)
    193 def _repr_html_(self):
--> 194     return self.render(context="html")

File [~/py_projects/great-tables/great_tables/gt.py:293](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:293), in GT.render(self, context)
    292 def render(self, context: str) -> str:
--> 293     html_table = self._build_data(context=context)._render_as_html()
    294     return html_table

File [~/py_projects/great-tables/great_tables/gt.py:272](https://file+.vscode-resource.vscode-cdn.net/Users/riannone/py_projects/great-tables/~/py_projects/great-tables/great_tables/gt.py:272), in GT._build_data(self, context)
    269 def _build_data(self, context: str) -> Self:
    270     # Build the body of the table by generating a dictionary
    271     # of lists with cells initially set to nan values
--> 272     built = self._render_formats(context)
    273     # built._body = _migrate_unformatted_to_output(body)
    274 
...
[...]

PR #290, opened 2024-04-14
Refactor `fmt_time`, `fmt_date` and `fmt_datetime`

jrycw:
Related to #222 , I'm thinking we could possibly delegate the validation of lifts to the built-in `datetime.datetime.fromisoformat` function.

Fixes: #299


machow:
@rich-iannone can you verify that `datetime.fromisoformat()` handles the cases you were thinking of? Can you be sure to enumerate in a comment the cases you try (for posterity)?

rich-iannone:
I mostly wanted to make sure that `datetime.fromisoformat()` handled most of the common cases and I found it's quite capable. The following input variations results in correct (the same) output:

```{python}
import datetime
from great_tables import vals

iso_str = [
    "2020-05-20",
    "2020-05-20 22:30",
    "2020-05-20T22:30",
    "2020-05-20 22:30:05",
    "2020-05-20 22:30:05.824" # milliseconds *must* have 3 decimal places
    ]

fmtd_dates = []

for val in iso_str:
    fmtd_dates.append((vals.fmt_date(datetime.datetime.fromisoformat(val), date_style="wd_m_day_year"))[0])

fmtd_dates
```

```
['Wed, May 20, 2020',
 'Wed, May 20, 2020',
 'Wed, May 20, 2020',
 'Wed, May 20, 2020',
 'Wed, May 20, 2020']
```

So I think the code change is great here. Thanks, @jrycw !

jrycw:
> I mostly wanted to make sure that `datetime.fromisoformat()` handled most of the common cases and I found it's quite capable. The following input variations results in correct (the same) output:
> 
> ```
> import datetime
> from great_tables import vals
> 
> iso_str = [
>     "2020-05-20",
>     "2020-05-20 22:30",
>     "2020-05-20T22:30",
>     "2020-05-20 22:30:05",
>     "2020-05-20 22:30:05.824" # milliseconds *must* have 3 decimal places
>     ]
> 
> fmtd_dates = []
> 
> for val in iso_str:
>     fmtd_dates.append((vals.fmt_date(datetime.datetime.fromisoformat(val), date_style="wd_m_day_year"))[0])
> 
> fmtd_dates
> ```
> 
> ```
> ['Wed, May 20, 2020',
>  'Wed, May 20, 2020',
>  'Wed, May 20, 2020',
>  'Wed, May 20, 2020',
>  'Wed, May 20, 2020']
> ```
> 
> So I think the code change is great here. Thanks, @jrycw !

Please correct me if I'm mistaken, but it seems you're testing `fmt_date` instead of `fmt_datetime`. Since `vals.fmt_datetime` isn't available, it needs to be tested from `GT`.

rich-iannone:
@jrycw you are right. I mistakenly thought that `.fmt_date()` used the same internal function. I checked out this branch and tried out some variations. I found that this could be further improved if there is a single validation and transformation from `datetime.fromisoformat(x)` (we could drop the use of `_validate_iso_datetime_str()`).

With that change, the following code works:

```
import datetime
from great_tables import GT
import polars as pl

df = pl.DataFrame({"x": [
    "2020-05-20",
    "2020-05-20 22:30",
    "2020-05-20T22:30",
    "2020-05-20 22:30:05",
    "2020-05-20T22:30:05.232" # milliseconds *must* have 3 decimal places
    ]})

GT(df).fmt_datetime(columns="x", date_style="wd_m_day_year", time_style="h_m_p")
```

It produces this:
fmt_datetime_refactor_tbl

This seems like a good improvement because the first and last cases failed before, but they provide formatted output now (and they work as can be expected).

To do this, I renamed `_iso_to_datetime()` to `_iso_str_to_datetime()` with this changed it to:

```python
def _iso_str_to_datetime(x: str) -> datetime:
    """
    Converts a string in ISO format to a datetime object.

    Args:
        x (str): The string to be converted.

    Returns:
        datetime: The converted datetime object.
    """
    return datetime.fromisoformat(x)
```

The code in `fmt_datetime()` changes to:

```python

        # If `x` is a string, assume it is an ISO datetime string and convert it to a datetime object
        if isinstance(x, str):

            # Convert the ISO datetime string to a datetime object
            x = _iso_str_to_datetime(x)
```

@jrycw @machow does this change make sense? I could push the changes to the branch (or you could modify). Also, some new test cases are needed here.

jrycw:
LGTM! If @machow is also on board with this, could @rich-iannone push your changes to the branch? I'll then proceed to add the relevant tests based on your demonstration.

machow:
This sounds good to me!

jrycw:
It turns out we could make similar modifications to `fmt_date`. Here's a summary of the changes:
* Removed `_validate_iso_date_str`.
* Added `_iso_str_to_date`.
* Retained `_iso_to_date` (since `_generate_data_vals` relies on this function).
* Added `test_fmt_date`.

I believe the modifications are correct, but please feel free to correct any mistakes I may have made.

jrycw:
I attempted to remove `_iso_to_date` and replace it with `_iso_str_to_date` in `_generate_data_vals`, and it seems to have worked.

rich-iannone (approved):
LGTM!

PR #294, opened 2024-04-17
Update `_data_color`

jrycw:
Hello team, I've attempted to simplify the logic in this PR for `_data_color`. I've updated the code according to my coding preferences, but please feel free to reject or modify the code based on your judgement.

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-from itertools import chain, repeat
 from typing import (
     TYPE_CHECKING,
     Union,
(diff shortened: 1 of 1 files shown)


great_tables/_data_color/base.py
```diff
@@ -563,17 +558,8 @@ def _expand_short_hex(hex_color: str) -> str:
...
-
-    # Get the second character of the hex color
-    second_char = hex_color[1]
-
-    # Get the third character of the hex color
-    third_char = hex_color[2]
-
     # Return the expanded 6-digit hexadecimal color value
-    expanded = "#" + first_char + first_char + second_char + second_char + third_char + third_char
+    expanded = "".join(chain("#", *zip(*repeat(hex_color, 2))))
```
machow, line 562: What about something like this?:

```pyhton
"#" + "".join([x*2 for x in hex_color])
```

It took me a bit of thinking to think through how zip and repeat combined (since I don't use repeat too often 😅 )

jrycw: No worries. The updated code came to mind while refactoring. I believe your solution works and is arguably more readable. Personally, I might write `"#" + "".join(x * 2 for x in hex_color)`, but feel free to update this PR based on your judgment.

PR #295, opened 2024-04-17
Update `_utils_nanoplots.py`

jrycw:
This PR primarily addresses the following three issues:

1. Offloading some of the statistical calculations to `numpy`.
2. Converting `_flatten_list` into a function that can recursively flatten the list. This feature may be unwanted, so feel free to reject this change.
3. Making `_gt_first` and `_gt_last` more general so they can accept any iterable as the parameter.

### DIFF
--- a/great_tables/_utils_nanoplots.py
+++ b/great_tables/_utils_nanoplots.py
@@ -72,8 +72,6 @@ def _check_any_na_in_list(x: List[Union[int, float]]) -> None:
     if _any_na_in_list(x):
         raise ValueError("The list of values cannot contain missing values.")
 
-    return None
-
 
 # Remove missing values from a list of values
 def _remove_na_from_list(x: List[Union[int, float]]) -> List[Union[int, float]]:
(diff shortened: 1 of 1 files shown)


great_tables/_utils_nanoplots.py
```diff
@@ -349,13 +343,9 @@ def _flatten_list(x: List[Any]) -> Union[List[float], List[int], List[Union[int,
...
 
     flat_list = []
-
-    # Iterate through the outer list
     for element in x:
-        if type(element) is list:
-            # If the element is of type list, iterate through the sublist
-            for item in element:
-                flat_list.append(item)
+        if isinstance(element, list):
```
machow, line 347: Thanks for this cleanup--I would be on board merging this fix

machow (changes requested):
Thanks for submitting this PR! I think we're likely going to try and remove numpy in the near future, so aren't looking to add it to functions like `_gt_min`. I flagged one block of code that would be nice to merge, though!

If you're up for it, we'd gladly accept a separate PR that removes numpy, but I realize it's a fairly gnarly task. No worries if it's outside of what you're looking to do. I really appreciate all this time you've put into cleaning things up 😅.

machow:
I've added an issue describing removing numpy, in case you want to tackle it. But no worries if not!

https://github.com/posit-dev/great-tables/issues/296

jrycw:
> I've added an issue describing removing numpy, in case you want to tackle it. But no worries if not!
> 
> #296

May I inquire about the possibility of handling or representing something like `nan` in the future without using `numpy`?

jrycw:
> Thanks for submitting this PR! I think we're likely going to try and remove numpy in the near future, so aren't looking to add it to functions like `_gt_min`. I flagged one block of code that would be nice to merge, though!
> 
> If you're up for it, we'd gladly accept a separate PR that removes numpy, but I realize it's a fairly gnarly task. No worries if it's outside of what you're looking to do. I really appreciate all this time you've put into cleaning things up 😅.

Removing `numpy` as a dependency is not a trivial task, and it may be beyond the scope of this PR. I've modified the PR to retain the block you marked. Additionally, there's a `return None` statement in `_check_any_na_in_list`, which I believe should be removed. As always, feel free to make further updates to the PR based on the team's judgment.

great_tables/_utils_nanoplots.py
jrycw: By the way, there are still some places in the codebase where direct type comparisons like `if type(x) == list` are used. Do you think it would be better if we try to replace all of them with `if isinstance(x, list)`?

machow:
> May I inquire about the possibility of handling or representing something like nan in the future without using numpy?

Yeah, we can use a combo of `None` for nan and `math.inf` for `np.inf` (similar to how polars converts types in `to_list()`)

PR #298, opened 2024-04-17
Update `x_locales.csv`

jrycw:
I'd like to submit a PR to update the information of `zh-Hant` in `x_locales.csv` based on my understanding.


machow:
cc @rich-iannone where did this locale data come from? Do you mind reviewing the change?

rich-iannone:
All the underlying locale data with associated parameter values is compiled from https://github.com/unicode-org/cldr-json. I maintain a package that tries to keep in step with their updates. I'll definitely look into this!

PR #300, opened 2024-04-17
ci: test on a windows runner

machow:
This PR addresses potential issues when using `GT.save()` on a windows machine. It removes an unnecessary tmp dir being used when initiating NamedTemporaryFile (which could cause bizarreness), and adds a windows runner to our CI.

Edit: let's push fine-tuning to https://github.com/posit-dev/great-tables/pull/276, which uploads saved table images as artifacts in our CI.


machow:
Okay, the remaining test failures are all related to paths in our fmt_image snapshot tests!

PR #310, opened 2024-04-20
Proposal for improving `GTData._replace` initialization

jrycw:
In our codebase, `GTData._replace` is widely used. I'm suggesting two potential improvements for the initialization of `GTData._replace`:

* Utilizing `set.difference` operations to construct the `missing set`.
* Deferring the `copy.copy(self)` step after the `missing` check.

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -46,14 +46,12 @@ class GTData:
     _has_built: bool = False
 
     def _replace(self, **kwargs: Any) -> Self:
-        new_obj = copy.copy(self)
-
-        missing = {k for k in kwargs if k not in new_obj.__dict__}
+        missing = set(kwargs).difference(vars(self))
         if missing:
             raise ValueError(f"Replacements not in data: {missing}")
 
+        new_obj = copy.copy(self)
         new_obj.__dict__.update(kwargs)
-
         return new_obj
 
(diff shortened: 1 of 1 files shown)


machow:
Thanks for looking into this---since GTData is a dataclass, we should be able to use the `from dataclass import fields` function to get field names. WDYT of that approach? IMO the biggest pain point of _replace() currently is that it doesn't have type hints, so the static checker doesn't warn when you use a kwarg that would error.

I think the `_replace()` method is mostly duplicating the dataclass `replace()` method. IIRC, it exists, because using `replace()` on the subclass `GT.replace()` doesn't quite work (due to GT using a totally different `__init__()` signature I think?).

It's sort of a balancing act between GTData (which we want for more limited scope in most our functions), and GT (where we want to have all the functions on it as methods).

jrycw:
I tend to be more of an old-school Python developer and haven't used `dataclass` extensively in my code (although I know I should). However, after reviewing the codebase for the past few days, it appears that we've already utilized `fields` extensively. Based on my limited knowledge, it seems to be functioning well. For type hints, it appears that an adapter data structure like `Enum` or `dataclass` would provide a smoother coding experience.

machow:
I'm going to close for now, but would love to revisit this in the future (and happy to reopen if useful:).

PR #317, opened 2024-04-25
fix: missing values in `fmt_number`

### LINKED ISSUE #314: The `GT.fmt_number()` method fails when formatting includes missing values in Polars DataFrames
Formatting with `fmt_number()` fails when the DF is a Polars DF and when the formatter hits a missing value. Here is a failing case

```python
import polars as pl

exibble_pl = pl.from_pandas(exibble)

GT(exibble_pl).fmt_number(columns="num")
```

Error is:

```console
TypeError                                 Traceback (most recent call last)
File .../great-tables/env/lib/python3.9/site-packages/IPython/core/formatters.py:344, in BaseFormatter.__call__(self, obj)
    342     method = get_real_method(obj, self.print_method)
    343     if method is not None:
--> 344         return method()
    345     return None
    346 else:

File .../great-tables/great_tables/gt.py:270, in GT._repr_html_(self)
    267 make_page = defaults["make_page"]
    268 all_important = defaults["all_important"]
--> 270 rendered = self.as_raw_html(
    271     make_page=make_page,
    272     all_important=all_important,
    273 )
    275 return rendered

File .../great-tables/great_tables/_export.py:33, in as_raw_html(self, make_page, all_important)
     12 def as_raw_html(
     13     self: GT,
     14     make_page: bool = False,
     15     all_important: bool = False,
     16 ) -> str:
     17     """
...
--> 297     x = x * scale_by
    299     # Determine whether the value is positive
    300     is_negative = _has_negative_value(value=x)

TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

Error has to do with indiscriminate scaling involving a multiplication (default is `1`) which assumes every value (NA or not) should undergo numeric scaling. Avoiding the missing value results in a successful table render with no error (e.g., try using `GT(exibble_pl.head(4)).fmt_number(columns="num")`).

I've tested variations of a Polars DF column with missing values with every other formatting method (e.g., `fmt_integer()`, `fmt_bytes()`, etc.) and it appears that this issue is isolated to `fmt_number()`.

jrycw:
Fix #314.

At first, I attempted to address this issue by utilizing the check from `fmt_scientific` to verify `x`:

```python
if is_na(self._tbl_data, x):
    return x
```

However, this caused a failure in `tests/test_export.py::test_html_string_generated`. Consequently, I opted to directly check whether `x` is `None` or not.

```python
import polars as pl

exibble_pl = pl.from_pandas(exibble)

GT(exibble_pl).fmt_number(columns="num")
```
image

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -293,7 +293,7 @@ def fmt_number(
         dec_mark: str = dec_mark,
         force_sign: bool = force_sign,
     ):
-        if x is None:
+        if is_na(self._tbl_data, x):
             return x
 
         # Scale `x` value by a defined `scale_by` value
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -281,7 +281,7 @@ def fmt_number(
 
     # Generate a function that will operate on single `x` values in the table body
     def fmt_number_fn(
-        x: float,
+        x: float | None,
         decimals: int = decimals,
         n_sigfig: Optional[int] = n_sigfig,
         drop_trailing_zeros: bool = drop_trailing_zeros,
@@ -293,6 +293,9 @@ def fmt_number(
         dec_mark: str = dec_mark,
         force_sign: bool = force_sign,
     ):
+        if x is None:
+            return x
+
         # Scale `x` value by a defined `scale_by` value
         x = x * scale_by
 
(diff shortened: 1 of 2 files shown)


machow:
shoot... thanks for looking into this. I looked at a diff of the snapshot, and it seems okay to use the same `is_na()` guard is this function, since it's behavior will match the others. Hope it's okay, I just added to this PR!

rich-iannone (approved):
LGTM!

rich-iannone:
Thank you @jrycw !

PR #319, opened 2024-04-26
fix: the display issue of integerlike values in nanoplots

### LINKED ISSUE #279: Integer or integerlike values should be displayed as integers in nanoplots
If all values for a single nanoplot are integerlike, the displayed values should not contain decimal values filled with zeroes. The problem can be seen in the example page at https://posit-dev.github.io/great-tables/get-started/nanoplots.html. Here's a screenshot showing the issue:

nanoplot_displayed_number

All values in the bottom plot don't have non-zero decimal parts (they are parsed from the string: `"2 0 15 7 8 10 1 24 17 13 6"`) so the third value should display as `15` and not `15.0`.

When numbers are autoformatted for nanoplots, the aim is to have the most compact and consistent display (and the user could provide a function for formatting as a customization). So, removing unnecessary digits makes sense here.

jrycw:
Fix #279 . This PR aims to improve the display of integer or integerlike values in nanoplots. During the implementation, I came across two interesting findings in the code:
* Some locales may use the minus symbol `−` instead of `-`. Therefore, a conversion like `str(n).replace("−", "-")` might be necessary.
* The negative values appear to be handled appropriately with this approach.

## demo
```python
from great_tables import GT
import polars as pl

random_numbers_df = pl.DataFrame(
    {
        "i": range(1, 5),
        "lines": [
            "20 23 6 7 37 23 21 4 7 16",
            "2.3 6.8 9.2 2.42 3.5 12.1 5.3 3.6 7.2 3.74",
            "-12 -5 6 3.7 0 8 -7.4",
            "-12 0 15 7 8 10 1 24 17 13 6", # negative values should be ok
        ],
    }
).with_columns(bars=pl.col("lines"))

(
    GT(random_numbers_df, rowname_col="i")
    .fmt_nanoplot(columns="lines")
    .fmt_nanoplot(columns="bars", plot_type="bar")
)
```
image

image

### DIFF
--- a/great_tables/_utils_nanoplots.py
+++ b/great_tables/_utils_nanoplots.py
@@ -1580,6 +1580,7 @@ def _is_intlike(n: Any, scaled_by: float = 1e17) -> bool:
 
     if isinstance(n, str):
         try:
+            # Replacement of minus sign (U+2212) with hyphen (necessary in some locales)
             n = float(n.replace("−", "-"))
         except ValueError:
             return False
@@ -1599,6 +1600,7 @@ def _remove_exponent(n: "str | int | float") -> int:
     from decimal import Decimal
 
     if isinstance(n, str):
+        # Replacement of minus sign (U+2212) with hyphen (necessary in some locales)
         n = str(n).replace("−", "-")
     d = Decimal(n)
     if d == d.to_integral():
--- a/great_tables/_utils_nanoplots.py
+++ b/great_tables/_utils_nanoplots.py
@@ -1592,7 +1592,7 @@ def _get_n_intlike(nums: list[Any]) -> int:
     return len([n for n in nums if _is_intlike(n)])
 
 
-def _remove_exponent(n: str | int | float) -> int:
+def _remove_exponent(n: "str | int | float") -> int:
     """
     https://docs.python.org/3/library/decimal.html#decimal-faq
     """
--- a/great_tables/_utils_nanoplots.py
+++ b/great_tables/_utils_nanoplots.py
@@ -1458,6 +1458,7 @@ def _generate_nanoplot(
     #
 
     if show_vertical_guides:
+        is_all_intify = len(y_vals) == _get_n_intlike(y_vals)
 
         g_guide_strings = []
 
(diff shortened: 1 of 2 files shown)


machow:
@rich-iannone do you mind taking a look? The test cases look good, but I'm a bit less familiar with this piece!

rich-iannone:
This looks great @jrycw ! I also took the opportunity to make the y-axis guides render in the same fashion (stripping decimal part of low and high value if *all* values of *y* are integerlike).

Here's what it looks like with my last change applied:

y-axis-as-integer-vals

I used the following table code in my tests to examine the rendered values:

```py
from great_tables import GT
import polars as pl

random_numbers_df = pl.DataFrame(
    {
        "i": range(1, 6),
        "lines": [
            "20 23 6 7 37 23 21 4 7 16",  # integer like
            "2.3 6.8 9.2 2.42 3.5 12.1 5.3 3.6 7.2 3.74",  # two decimal places
            "-12 -5 6 3.7 0 8 -7.4",  # not integerlike, one decimal place
            "-12 0 15 7 8 10 1 24 17 13 6",  # negative values should be ok
            "5.23e-11 15E-12 7E-12 8e-12 10e-12 1e-12 24e-12 17e-12 13e-12 6e-12",  # small values
        ],
    }
).with_columns(bars=pl.col("lines"))

(
    GT(random_numbers_df, rowname_col="i")
    .fmt_nanoplot(columns="lines")
    .fmt_nanoplot(columns="bars", plot_type="bar")
)
```

Let me know if the extra assignments look okay in terms of good coding style.

jrycw:
@rich-iannone, that looks fantastic. Honestly, I had wanted to make similar modifications initially. However, I got a bit lost in the `_generate_nanoplot` function. `Nanoplots` is definitely a standout feature of `Great Tables`, and I truly appreciate it. However, as more features are added, the codebase grows quickly, and it took me a while to figure out my place within each `if` statement. I feel that we might need to refactor this function into smaller pieces for better maintenance."

rich-iannone:
Thanks again for (so quickly!) looking at my final changes. Agreed that the nanoplots code could use a refactor and some abstractions, and we'll get there!

Will do a final review now :)

PR #320, opened 2024-04-27
fix: add tests for `cols_align` and extend support for `Polars` expressions

### LINKED ISSUE #51: Add test of column alignment
It looks like from [this failed docs build](https://github.com/posit-dev/great-tables/actions/runs/7093672362/job/19307526874?pr=50) that the only call to `GT.cols_align()` might be in a docstring. Open to track us adding tests of `.cols_align()`.

jrycw:
Fix #51. This PR focuses on adding tests for `cols_align`. 

During the process, I discovered that it should be feasible for `cols_align` to support `Polars` expressions. This enhancement is achieved by adding `sel_cols = resolve_cols_c(data=self, expr=columns)`.

### DIFF
--- a/great_tables/_boxhead.py
+++ b/great_tables/_boxhead.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Optional, Any
 
+from ._locations import resolve_cols_c
 from ._utils import _assert_list_is_subset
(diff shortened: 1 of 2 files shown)


machow (approved):
This looks great--thanks so much for this! @rich-iannone do you mind merging?

Issue #35, opened 2023-11-28
Test use of polars selectors throughout library


machow:
Related to https://github.com/posit-dev/great-tables/pull/34, I've implemented the path for using polars selectors, but need to add more tests of its usage. It seems like aspects of selection can be consolidated in places (e.g. the handling of `columns` in the `cols_*` functions).

Most functions that take a selection argument do a bit of custom checking, and these checks seem largely the same across functions..!

machow:
To check:

* select via scalar (e.g. "some_col")
* select via list
* select via polars
* select via lambda
* (what should happen on a list of polars selectors?)

PR #325, opened 2024-05-02
Enhance test coverage for column selection

jrycw:
Related to #35.

This PR aims to enhance test coverage for `cols_move`, `cols_move_to_start`, `cols_move_to_end`, and `cols_hide` functions in both `Pandas` and `Polars`. 

Meanwhile, given the substantial reliance on column selection within the `fmt_*` functions, should we conduct thorough tests for them using `Polars` selectors, or are we satisfied with the current state?

### DIFF
--- a/tests/test_spanners.py
+++ b/tests/test_spanners.py
@@ -1,10 +1,14 @@
 import pandas as pd
+import polars as pl
+import polars.selectors as cs
 import pytest
 from great_tables import GT
 from great_tables._gt_data import Boxhead, ColInfo, ColInfoTypeEnum, SpannerInfo, Spanners
 from great_tables._spanners import (
     cols_hide,
     cols_move,
+    cols_move_to_end,
+    cols_move_to_start,
     empty_spanner_matrix,
     spanners_print_matrix,
     tab_spanner,
@@ -12,7 +16,7 @@ from great_tables._spanners import (
 
 
 @pytest.fixture
-def spanners():
+def spanners() -> Spanners:
     return Spanners(
         [
             SpannerInfo(spanner_id="a", spanner_level=0, vars=["col1"], built="A"),
@@ -22,7 +26,7 @@ def spanners():
 
 
 @pytest.fixture
-def boxhead():
+def boxhead() -> Boxhead:
     return Boxhead(
         [
             ColInfo(var="col1"),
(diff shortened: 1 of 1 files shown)


tests/test_spanners.py
```diff
@@ -231,3 +197,75 @@ def test_cols_width_fully_set_pct_2():
     assert gt_tbl._boxhead[0].column_width == "10%"
     assert gt_tbl._boxhead[1].column_width == "10%"
     assert gt_tbl._boxhead[2].column_width == "40%"
+
+
+@pytest.mark.parametrize("df_lib, columns", [(pd, "a"), (pl, cs.starts_with("a"))])
```
machow, line 202: WDYT of making `df_lib` a fixture, that passes in the dataframe directly? Similar to this fixture in tbl_data tests

https://github.com/posit-dev/great-tables/blob/main/tests/test_tbl_data.py#L19-L25

machow (changes requested):
Thanks for this. These tests are so helpful! I left a question about potentially changing `df_lib` to be a fixture. WDYT?

machow:
> Meanwhile, given the substantial reliance on column selection within the fmt_* functions, should we conduct thorough tests for them using Polars selectors, or are we satisfied with the current state?

It seems okay to lean on the tests of `fmt()`, which the `fmt_*()` functions should end up calling (and handles the column selection).

https://github.com/posit-dev/great-tables/blob/main/tests/test_formats.py#L74

tests/test_spanners.py
jrycw: Please correct me if I'm mistaken, but it appears that the current test case will generate two small test cases for each test function. However, if we do:

```python
params_frames = [pytest.param(pd.DataFrame, id="pandas"), pytest.param(pl.DataFrame, id="polars")]


@pytest.fixture(params=params_frames, scope="function")
def df(request):
    return request.param({"a": [1, 2], "b": [3, 4], "c": [5, 6], "d": [7, 8]})


@pytest.mark.parametrize("columns", ["a", cs.starts_with("a")])
def test_cols_move_single_col(df, columns):
    src_gt = GT(df)
    new_gt = cols_move(src_gt, columns=columns, after="b")
    assert [col.var for col in new_gt._boxhead] == ["b", "a", "c", "d"]
```

It seems that we'll have four small test cases for each function, resulting in two types of dataframes with two types of columns. It's evident that when `Pandas` interacts with `Polars` columns, the test will fail.

```python
tests/test_spanners.py::test_cols_move_single_col[pandas-a] PASSED                                                                       [ 25%]
tests/test_spanners.py::test_cols_move_single_col[pandas-columns1] FAILED                                                                [ 50%]
tests/test_spanners.py::test_cols_move_single_col[polars-a] PASSED                                                                       [ 75%]
tests/test_spanners.py::test_cols_move_single_col[polars-columns1] PASSED                                                                [100%]

=========================================================== short test summary info ============================================================
FAILED tests/test_spanners.py::test_cols_move_single_col[pandas-columns1] - NotImplementedError: Unsupported selection expr: col("^(a).*$")
1 failed, 3 passed, 1 warning in 0.96s
```

machow: Ah, I didn't realize this is testing specific combinations, where the suggestion just crosses everything. I seems reasonable to keep it as is!

machow (approved):

machow:
Thanks so much for making sure things are covered! @rich-iannone do you want to merge?

PR #326, opened 2024-05-04
fix: `CellStyleBorders` is not properly constructed when `sides` is set to "all"

jrycw:
It seems that the following code does not render as expected when `sides` is set to "all" for `style.borders`. The main reason might be the omission of other values while constructing a new `CellStyleBorders`.

```python
import polars as pl
from great_tables import GT, md, html, style, loc
from great_tables.data import islands

islands_mini = (
    pl.from_pandas(islands).sort("size", descending=True)
    .head(10)
)

(
    GT(islands_mini)
    .tab_style(
      style=style.borders(sides=["all"], style="solid", weight="4px", color="red"),
      locations=loc.body(rows=[-1])
    )
)
```

The table comparison before and after fixing can be found as below:

comparison

### DIFF
--- a/great_tables/_styles.py
+++ b/great_tables/_styles.py
@@ -322,7 +322,12 @@ class CellStyleBorders(CellStyle):
 
         # If 'all' is provided then call the function recursively with all sides
         if "all" in self.sides:
-            return CellStyleBorders(sides=["top", "bottom", "left", "right"])._to_html_style()
+            return CellStyleBorders(
+                sides=["top", "bottom", "left", "right"],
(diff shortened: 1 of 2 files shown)


machow:
Ah, nice catch, thanks! I added a small lil test, and everything looks good!

machow (approved):

machow:
@rich-iannone do you mind merging? :)

jrycw:
> Ah, nice catch, thanks! I added a small lil test, and everything looks good!

@machow, thanks for adding a test.

PR #327, opened 2024-05-05
Fix issue where `fmt` cannot accept multiple `fns`

jrycw:
Currently , the following code will raise `AttributeError: 'list' object has no attribute 'default'`.
```python

(
    GT(exibble[["fctr"]])
    .fmt([lambda x: x], columns="fctr")
)
```
The main issue appears to be that `fmt` did not handle `fns` properly when it is a `list`. As a result, the callables are not wrapped in `FormatFns`. Additionally, `Body.render_formats` needs to be refactored to accommodate this change by retrieving the result of `_set_cell` for each call. Given the critical role of `fmt` within this project, while the proposed PR may not be flawless, it serves as a reference for our refactoring efforts. With some snapshot tests failing, I would greatly appreciate the team's assistance in reviewing the PR to ensure the failed tests are acceptable.

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -99,23 +99,25 @@ def fmt(
         The GT object is returned. This is the same object that the method is called on so that we
         can facilitate method chaining.
     """
+    row_res = resolve_rows_i(self, rows)
+    row_pos = [name_pos[1] for name_pos in row_res]
+    col_res = resolve_cols_c(self, columns)
 
     # If a single function is supplied to `fns` then
     # repackage that into a list as the `default` function
+    formatter: list[FormatInfo] = []
     if isinstance(fns, Callable):
-        fns = FormatFns(default=fns)
-
-    row_res = resolve_rows_i(self, rows)
-    row_pos = [name_pos[1] for name_pos in row_res]
-
-    col_res = resolve_cols_c(self, columns)
-
-    formatter = FormatInfo(fns, col_res, row_pos)
+        _fmtfs = FormatFns(default=fns)
+        formatter.append(FormatInfo(_fmtfs, col_res, row_pos))
+    elif isinstance(fns, list):
+        for _fns in fns:
+            _fmtfs = FormatFns(default=_fns)
+            formatter.append(FormatInfo(_fmtfs, col_res, row_pos))
 
     if is_substitution:
-        return self._replace(_substitutions=[*self._substitutions, formatter])
+        return self._replace(_substitutions=[*self._substitutions, *formatter])
 
-    return self._replace(_formats=[*self._formats, formatter])
+    return self._replace(_formats=[*self._formats, *formatter])
 
 
 def fmt_number(
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -197,11 +197,13 @@ def _(data, row: int, column: str, value: Any) -> None:
     # if this is violated, get_loc will return a mask
     col_indx = data.columns.get_loc(column)
     data.iloc[row, col_indx] = value
+    return data
 
 
 @_set_cell.register(PlDataFrame)
 def _(data, row: int, column: str, value: Any) -> None:
     data[row, column] = value
+    return data
 
 
 # _get_column_dtype ----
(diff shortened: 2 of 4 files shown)


great_tables/_tbl_data.py
```diff
@@ -197,11 +197,13 @@ def _(data, row: int, column: str, value: Any) -> None:
     # if this is violated, get_loc will return a mask
     col_indx = data.columns.get_loc(column)
     data.iloc[row, col_indx] = value
+    return data
```
machow, line 200: Since `_set_cell()` mutates data, it seems useful to continue not returning `data` (sometimes called [Command query separation](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation)).

great_tables/_formats.py
```diff
@@ -99,23 +99,25 @@ def fmt(
...
-
     row_res = resolve_rows_i(self, rows)
     row_pos = [name_pos[1] for name_pos in row_res]
-
     col_res = resolve_cols_c(self, columns)
 
-    formatter = FormatInfo(fns, col_res, row_pos)
+    # If a single function is supplied to `fns` then
+    # repackage that into a list as the `default` function
+    formatter: list[FormatInfo] = []
```
machow, line 108: I think unfortunately GT.fmt never expects a list of functions as inputs, but uses the maybe trickily named FormatFns to mean, a dataclass that holds functions for different renderers (e.g. html, in the future latex, etc..).

It's a good point though that `GT.fmt` is an important entry point, so it's worth thinking carefully about its behavior. I think if it comes down to it, people could always pass `lambda x: formatter2(formatter1(x))`? Or use some implementation of a `compose()` function?

These docs:

```
    Parameters
    ----------
    fns
        Either a single formatting function or a named list of functions.
```

Unfortunately are copied over from R, and so named list is referring to something more like a dictionary (which we're currently using the FormatFns dataclass for)

machow (changes requested):
Hey @jrycw, thanks for taking a look at this aspect of formatting. I left a comment explaining a bit more, but I think a list of formatting functions isn't expected (but an unfortunate bit of documentation copied from R made it sound like the case ☹️ )

It seems like some of the failing snapshots have changed because their formatting isn't running anymore, but I'm not sure why (based on the new code). For example,

```
def test_html_string_generated(gt_tbl: GT, snapshot: str):
>       assert snapshot == gt_tbl.as_raw_html()
E       assert [- snapshot] == [+ received]
E           '''
E             ...
E               <td class="gt_row gt_left">apricot</td>
E         -     <td class="gt_row gt_right">$49.95</td>
E         +     <td class="gt_row gt_right">49.95</td>
E             </tr>
E              ...
E               <td class="gt_row gt_left">banana</td>
E         -     <td class="gt_row gt_right">$17.95</td>
E         +     <td class="gt_row gt_right">17.95</td>
E             </tr>
```

In this output, it seems like the dollar signs are expected, since it runs the `fmt_currency()` method.

jrycw:
@machow , thank you so much for explaining the details to me. I believe making a small modification to the docs will make it easier for people who haven't used `{gt}` before (like me). Since I'll be out of town for a few days, would you mind taking over the PR to either close it or modify the docs?

great_tables/_formats.py
jrycw: I'm curious about the possibility of implementing a `compose()` function. In my imagination, the code might look something like this:
```python
from functools import partial
from typing import Any, Callable


def compose(funcs: list[Callable[Any, Any]]):
    def _compose(x: Any, funcs: list[Callable[Any, Any]]):
        for f in funcs:
            x = f(x)
        return x
    return partial(_compose, funcs=funcs)


(
    GT(exibble[["fctr"]]).fmt(
        compose([lambda x: f"_{x}_", lambda x: f"*{x}*", lambda x: f"<{x}>"]),
        columns="fctr",
    )
)
```
image

machow:
>  I believe making a small modification to the docs will make it easier for people who haven't used {gt} before (like me)

That makes a ton of sense --- I can make sure to change the docstring Monday. I'll look back through to see if there are pieces to keep from the PR (since it seemed like there were some other improvements in here :)

jrycw:
@machow, thanks a lot.

jrycw:
@machow, this PR has been open for a while. Could you please help edit the docstring, and then we might be able to close it?
```
fns : FormatFn | FormatFns
    Either a single formatting function or a named list of functions.
```

machow:
@rich-iannone let's keep fmt()'s current behavior. Do you mind checking to make sure any of the docstring corrections are applied (can close this and open a new PR is useful).

rich-iannone:
I'm closing this in favor of fixing the misleading documentation in the new PR https://github.com/posit-dev/great-tables/pull/818. Thanks!

PR #329, opened 2024-05-07
Fix fmt_image erroring on missing values

machow:
This PR fixes a bug, where fmt_image errored on missing values.


rich-iannone (approved):
LGTM!

PR #330, opened 2024-05-07
Fix nanoplots listcols

machow:
This PR ensures Great Tables can render a polars DataFrame that has a pl.List column. Previously, it would error, and the error was early enough that formatters couldn't work with this data.

(Note that most formatters are not designed to work with list columns. We may need to support this down the road, or add helpful messaging when erroring in these cases).

(Note also that this might fail for complex list schemas, eg. nested lists like `pl.List(pl.List(int))` rather than `pl.List(int)`. But this case also fails in the current code).

Here's a hokey example displaying a table with list columns:

```python
from great_tables import GT
import polars as pl

GT(pl.DataFrame({"x": [[1, 2], [3]], "y": [[.9], [1.1]]}))
```

image

And with a tasty nanoplot

```python
GT(pl.DataFrame({"x": [[1, 2], [3]], "y": [[.9], [1.1]]})).fmt_nanoplot("x")
```

image


tests/__snapshots__/test_formats.ambr
```diff
@@ -1,4 +1,16 @@
...
...M 50.0,115.0 150.0,15.0 150.0,125 50.0,125 Z" stroke="transparent" stroke-width="2" fill="url(#area_pattern)" fill-opacity="0.7"></path><path d="M 50.0,115.0 C 75.0,115.0 125.0,15.0 150.0,15.0" stroke="#4682B4" stroke-width="8" fill="none"></path><circle cx="50.0" cy="115.0" r="10" stroke="#FFFFFF" stroke-width="4" fill="#FF0000"></circle><circle cx="150.0" cy="15.0" r="10" stroke="#FFFFFF" stroke-width="4" fill="#FF0000"></circle><g class="y-axis-line"><rect x="0" y="0" width="65" height="130" stroke="transparent" stroke-width="0" fill="transparent"></rect><text x="0" y="19.0" fill="transparent" stroke="transparent" font-size="25">2</text><text x="0" y="126.0" fill="transparent" stroke="transparent" font-size="25">1</text></g><g class="vert-line"><rect x="40.0" y="0" width="20" height="130" stroke="transparent" stroke-width="12" fill="transparent"></rect><text x="60.0" y="20" fill="transparent" stroke="transparent" font-size="30px">1</text></g><g class="vert-line"><rect x="140.0" y="0" width="20" height="130" stroke="transparent" stroke-width="12" fill="transparent"></rect><text x="160.0" y="20" fill="transparent" stroke="transparent" font-size="30px">2</text></g></svg></div></td>
```
machow, line 6: Hmm.. I wonder if there's a better way to test this simple case, but it might be a resolve down the road situations 😭

rich-iannone (approved):
Looks great, thank you!

PR #338, opened 2024-05-14
fix: minor formatting issues in the docs

jrycw:
This PR addresses minor formatting issues in the docs. 

Additionally, I noticed that some functions lack parentheses when browsing the docs. This might be because they are categorized as attributes rather than methods during rendering. Is there a way to improve this?

image


machow (approved):
LGTM, thanks for catching these!

PR #339, opened 2024-05-14
Enhance the test coverage

jrycw:
This PR tries to enhance our test coverage.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -1,10 +1,10 @@
-from great_tables._helpers import LETTERS, letters, pct, px, random_id, _get_font_stack
+from great_tables._helpers import LETTERS, letters, pct, px, random_id, _get_font_stack, FONT_STACKS
 import pytest
 
 
 @pytest.fixture
 def font_stack_names():
-    yield [
+    yield {
         "system-ui",
         "transitional",
         "old-style",
@@ -20,7 +20,7 @@ def font_stack_names():
         "antique",
         "didone",
         "handwritten",
-    ]
+    }
 
 
 def test_px():
@@ -62,7 +62,7 @@ def test_uppercases():
     assert set(bad_letters).difference(uppercases)
 
 
-def test_get_font_stack_raises(font_stack_names):
+def test_get_font_stack_raises():
     name = "fake_name"
     with pytest.raises(ValueError) as exc_info:
         _get_font_stack(name)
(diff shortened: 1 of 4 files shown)


tests/test_helpers.py
```diff
@@ -38,3 +60,144 @@ def test_uppercases():
...
+                "serif",
+            ],
+        ),
+        (
+            "handwritten",
+            ["Segoe Print", "Bradley Hand", "Chilanka", "TSCu_Comic", "casual", "cursive"],
+        ),
+    ],
+)
+def test_get_font_stack_add_emoji_false(name, font_stack):
```
machow, line 195: This test seems like it has to recode the data in `_get_font_stack()`, which is a bit cumbersome. WDYT of...

* pulling the data in `_get_font_stack` out of the function. For example, `_helpers.py` could have a top-level variable called `FONT_STACKS` that is a dictionary.
* since add emoji does a single action---extend the font list---how about a single test case, checking that add emoji did that in one situation?

This will have the effect that a couple of lines in `_get_font_stack()` can replace all the if/elif font name checks.

Here's an example of what FONT_STACKS my look like in `_helpers.py`:

```python
FONT_STACKS = {
    "system-ui": ["system-ui", "sans-serif"],
    "transitional": ["Charter", ...],
    ...
}

EMOJI_FONTS =  ["Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"]
```

This way, the test of add emoji can just check that the flag does something like...

```

# this could be wrong, but hopefully the gist of it makes sense :o 
# CASE 1:
assert _get_font_stack("system-ui", add_emoji=False) == FONT_STACK["system-ui"]
```

tests/test_helpers.py
```diff
@@ -38,3 +60,144 @@ def test_uppercases():
 
     bad_letters = "#$!^%#tables"
     assert set(bad_letters).difference(uppercases)
+
+
+def test_get_font_stack_raises(font_stack_names):
```
machow, line 65: It looks like `font_stack_names` might not be used here

machow (changes requested):
Thanks for working on test coverage! This is asking for changes beyond tests, so feel free to punt to me if you want.

I left a comment discussing whether pulling the font data outside of the `_get_font_stack()` function can...

* cut out some of the recoding of font data in the tests
* up test coverage by removing the big if/elif font name block

Thanks again for taking a look at this. It was really helpful to see `_get_font_stack()` surfaced, and to think of coverage here!

tests/test_helpers.py
jrycw: Good spot 😅.

tests/test_helpers.py
jrycw: @machow , thanks for your review. I'm not sure if the latest commit meets your expectations. I'm happy to hear your feedback if it doesn't.

machow (approved):
This looks great, thanks a ton! One tiny improvement could be adding a trailing comma to the shorter lists of fonts. E.g. 

* old: `["system-ui", "sans-serif"]`
* new: `["system-ui", "sans-serif",]`

This way, the list entries will be put onto newlines. But I'm happy to merge as is!

jrycw:
Thanks for sharing this tip with me. Actually, there are times when I get confused about why the formatter doesn't perform as I expected, but now I understand the reason.

machow:
Thanks again for this---this kind of stuff is so helpful!

Issue #340, opened 2024-05-14
`Polars` expressions unsupported for column selection


jrycw:
It appears that while we currently support column selection using `Polars` selectors, `Polars` expressions are not yet supported. Is this intentional? Should we consider adding support for `Polars` expressions as well?

```python
from great_tables import GT
import polars as pl
import polars.selectors as cs
from great_tables.data import exibble


lil_exibble = exibble[["num", "char", "fctr", "date", "time"]].head(4)
pl_df = pl.from_pandas(lil_exibble)

# ok
GT(pl_df).cols_move_to_start(columns=cs.starts_with("char"))

# TypeError: Unsupported selection expr type: <class 'polars.expr.expr.Expr'>
GT(pl_df).cols_move_to_start(columns=pl.col("char"))
```

jrycw:
I'm curious if we could implement this feature by changing the line from
https://github.com/posit-dev/great-tables/blob/7eae1f269a7d1f787de7a2c9198d775933794fd5/great_tables/_tbl_data.py#L322
to `if not isinstance(expr, (list, cls_selector, pl.Expr)):`.

machow:
Hey, thanks for raising this -- I think this was intentional, but supporting expressions seems fair to consider.

I think the rationale was:

* `pl.col("a")` is a selection step that's part of a broader API expressing computation (e.g. `pl.col("a") + pl.col("b")`)
* `cs.by_name("a")` seems equivalent to `pl.col()`, but its type is selector specific

However, since...

* `pl.col("a")` can still be fed to `cs.expand_selector()`
* selectors can also specify computation (e.g. `cs.by_name("a") + pl.col("b")`)

I could see how people might want to use `pl.col()`? It seems like where selection should occur was pretty actively discussed a while back (https://github.com/pola-rs/polars/issues/13757; see https://github.com/pola-rs/polars/issues/13757#issuecomment-1903966834 for good language on selection vs expression, etc..).

It might be a good idea to wait a bit and see where the API is a year from now. IMO it's helpful that requiring a selector nudges people away from expressions in column selection. But I could see why `pl.col()` is convenient (as evidenced by that issue).

As a slightly related point, I noticed that `cs.expand_selector()` actually returns the resulting column names. We use it to select columns, so this behavior could cause issues for us. I opened an issue in case it's a bug.

https://github.com/pola-rs/polars/issues/16242

jrycw:
[...]

PR #346, opened 2024-05-15
Enhance column selection to support `Polars` expressions

jrycw:
Related to #340.

This PR aims to extend column selection to support `Polars` expressions.

### DIFF
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -319,7 +319,7 @@ def _(data: PlDataFrame, expr: Union[list[str], _selector_proxy_], strict: bool 
     # it inherits from Expr, so we can just use that in a pinch
     cls_selector = getattr(selectors, "_selector_proxy_", Expr)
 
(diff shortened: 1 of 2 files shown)


machow:
Thanks for taking the time to implement this---I left some thoughts in the comments. I'm open to this, but trying to get a feel for where the selection API is going? (suspecting it might not change too much?). If it's okay with you, let's wait a bit and collect more info on the `pl.col()` vs selector situation.

jrycw:
@machow , thank you for your response. Let's hold off for now and see how things progress. I'll mark this PR as a draft for the time being. On another note, it appears that our project is gaining more attention from the community. I suggest that we tidy up the finished issues before the next minor release (0.6.0). This way, future contributors can easily identify and tackle the issues they want to help with.

jrycw:
I'll close this PR since the column selection of `Polars` is being refactored through #360 and is included in version 0.6.1.

PR #349, opened 2024-05-15
Docs: add example coffee table with nanoplots

machow:
This PR adds the coffee table example from https://github.com/machow/coffee-sales-data/

Note these extra code pieces:

* tiny fix: remove two spaces in the base64 encoding in `` tags. Necessary for displaying in jupyter notebooks on github. E.g.
  - old: `<img src="data: image/png...`
  - new: `<img src="data:image/png...`
* chore: remove polars map_elements warning by specifying output type

image


rich-iannone (approved):
LGTM!

Issue #96, opened 2023-12-18
Changelog in documentation page is not updated


karangattu:
## Description

The changelog listed at this [page](https://posit-dev.github.io/great-tables/changelog.html) shows the version `v0.0.2` but the current version published to PyPI is `v0.1.3`

## Reproducible example

Navigate to [page](https://posit-dev.github.io/great-tables/changelog.html) and see.

<img width="1676" alt="Cursor_and_great_tables_-_Changelog" src="https://github.com/posit-dev/great-tables/assets/4220325/647ecb02-9e7e-4d6c-aec1-7b179415a67d">

## Expected result

Should show all the changelog up to `v0.1.3`

PR #351, opened 2024-05-16
Remove unneeded files

rich-iannone:
This PR simply cleans out some unneeded top-level files. Some of these are redundant since more up-to-date versions are present in the `.github` directory. Others of these are redundant because the GitHub releases pages provide information on releases. The extra authors file is not needed because the author list is in `pyproject.toml`.

Fixes: https://github.com/posit-dev/great-tables/issues/96


machow (approved):
LGTM thanks for sprucing up the place!

PR #352, opened 2024-05-17
fix: issue with `_save_screenshot()`

jrycw:
It seems that we need to use `f.read()` to avoid the `TypeError: a bytes-like object is required, not '_io.BufferedReader'`.


machow:
Merged in main so we could re-run tests

machow (approved):
Thanks for catching---hope it's okay I chucked in a quick test (which I should have included in the original PR 😓 ). I double checked a pdf output for sanity and it looked good

PR #360, opened 2024-05-23
fix: refactor column selection logic for `Polars`

jrycw:
Related to #340, [pola-rs/polars#16242](https://github.com/pola-rs/polars/issues/16242), [pola-rs/polars#16250](https://github.com/pola-rs/polars/issues/16250).

This PR provides a quick fix for the column selection logic in `Polars`. Further investigation may be needed from the team.

### DIFF
--- a/docs/blog/polars-styling/index.qmd
+++ b/docs/blog/polars-styling/index.qmd
@@ -198,12 +198,12 @@ gt_with_spanners = (
 
     # Table column spanners ----
     .tab_spanner(
-        label = "Time",
-        columns = time_cols
+        label="Time",
+        columns=time_cols
     )
     .tab_spanner(
-        label = "Measurement",
-        columns = cs.all().exclude(time_cols)
+        label="Measurement",
+        columns=cs.exclude(time_cols)
     )
 )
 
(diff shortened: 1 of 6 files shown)


machow:
Shoot, thanks for catching this! It seems like there are three behaviors in the new polars release that broke our current approach (afaict this is just a side-effect of us living on the bleeding edge of `expand_selector()` 😈 ):

1. quick fix: `expand_selector()` no longer accepts a list
2. quick fix: `expand_selector()` no long accepts a raw string or int
3. bigger discussion: `expand_selector()` only sometimes considers `.exclude()` a selector

I've opened an issue on polars RE the `.exclude()` behavior:

* https://github.com/pola-rs/polars/issues/16448

machow (approved):
Thanks so much for the fixes (within 24 hours of the polars release 🤯 ). I hope it's okay, I tweaked a little bit to...

* try and give a detailed error for when a list has things that aren't selectors
* use functools reduce to accumulate selectors

We're going to make a small tweak to our CI, so we can quickly preview the docs builds for PRs off of forks, like this. And then should be able to merge!

machow:
Here is the docs preview I've pushed to a branch:

https://docs-preview-test--gt-python.netlify.app/articles/intro.html

jrycw:
@machow and @rich-iannone , everything looks great. It seems the developers from `Polars` are actively supporting and clarifying our use case, which is fantastic.

Issue #248, opened 2024-03-19
Using .data_color for a subset of rows


tomas19:
Is it possible to use the ```.data_color``` method for a subset of the rows?

I have a dataframe where the last row and columns are the sum of the values of each column and row, respectively. It is easy to exclude the last column from the colormap, specifying the column names in ```columns```, but I couldn't figure out how to exclude the last row. 

It should be possible to do it with the summary_rows(), but from the video below, apparently it is not implemented yet (?)
https://www.youtube.com/watch?v=p42Xc6dl3Ek 

Thanks!

rasmi:
Similar question here. The use cases is specifically to apply colors to a table, factoring in all values into the domain except the Grand Total. It is easy to exclude specific columns with the `columns=` argument, but there is no such argument for rows. One could compute the palette manually as is done in the underlying code, but it adds another step. In that sense, related to #172 and #173 for this specific use case.

rich-iannone:
Sorry for the delay in responding but `data_color()` needs a `rows` argument. We will try to get that in soon.

rasmi:
I did a hacky workaround for this here (may not work in many cases, e.g. assumes pandas, expects a list of ints):

Added a `rows` argument to `_data_color/base.py`'s `data_color` method:
```py
def data_color(
    self: GTSelf,
    columns: Union[str, List[str], None] = None,
    palette: Union[str, List[str], None] = None,
    domain: Union[List[str], List[float], List[int], None] = None,
    na_color: Optional[str] = None,
    alpha: Optional[Union[int, float]] = None,
    reverse: bool = False,
    autocolor_text: bool = True,
    rows: Optional[List[int]] = None,
) -> GTSelf:
```

Further down in the `data_color` method, grab the desired rows:
```py
    # Get the internal data table
    data_table = self._tbl_data
    # Use only selected rows.
    if rows:
      data_table = data_table.copy().iloc[rows]
```

Not ideal but it allowed me to create and render my table in the interim!

rich-iannone:
Completed with https://github.com/posit-dev/great-tables/pull/364.

PR #364, opened 2024-05-25
Support specifying a subset of rows in `GT.data_color()`

jrycw:
Related to #248.

This PR aims to support specifying a subset of rows in `GT.data_color()`. Based on the discussions, it appears that the most common use case involves excluding the last row, which is typically the total row. Consequently, rows that are not selected should not be included in the color "calculation." In this PR, I first select only the specified columns and rows, then "re-map" the "calculated color" back to the original table.

A simple example might illustrate the concept well:
```python
import polars as pl
import polars.selectors as cs
from great_tables import GT

df = pl.DataFrame({"a": [1, 2, 3, 4, 5, 200], "b": [51, 52, 53, 54, 55, 200]})

GT(df).data_color(columns=cs.numeric(), rows=pl.col("a").lt(6), palette="PuBu")
```
image

```python
GT(df).data_color(columns=cs.numeric(), rows=pl.col("a").lt(300), palette="PuBu")
```
image

### Bonus
In addition, I have added support for using Polars-style column and row selection. This allows for more flexible operations, such as:
```python
GT(df).data_color(columns=cs.starts_with("b"), rows=pl.col("a").lt(6), palette="PuBu")
```
image

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -21,7 +21,7 @@ RGBColor: TypeAlias = tuple[int, int, int]
 def data_color(
     self: GTSelf,
     columns: SelectExpr = None,
-    rows: RowSelectExpr = None,
+    include_last_row: bool = True,
     palette: str | list[str] | None = None,
     domain: list[str] | list[int] | list[float] | None = None,
     na_color: str | None = None,
@@ -49,10 +49,10 @@ def data_color(
     columns
         The columns to target. Can either be a single column name or a series of column names
         provided in a list.
-    rows
-        In conjunction with `columns=`, we can specify which rows should be colored. By default,
-        all rows in the targeted columns will be colored. Alternatively, we can provide a list
-        of row indices.
+    include_last_row
+        Whether to include the last row for color styling. The default is set to `True`, but it
+        can be useful to set it to `False` if the last row is a summary or subtotal row of the
+        table.
     palette
         The color palette to use. This should be a list of colors (e.g., `["#FF0000", "#00FF00",
         "#0000FF"]`). A ColorBrewer palette could also be used, just supply the name (reference
@@ -213,8 +213,12 @@ def data_color(
     else:
         columns_resolved = resolve_cols_c(data=self, expr=columns)
 
-    row_res = resolve_rows_i(self, rows)
+    row_res = resolve_rows_i(self)  # get all rows
     row_pos = [name_pos[1] for name_pos in row_res]
+    if not include_last_row:
+        # Excluding the last row for now;
+        # adjustments may be needed after implementing `summary_rows()`.
+        row_pos = row_pos[:-1]
 
     gt_obj = self
 
@@ -268,7 +272,7 @@ def data_color(
 
         # for every color value in color_vals, apply a fill to the corresponding cell
         # by using `tab_style()`
-        for i, color_val in zip(row_pos, color_vals):
+        for i, color_val in enumerate(color_vals):
             if autocolor_text:
                 fgnd_color = _ideal_fgnd_color(bgnd_color=color_val)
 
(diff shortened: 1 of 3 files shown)


great_tables/_data_color/base.py
```diff
@@ -258,21 +266,25 @@ def data_color(
...
-
-            else:
-                gt_obj = gt_obj.tab_style(
-                    style=fill(color=color_val), locations=body(columns=col, rows=[i])
-                )
+        # For each row, we check if the row index is selected, and then apply a fill to the
+        # corresponding cell using `tab_style()`
+        n_rows = len(resolve_rows_i(self))
+        iter_color_vals = iter(color_vals)
+        for i in range(n_rows):
```
machow, line 273: Would something like...

```python
for i, color_val in zip(row_pos, color_vals):
    ...
```

work? (in this case, you wouldn't need to check `if i in row_pos`)

machow (changes requested):
Thanks, this looks really great! I left one small question on whether we might be able to tweak the for loop a little bit, but it's very possible I overlooked something there.

great_tables/_data_color/base.py
jrycw: @machow Clever suggestion! I believe this code will work better.

jrycw:
Please hold off on merging this PR for now. I've discovered some personal notes in my Jupyter playbook explaining why I wrote the code this way. It appears that the logic between these two implementations may differ. I'll conduct further investigation soon.

jrycw:
I think both approaches should yield the same result. However, a more interesting question arises: what should users expect if we include the `rows` parameter? For instance, if `rows = [1, 3, 4]`, should we first calculate the palette for the entire column and then retrieve the corresponding colors for those rows, or should we calculate the palette based solely on the selected rows? Both cases might not look aesthetically pleasing, and I doubt their practical application. 

Perhaps what we really need is an option like `include_last_row=True` or `include_last_row=False` (the parameter name might not be optimal, but you get the idea). I've implemented alternative logic in the last commit, although this logic might need adjustments in the future if we introduce something like `GT.summary_rows()`, which could potentially affect more than one row.

```python
import polars as pl
import polars.selectors as cs
from great_tables import GT

df = pl.DataFrame(
    {
        "a": [1, 2, 3, 4, 5, sum([1, 2, 3, 4, 5])],
        "b": [51, 52, 53, 54, 55, sum([51, 52, 53, 54, 55])],
    }
)


GT(df).data_color(columns=cs.numeric(), include_last_row=True, palette="PuBu")
```

image

```python
GT(df).data_color(columns=cs.numeric(), include_last_row=False, palette="PuBu")
```

image

machow:
> should we first calculate the palette for the entire column and then retrieve the corresponding colors for those rows, or should we calculate the palette based solely on the selected rows?

@rich-iannone do you know what happens in the gt R package? AFAICT data_color there doesn't use excluded rows, and the rationale is that this let'd you exclude extremely large values etc..

I would lean toward not using excluded in the calculations. Then, if people want to use all rows, they can always set the `domain=` argument, to specify the exact range for the color values.

rich-iannone:
> > should we first calculate the palette for the entire column and then retrieve the corresponding colors for those rows, or should we calculate the palette based solely on the selected rows?
> 
> @rich-iannone do you know what happens in the gt R package? AFAICT data_color there doesn't use excluded rows, and the rationale is that this let'd you exclude extremely large values etc..
> 
> I would lean toward not using excluded in the calculations. Then, if people want to use all rows, they can always set the `domain=` argument, to specify the exact range for the color values.

The R package doesn't use the excluded rows in the calculations or colorization. You're right that the `domain` argument approach is better for specifying the exact range. If we did include all the rows in the calculation, it might be surprising to users and I don't think it would be generally desired behavior through an extra argument.

jrycw:
Thanks @machow and @rich-iannone . Would [43b0ce5](https://github.com/posit-dev/great-tables/pull/364/commits/43b0ce5c6e5edc1280bcf13b6be8e37a615fc37f) achieve our goal, or are there any other modifications I should make?

rich-iannone:
I think so. Just reverting the last commit would make this right.

machow (approved):
LGTM, thanks for thinking through the different options here!

PR #365, opened 2024-05-26
Refactor `seq_groups()` to accept `Iterable`

jrycw:
Our project is progressing steadily, and it's been an enjoyable journey as one of the contributors. However, I've noticed that we might be overly reliant on the `list` type in the codebase. While `list` is excellent and suitable for prototyping, it can become problematic as the codebase grows. Therefore, I propose replacing `list` with `Generator` where possible.

I have chosen `seq_groups()` as a starting point since it is currently functioning as a kind of `Generator`. Currently, we request the input as a `list`, but I believe we can improve this by refactoring it to accept an `Iterable` as input. 

I hope this PR makes sense, and I welcome input from the team on this concept.

### DIFF
--- a/great_tables/_spanners.py
+++ b/great_tables/_spanners.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 import itertools
-from typing import TYPE_CHECKING, Any
-
+from typing import TYPE_CHECKING, Any, Iterable
+from collections.abc import Generator
 from ._gt_data import SpannerInfo, Spanners
 from ._locations import resolve_cols_c
 from ._tbl_data import SelectExpr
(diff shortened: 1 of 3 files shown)


great_tables/_spanners.py
```diff
@@ -565,26 +565,49 @@ def empty_spanner_matrix(
...
+    https://docs.python.org/3/library/itertools.html#itertools.pairwise
+    pairwise('ABCDEFG') → AB BC CD DE EF FG
+    """
+    # This function can be replaced by `itertools.pairwise` if we only plan to support
+    # Python 3.10+ in the future.
+    iterator = iter(iterable)
+    a = next(iterator, None)
+    for b in iterator:
+        yield a, b
+        a = b
```
machow, line 579: Is this `a = b` a typo?

machow (approved):
Thanks for this PR! I think these changes look good, and are a nice improvement. However, I do suspect that in general we won't get a lot of practical gains out of optimizing list code to iterators.

The two reasons I suspect for this are:

* The data volumes Great Tables processes are smaller than most DataFrame libraries need to handle (since the tables are for display to humans).
* For anything column related, the number of entries tends to be small

Have you tried snakeviz? It's a handy tool for checking what is taking a long time to run. 

For example, in a jupyter notebook...

```python
import pandas as pd

ex2 = pd.concat([exibble]*100)
%load_ext snakeviz


# new cell ----
%%snakeviz

GT(ex2, rowname_col="row", groupname_col="group").tab_spanner("A", ["fctr", "date"])
```

image

It looks like a surprising amount of time is spent on init, especially in the `align_from_data(...) function call, which is doing a lot of regexing. I haven't looked at this code, but it looks like it's iterating over all the data values for columns of dtype object and performing a regex on every string value.

(This isn't necessarily bad behavior, @rich-iannone can say more, and slow in this case means only ~10 milliseconds for the function!).

great_tables/_spanners.py
jrycw: I believe it's not. Let's take "ABCDEFG" as an example. First, we get the first element, `a`, which is "A". At this point, "BCDEFG" remains in the iterator. 

Next, we iterate through the iterator and get the next element, `b`, which is "B", leaving "CDEFG" in the iterator. We then yield the tuple ("A", "B") and set `a` to `b`, so `a` now equals "B".

We continue iterating through the iterator. The next element is "C", with "DEFG" remaining in the iterator. We yield the tuple ("B", "C"), and the process repeats in the same manner.

jrycw:
> snakeviz

From a performance perspective, using an `iterator` might not offer significant gains. However, there are notable benefits. We avoid the need to convert data structures into `lists` just for indexing with `[]`. By accepting any `iterable`, we maintain flexibility, a concept I learned from the `itertools` module. This approach can also save memory since we don't need to materialize everything into a list, even though memory consumption might not be a major concern for processing small data in our use case. By the way, thanks for introducing me to `snakeviz`, another new and useful package.

great_tables/_spanners.py
machow: ah shoot--that makes sense! Sorry, I spaced very briefly on how iterators work 😓

machow:
These are all fair points, especially that memory usage tends to be better with iterators. I definitely think there will be cases where iterators lead to big improvements. The one thing I feel cautious about when considering iterators, is that developers tend to use lists, even in situations when they could have used iterators.

I think there's likely something about lists that people find a bit easier to understand, that tends to push them in that direction (whether it's the ease of seeing the data when debugging vs a stream, or something else). It seems to be on the desire-path for developers.

In any event, at this point I think @rich-iannone and I trust your judgement on which pieces could be improved by iterators. So definitely don't take this as discouraging changing anything where it seems like they'd be useful :)

jrycw:
@machow, thanks for sharing your thoughts. I completely agree with your points. I would aim for a function that:

* Takes a list as input primarily for its indexing behavior.
* Has a downstream function that only needs to iterate through the returned value once without checking its length or performing further indexing.

Under these conditions, it might be a good candidate to take an iterator as input and return an iterator as the output.

PR #366, opened 2024-05-26
Enhance `cols_*()` functions

jrycw:
This PR aims to enhance the `cols_*()` functions by:

* Abstracting the validation of `sel_cols` into a separate function and adding related tests.
* Suggesting renaming `vars` to `col_vars` in the function. While `vars` might be acceptable in the function signature, it’s better to avoid using `vars` as a temporary variable name if possible.
* Updating the syntax for constructing the `DataFrame` in the `cols_*()` test functions.

### DIFF
--- a/great_tables/_spanners.py
+++ b/great_tables/_spanners.py
@@ -202,6 +202,13 @@ def tab_spanner(
     return new_data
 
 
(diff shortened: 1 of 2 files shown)


machow (approved):
This looks great, thanks!

PR #368, opened 2024-05-29
feat: support polars non-strict expand_selector

machow:
This PR adds support for polars expressions to be used in column selection in addition to polars selectors. It requires polars v0.20.30, but falls back to the old behavior (or requiring a selector object) otherwise.

Note that this is the result of a chain of things:

* https://github.com/posit-dev/great-tables/pull/360
* https://github.com/pola-rs/polars/issues/16448

Which resulted in polars super quickly adding broader support for accepting Expressions and selectors that only result in column selection (and not e.g. transformation, or aliasing; PR https://github.com/pola-rs/polars/pull/16479).

I tried to give reasonably helpful error messages for people who are not on v0.20.3 that use Expr, but also suspect people will upgrade polaras patch versions fairly quickly.

## Example

```python
from great_tables import exibble, GT

GT(pl.from_pandas(exibble)).tab_spanner("A", [pl.col("fctr"), pl.all().exclude("date")])
```

Note both that a list is supported along with the use of expressions (e.g. `pl.col("fctr")`)

### DIFF
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -283,6 +283,7 @@ def _(data: PlDataFrame, group_key: str) -> dict[Any, list[int]]:
 SelectExpr: TypeAlias = Union[
     list["str | int"],
     PlSelectExpr,
+    list[PlSelectExpr],
     str,
     int,
     Callable[[str], bool],
@@ -329,7 +330,7 @@ def _(data: PlDataFrame, expr: Union[list[str], _selector_proxy_], strict: bool 
     from polars import Expr
 
     pl_version = _re_version(version("polars"))
-    expand_opts = {"strict": False} if pl_version >= (0, 20, 3) else {}
+    expand_opts = {"strict": False} if pl_version >= (0, 20, 30) else {}
 
     # just in case _selector_proxy_ gets renamed or something
     # it inherits from Expr, so we can just use that in a pinch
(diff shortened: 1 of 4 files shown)


machow:
@jrycw if you have a min, do you mind taking a look? Hopefully this rounds out our polars column selection work!

rich-iannone (approved):
LGTM!

jrycw:
@machow it's amazing how quickly `Polars` is supporting our use case. I need to grab a coffee and take a detailed look at this, as it might impact how users utilize it.

great_tables/_tbl_data.py
```diff
@@ -313,38 +329,64 @@ def _(data: PlDataFrame, expr: Union[list[str], _selector_proxy_], strict: bool
     from operator import or_
```
jrycw, line 329: Are these legacy imports?

great_tables/_tbl_data.py
```diff
@@ -313,38 +329,64 @@ def _(data: PlDataFrame, expr: Union[list[str], _selector_proxy_], strict: bool
...
     if isinstance(expr, list):
+        # convert str and int entries to selectors ----
         all_selectors = [
             cs.by_name(x) if isinstance(x, str) else cs.by_index(x) if isinstance(x, int) else x
             for x in expr
         ]
 
-        _validate_selector_list(all_selectors)
+        # validate all entries ----
+        _validate_selector_list(all_selectors, strict=False if expand_opts else True)
```
jrycw, line 350: Personally, I would call `_validate_selector_list(all_selectors, **expand_opts)` directly.

jrycw (commented):
@machow Overall, it looks good. We might need to update the documentation after merging this PR.

great_tables/_tbl_data.py
jrycw: I'm referring to the lines `from functools import reduce` and `from operator import or_`.

great_tables/_tbl_data.py
machow: ah, good catch!

PR #371, opened 2024-05-31
Feat init methods

machow:
This WIP PR will enable options that currently must be set on construction of GT() to be set with methods.

The options are...

* rowname_col
* groupname_col
* auto_align
* id
* locale

The most challenging options are `rowname_col` and `groupname_col`, since setting these options means checking, and potentially undoing a number of things.

For example, a column that is being set to `rowname_col`, should:

* remove itself from any Spanner variable lists
* change its type in the Boxhead
* update the names in Stub.rows

Similarly, a new `groupname_col` will need to remove any styles that were set on it (when it was in the body).

## Main changes

This PR required a dive into components like Stub. While I was there, it seemed like a good idea to consolidate things into it, so that GT can't get into funky states, and it's clear which class drives things like final row order. This makes things a bit less noodly when updating things like groupname_cols(), since the Stub and Boxhead can now handle the whole task!

Big items:

* Refactor Stub to contain what were prevoiusly RowGroups and GroupRows
  - these classes were tightly coupled, and it's important they don't get out of sync
  - previously, the Stub was reordered but this was never used, and the GroupRows ended up determining order. Now, the Stub doesn't reorder rows and ultimate handles final display order (through its control of GroupRows).
* Feat: add `.row_group_order()` method
* Fix: support missing values in the grouping column for pandas
* ~add `.with_rowname_col()`, `with_groupname_col()`, `with_locale()`, and `.with_id()`~
* feat: add `.tab_stub()`, `.with_locale()`, `.with_id()`
* ~TODO: handle auto_align~

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -47,7 +47,6 @@ class GTData:
     _body: Body
     _boxhead: Boxhead
     _stub: Stub
-    _group_rows: GroupRows
     _spanners: Spanners
     _heading: Heading
     _stubhead: Stubhead
@@ -94,7 +93,6 @@ class GTData:
             _body=Body.from_empty(data),
             _boxhead=boxhead,  # uses get_tbl_data()
             _stub=stub,  # uses get_tbl_data
-            _group_rows=stub.group_rows,
             _spanners=Spanners([]),
             _heading=Heading(),
             _stubhead=None,
@@ -115,8 +113,8 @@ class GTData:
         id: str | None = None,
         locale: str | None = None,
     ):
-        stub, boxhead, group_rows = _prep_gt(self._tbl_data, rowname_col, groupname_col, auto_align)
-        row_groups = stub.group_ids
+        stub, boxhead = _prep_gt(self._tbl_data, rowname_col, groupname_col, auto_align)
+
         # set id if not None
         # self._options
         locale = Locale(locale) if locale is not None else self._locale
@@ -129,8 +127,6 @@ class GTData:
         return res._replace(
             _stub=stub,
             _boxhead=boxhead,
-            _group_rows=group_rows,
-            _row_groups=row_groups,
             _locale=locale,
         )
 
@@ -599,6 +595,9 @@ class Stub:
         new_rows = [self.rows[ii] for ii in indices]
 
         return self.__class__(new_rows, self.group_rows)
+
+    def group_indices_map(self) -> list[tuple[int, str | None]]:
+        return self.group_rows.indices_map(len(self.rows))
 
     def __iter__(self):
         return iter(self.rows)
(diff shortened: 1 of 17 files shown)


machow:
@rich-iannone, WDYT of these alternative approaches for method names:

1. the current approach, all init args now available via `.with_*()` methods)
  - `.with_rowname_col(name)`
  - `.with_groupname_col(name)`
  - `.with_locale(locale)`
  - `.with_id(id)`
  - `.with_autoalign(...)`
2. exposing something like `.tab_stub(rowname_col=..., groupname_col=...)`
  - `.tab_stub(...)`
  - `.with_locale(...)`
  - `.with_id(...)`
  - `.with_autoalign(...)`
3. alternatively, could name other things `with_context(...)` or something...
  - `.tab_stub(...)`
  - `.with_context(locale=, id=, autoalign=)`

My suspicion is that (1) will start to feel cumbersome, because setting both row and group names feels very common. I like that (2) hoists that activity up to a top-level structure activity, with the other tab methods. However, the consistency of just using `with_` seems nice? 🤷🏼

machow:
One smaller piece to resolve: `auto_align` is currently set to True. So it's possible to expose options to not auto_align in the final render, but that won't stop GT(some_data) from running alignment logic. We might need some kind of global option or something.

great_tables/_gt_data.py
```diff
@@ -514,58 +527,108 @@ class RowInfo:
...
 
     def _get_stub_components(self) -> list[str]:
         stub_components: list[str] = []
 
-        if any(entry.group_id is not None for entry in self):
+        if any(entry.group_id is not None for entry in self.rows):
             stub_components.append("group_id")
 
-        if any(entry.rowname is not None for entry in self):
+        if any(entry.rowname is not None for entry in self.rows):
```
jrycw, line 623: @machow , I'm just browsing through the code. It looks great, but I suspect you could iterate over `self` directly without changing these lines, since `__iter__` is already implemented.

machow: Ah, nice catch! Since Stub now has two pieces inside it (rows and group_rows), I wasn't sure whether it should allow iterating over it, or require grabbing rows directly. I think the `__iter__` was to avoid breaking code while refactoring. WDYT, is it worth keeping `__iter__`?

machow:
Alright I think the key stages of this PR were:

* refactor group rows, row groups into Stub: https://github.com/posit-dev/great-tables/pull/371/files/34f1bb97740d28cefc60cd3c7b2bce7ef19c1d18
* add new methods: https://github.com/posit-dev/great-tables/pull/371/files/88e453fb9cb45ce396671dd1c2335aa17c0b8153..2d94af0aa52647cdbd99915ced0d8042453d3ff4

rich-iannone (approved):
This is amazing work! Everything looks good to me.

great_tables/_gt_data.py
jrycw: I see. Personally, I would prefer keeping `__iter__`.

machow:
I tucked in some quick tweaks to docs to mention `.tab_stub()`, and swapped it into our examples!

Issue #100, opened 2023-12-25
Return browsable gt object in Python repl


jooyoungseo:
Thanks for this great package! I am suggesting one feature request below.

In R {gt} package, `gt()` object returns browsable table in user's default browser. This is very convenient way for screen reader users to navigate data table in an alternative way. Could you please consider this in Python package as well?

Currently, `GT()` method returns html code in Python repl. Browsable GT object might be more useful in an interactive terminal from end-user's perspective.

machow:
Thanks for looking into this issue! I've been looking around for python libraries that do something similar, and it seems like plotly might be a good example of this:

```python
import plotly.express as px

fig = px.scatter(x=[0, 1, 2, 3, 4], y=[0, 1, 4, 9, 16])

# opens a browser window
fig.show()
```

Since the gt for R opens a browser aimed at a temporary .html file, we should be able to do the same. WDYT of that approach? In python, I wonder if we should add a method, like `.preview()`, that triggers this behavior.

I think the `.__repr__` method triggering a browser window to open might be surprising to python users but a `.preview()`. For plotly, it looks like they do trigger a browser window when you dump to console, but only for IPython, not for the python repl (there, you have to use `fig.show()`)...

jooyoungseo:
Maybe, I could be wrong, but in R, {gt} seems to rely on {htmltools} and `gt::gt()` appears to return `htmltools::browsable()` object.

I created a feature request https://github.com/posit-dev/py-htmltools/issues/79 to ask if {py-htmltools} could possibly support `htmltools.browsable()` methods. It would be great if {great_tables} depends on {py-htmltools} once it gets more reliable.

jooyoungseo:
@machow -- I have just learned from @cpsievert that we could use `show()` methods against {py-htmltools} object. What if {great_tables} depends on {py-htmltools} for this feature and beyond?

machow:
Great Tables uses htmltools, so this is perfect! I was able to get it to appear in the browser using this code in the IPython console...

```python
from great_tables import GT, exibble
from htmltools import HTML, TagList

res = GT(exibble).render("html")
el = HTML(res)

TagList(el).show("browser")
```

Note that I had to pass "browser" as an argument. The default argument of "auto" didn't seem to do anything.

It seems like at the very least we should add a `.show()` or `.preview()` method, which holds the logic for viewing the table anywhere (jupyterlab, IPython console, etc..). This could use something similar to the code above.

cpsievert:
[...]

PR #379, opened 2024-06-17
Feat gt show

machow:
This PR addresses #100 by adding a `GT.show()` method, which does the following:

* notebook-like environment: display table in cell output
* repl environment (e.g. ipython shell, python repl): open a browser with table

Fixes: #100

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -58,6 +58,7 @@ filters:
   - interlinks
 
 interlinks:
+  fast: true
   sources:
     numpy:
       url: https://numpy.org/doc/stable/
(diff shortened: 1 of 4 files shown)


machow:
One trick with this pull request is that we've tested paths up to running `get_ipython()`, and it's tricky testing the cases where (1) we display in ipython, or (2) some of the bits around serving the table in the browser.

We use `.show()` in our executed docs code, which helps as a kind of smoketest. So the codecov is low. I think hopefully we've gotten close enough to things with our tests, but we might need to either eat codecov being mad or mark some of the parts as deliberately untested in unit tests.

rich-iannone (approved):
LGTM!

rich-iannone:
I’m generally in favor with marking parts that cannot be tested easily with comments signaling it should be omitted from codecov consideration.

PR #380, opened 2024-06-18
Feat: add ability to express units in `.cols_label()`

rich-iannone:
This PR enhances the `.cols_label()` method by allowing units (in a specialized *units notation*) to be combined with label text. The result of this is the ability to have a column label suitable for presentation that also has units information that relate to the figures below a column. The *units notation* text is set off from the regular label text with `"{{"` / `"}}"`.

Here's a basic example for how this works:

```python
import pandas as pd
from great_tables import GT

df = pd.DataFrame({"x": [1.234, 2.345], "y": [3.456, 4.567], "z": [5.678, 6.789]})

GT(df).cols_label(x="Area ({{m^2}})", y="Density ({{kg / m^3}})", z = "Zee")
```


great_tables/_boxhead.py
```diff
@@ -99,7 +100,18 @@ def cols_label(self: GTSelf, **kwargs: str) -> GTSelf:
...
     _assert_list_is_subset(mod_columns, set_list=column_names)
 
-    boxhead = self._boxhead._set_column_labels(kwargs)
+    # Handle units syntax in labels (e.g., "Density ({{ppl / mi^2}})")
+    new_kwargs: dict[str, UnitStr | str] = {}
+
+    for k, v in kwargs.items():
+        # If the text is a string with has units notation within (detectable by "{{" and "}}")
+        # convert it to a UnitStr object
+        if isinstance(v, str) and ("{{" in v and "}}" in v):
```
machow, line 109: Can't a UnitStr be created from any string? Is it necessary to check for "{{"? This will also flag the case `}} {{` which is not a unit pattern.

great_tables/_gt_data.py
```diff
@@ -1044,6 +1062,9 @@ class Options:
         True, "column_labels", "value", "#D3D3D3"
     )
     column_labels_hidden: OptionsInfo = OptionsInfo(False, "column_labels", "boolean", False)
+    column_labels_units_pattern: OptionsInfo = OptionsInfo(
```
machow, line 1065: This appears unused please remove if so

great_tables/_helpers.py
```diff
@@ -759,12 +749,62 @@ def to_html(self):
         return units_str
 
 
+class UnitStr:
```
machow, line 752: This is untested please test (class construction and methods)

great_tables/_options.py
```diff
@@ -335,6 +336,10 @@ def tab_options(
     column_labels_hidden
         An option to hide the column labels. If providing `True` then the entire `column_labels`
         location won't be seen and the table header (if present) will collapse downward.
+    column_labels_units_pattern
```
machow, line 339: See comment on being unused. What happens if people set this? If it currently doesn't do anything then we need to set expectations.

great_tables/_text.py
```diff
@@ -45,24 +52,23 @@ def _md_html(x: str) -> str:
 
 
 def _process_text(x: str | Text | None) -> str:
+    from great_tables._helpers import UnitStr
```
machow, line 55: needs tests. If UnitStr isn't a subclass of text it needs to be added to the type hint on line 54

great_tables/_gt_data.py
```diff
@@ -407,6 +413,18 @@ def _set_column_labels(self, col_labels: dict[str, str]) -> Self:
 
         return self.__class__(out_cols)
 
+    # Set column units
+    def _set_column_units(self, col_units: dict[str, str]) -> Self:
```
machow, line 417: Is this used? if not please remove

machow (changes requested):
Can you provide some examples (in a gh response is okay) illustrating each new feature this PR provides?

Do you mind adding tests (I've left comments to flag)? I've noticed a number of places where code was added but appears unused. Can you remove any unused code?

great_tables/_boxhead.py
rich-iannone: From pairing, we changed this to use logic from `UnitStr`.

great_tables/_gt_data.py
rich-iannone: Not used any more. Just removed.

great_tables/_gt_data.py
rich-iannone: Now removed.

great_tables/_options.py
rich-iannone: Now removed as it doesn't do anything now (with the other removals).

great_tables/_text.py
rich-iannone: A number of tests for `UnitStr` were added during pairing.

machow:
@rich-iannone I think this branch might have accidentally been merged directly into main

machow (approved):
Do you mind adding a units example to the cols_label() docstring? (I flagged it with TODO:). But other than that, it looks great! Feel free to merge after the example is in!

PR #382, opened 2024-06-21
Feature: Add several new datasets (bringing total number up to 16)

rich-iannone:
This adds several new datasets to the package: `peeps`, `films`, `gibraltar`, `reactions`, `photolysis`, and `nuclides`.

We also update the CSV for the `metro` dataset (to incorporate new stations that opened on 2024-06-13).

Some wording inside the dataset descriptions was changed to better describe the types of data within the various datasets.


machow (changes requested):
Excited for these! From pairing, I was going to suggest zipping some of these, but just noticed that pizzaplace (already in data) is the largest at 3Mb. I definitely think we can merge as is, and then zip down the road as an enhancement.

Do you mind adding a test file that just imports each dataset, and verifies it's a DataFrame?

You should be able to do something like...

```python
from great_tables import data

@pytest.mark.parametrize("name", ["pizzaplace", "peeps",])
def test_data(name):
    df = getattr(data, name)
    assert isinstance(df, pd.DataFrame)
```

machow (approved):
LGTM thanks!

Issue #381, opened 2024-06-19
`GT.show()`: Exception: Unknown target display: browser


jooyoungseo:
Thanks for working on `.show()` method. I have tested with the following code:

```python
from great_tables import GT
from great_tables.data import sp500

GT(sp500.head()).show()
```

I see the following result in console:

```bash
<socket.socket fd=2528, family=2, type=1, proto=0, laddr=('127.0.0.1', 64221), raddr=('127.0.0.1', 64222)>
<class 'socket.socket'>
127.0.0.1 - - [19/Jun/2024 09:18:26] "GET /index.html HTTP/1.1" 200 -
 -
  File "<stdin>", line 1, in <module>
  File "C:\Users\jseo1005\AppData\Local\miniconda3\Lib\site-packages\great_tables\_export.py", line 123, in show
    raise Exception(f"Unknown target display: {target}")
Exception: Unknown target display: browser
>>>
```

rich-iannone:
Fixed by https://github.com/posit-dev/great-tables/pull/384.

PR #384, opened 2024-06-21
fix: show no longer raises or prints to stderr

machow:
This PR addresses https://github.com/posit-dev/great-tables/issues/381, by...

* not raising at the end of every `.show()` call
* not logging http request info to stderr (which `http.server` handlers do by default)


machow:
@rich-iannone do you mind trying this locally, in a notebook and then in the IPython repl?

```

GT(exibble).show()
```

from the repl it should open in the browser (and not raise an error afterward).

rich-iannone (approved):
LGTM!

Issue #283, opened 2024-04-12
fmt_nanoplot includes excluded rows when scaling plot


rasmi:
I am using fmt_nanoplot to create a table with a histogram on the margins. To do this, I want to exclude the last row, which is a Total row. When I exclude the row using the `rows=` parameter, the scale of the plot appears as if it still factors in the row that was not plotted. I would expect that the plot scales only with the values included in the plot.

## Reproducible example

(adapted from [docs](https://posit-dev.github.io/great-tables/reference/GT.fmt_nanoplot.html#great_tables.GT.fmt_nanoplot))
![Screenshot 2024-04-12 14 20 51](https://github.com/posit-dev/great-tables/assets/2267370/7fabe130-4457-4606-aa43-3c6e72f0107c)

```py
import polars as pl
single_vals_df = pl.DataFrame(
    {
        "i": list(range(1, 6)),
        "lines_small": [12.44, 6.34, 5.2, -8.2, 10],
        "lines_large": [12.44, 6.34, 5.2, -8.2, 100]
    }
)
(
    GT(single_vals_df, rowname_col="i")
    .fmt_nanoplot(columns="lines_small", rows=[0,1,2,3], plot_type="line")
    .fmt_nanoplot(columns="lines_large", rows=[0,1,2,3], plot_type="line")
)
```

## Development environment

- Operating System: Colab notebook
- **great_tables**: 479eb732c912cb5062c4d7e61c6b31301787b00c (v0.5.0)

PR #404, opened 2024-07-18
Issue 283 nano plot scaling

minimav:
# Summary

This PR fixes a bug whereby values from unselected rows affect the scaling in nanoplots. In addition to the single values case given in #283, a simple example of the bug in the context of multiple values per row is below (when `autoscale` is used):

```python
multiple_vals_df = pd.DataFrame(
    {
        "i": list(range(1, 3)),
        "lines_small": ["12.44 6.34", "5.2 -8.2 10"],
        "lines_large": ["12.44, 6.34", "5.2 -8.2 100"],
    }
)
(
    GT(multiple_vals_df, rowname_col="i")
    .fmt_nanoplot(columns="lines_small", rows=[0], plot_type="line", autoscale=True)
    .fmt_nanoplot(columns="lines_large", rows=[0], plot_type="line", autoscale=True)
)
```

Screenshot 2024-07-18 at 20 49 20

Single values example post-fix:

Screenshot 2024-07-18 at 19 53 06

Multiple values example post-fix:

Screenshot 2024-07-18 at 19 53 29

I have added a couple of unit tests based on existing ones, let me know if they work. Thanks for your work on this library, I have been using it at work and really enjoying it 🙂 

- Ref: #283 

# Checklist


machow:
I'm so sorry it took us so long to get to this. @rich-iannone do you mind looking through? If needed, we may just need to start a new PR to address, since we've made a lot of changes since then.

rich-iannone:
This PR fixes the issue that `fmt_nanoplot()` used *all* rows for scaling of single-value nanoplots (one value per cell), even when `rows=` restricted which rows received nanoplots. This meant that an extreme value in a non-selected row would distort the scale of the selected rows' plots.

With this PR, here's an example where we generate nanoplots in a column of values that are common but exclude the last value (which are different between the two DFs, and very large in `df_extreme`).

```{python}
import polars as pl
from great_tables import GT

df_normal = pl.DataFrame({"vals": [-5.3, 6.3, 5.2, -8.2, 10]})
df_extreme = pl.DataFrame({"vals": [-5.3, 6.3, 5.2, -8.2, 500.0]})

gt_normal = GT(df_normal).fmt_nanoplot(
    columns="vals",
    plot_type="bar",
    rows=[0, 1, 2, 3],
)

gt_extreme = GT(df_extreme).fmt_nanoplot(
    columns="vals",
    plot_type="bar",
    rows=[0, 1, 2, 3],
)
```

In the first table (`gt_normal`), the excluded value is `10` and so it's difficult to notice any scale effects.

```python
gt_normal.show("browser")
```

image

However, the same values that exclude a large value of `500` show the same bar plots. The extreme value of `500` isn't considered in the scaling of the other values (they would appear compressed to the zero line).

```python
gt_extreme.show("browser")
```

image

The unit tests compare single SVGs from two different DataFrames (with different unselected values). The equivalence check of the first row's SVG across the DFs indirectly shows that there aren't scaling effects from unselected cells.

rich-iannone (approved):
LGTM!

PR #415, opened 2024-08-07
Feat: DataFrame functions subset_frame and to_column_dict

machow:
This PR adds two new DataFrame generic functions:

* subset_frame: return a new DataFrame that is a subset of `rows=` and `cols=`.
* to_column_dict: convert dataframe to a dictionary, where each column is an entry.

These functions were used in prototyping rendering GT as an interactive table in https://github.com/machow/reactable-py/blob/main/reactable/render_gt.py, but I suspect they'll be useful here.

(Alternatively, we could implement them in reactable-py, which also has a `_tbl_data.py` file?!)

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -446,18 +446,6 @@ class Boxhead(_Sequence[ColInfo]):
             return None
         return column[0]
 
-    # Get a list of visible column labels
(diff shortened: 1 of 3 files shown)


machow:
Going to close, since I moved the logic into reactable-py

Issue #403, opened 2024-07-18
Wrap header rows in `<thead>`


cwickham:
## Prework

- [X] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [X] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

Header rows should always be wrapped in `<thead>`. In particular, this will mean tables output to Typst that break over pages get the header rows repeated on each page.

## Reproducible example

Consider the following example:
```python
from great_tables import GT, md, html
from great_tables.data import islands

gt_tbl = GT(islands)
# Show the output table
gt_tbl
```

This results in the following (partial) HTML:

```html
<table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">

<tr class="gt_col_headings">
  <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col" id="name">name</th>
  <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="size">size</th>
</tr>
<tbody class="gt_table_body">
...
```

If this table is used in Quarto with `format: typst` and is allowed to break over pages, the header rows are not repeated on the second page:

<img width="634" alt="Screenshot 2024-07-18 at 11 01 42 AM" src="https://github.com/user-attachments/assets/b8a04639-6c69-4473-951e-fad950036b30">

<details>
  <summary>QMD</summary>

````markdown
---
title: "Untitled"
format: typst
keep-md: true
---

```{=typst}
#show figure: set block(breakable: true)
```
```{python}
from great_tables import GT, md, html
from great_tables.data import islands

gt_tbl = GT(islands)
# Show the output table
gt_tbl
```
````
</details>

## Expected result

E.g. in `gt` the header rows are nested inside `<thead>`, 
```{html}
<table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">
  <thead>
    <tr class="gt_col_headings">
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col" id="name">name</th>
[...]

PR #421, opened 2024-08-26
fix!: ensure that `<thead>` element encloses both title/subtitle and column labels

rich-iannone:
This makes it so that `<thead>` doesn't just enclose the title/subtitle but rather both the header and column labels are included in a `<thead>` element. This is needed for overall compliance with HTML5 (this change eliminates the issue found through checking in https://validator.w3.org/) and though browsers can typically handle non-compliant HTML, Pandoc (thus GT tables rendered in Quarto) cannot.

Fixes: https://github.com/posit-dev/great-tables/issues/403

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -8,12 +8,11 @@ from htmltools import HTML, TagList, css, tags
 
 from ._gt_data import GTData
 from ._tbl_data import _get_cell, cast_frame_to_string, n_rows, replace_null_frame
-from ._text import StringBuilder, _process_text, _process_text_id
+from ._text import _process_text, _process_text_id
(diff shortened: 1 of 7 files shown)


great_tables/_utils_render_html.py
```diff
@@ -39,22 +39,20 @@ def create_heading_component_h(data: GTData) -> StringBuilder:
     )
 
     if has_subtitle:
-        heading = f"""
-  <tr>
-    <th colspan="{n_cols_total}" class="gt_heading gt_title gt_font_normal">{title}</th>
+        heading = f"""<tr class="gt_heading">
```
machow: Can you change this to match the previous formatting? (so there is a line break at the beginning of this string?)

machow (approved):
LGTM

PR #422, opened 2024-08-26
Remove mentions of `accounting` in `force_sign=` argument

rich-iannone:
In several of the formatting methods, the common argument `force_sign=` erroneously mentions accounting notation. That feature does not yet exist in Great Tables so this text is removed here.

Fixes: https://github.com/posit-dev/great-tables/issues/241


machow (approved):
LGTM

Issue #374, opened 2024-06-07
Add `google_font()`


alvinwanyeki:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal

[Helper function for specifying a font from the Google Fonts service.](https://gt.rstudio.com/reference/google_font.html#examples)

PR #423, opened 2024-08-27
feat: add google font helper, implement in `opt_table_font()` and `style.text()`

rich-iannone:
This adds support for incorporating Google fonts into the table text, via:

- `opt_table_font(font=google_font(...))`
- `tab_style(style=style.text(font=google_font(...)`

The necessary additions are:

- the `table_additional_css=` argument in `tab_options()`
- the `google_font()` function
- statements in `compile_scss()` to add the necessary `@import` statement at the top of the `<style>` block in an HTML table render

A simple example of this is through the use of `opt_table_font()`

```python
from great_tables import GT, exibble, google_font

GT(exibble.head()).opt_table_font(font = google_font("IBM Plex Mono"))
```

image

Fixes: https://github.com/posit-dev/great-tables/issues/374


great_tables/_styles.py
```diff
@@ -220,12 +220,23 @@ class CellStyleText(CellStyle):
...
 
     def _to_html_style(self) -> str:
+
         rendered = ""
 
         if self.color:
             rendered += f"color: {self.color};"
         if self.font:
-            rendered += f"font-family: {self.font};"
+            font = self.font
```
machow, line 229: Can you test the lines here that are flagged as untested in codecov? 

app.codecov.io/gh/posit-dev/great-tables/pull/423?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=posit-dev

great_tables/_tab_create_modify.py
```diff
@@ -118,6 +119,34 @@ def tab_style(
...
+                # Replace GoogleFont class with font name
+                s.font = font_name
+
+                # Append the import statement to the `table_additional_css` list
+                existing_additional_css = self._options.table_additional_css.value + [
+                    font_import_stmt
+                ]
+
+                # Add revised CSS list via the `tab_options()` method
+                new_data = new_data.tab_options(table_additional_css=existing_additional_css)
```
machow, line 148: For pairing Thursday, let's go over alternative paths this could go through. Currently, CellStyle as an interface has a `_to_html_style()` method, that gets used on render.

Here, it looks like the new behavior to support is something like this:

* A certain type of style requires setting some initial table-level CSS (once for the importing of fonts)

Currently, the path used for this is:

* in tab_style, detect a particular CellStyleText (e.g. with a google font)
* create import statement for font
* set initial table css for importing font in the additional_css option
* additional css gets included on render

However, this means that if you use tab_style multiple times with google font, etc.., the additional css gets appended multiple times. Moreover, if the person modifies additional_css, they can remove the necessary import.

A second issue is that this exact code is pasted into the `opt_table_font()` method. We likely don't want to put custom handling for specific CellStyle types (with specific attributes) into these methods.

Repeating code twice isn't necessarily an issue though, so it could be that we map out paths for these kinds of activities, merge, and punt the work to a new issue

rich-iannone: Sounds good!

great_tables/_scss.py
```diff
@@ -141,7 +141,23 @@ def compile_scss(
...
+
+    # Determine if there are any additional CSS statements
+    has_additional_css = (
+        additional_css is not None and isinstance(additional_css, list) and len(additional_css) > 0
+    )
+
+    # Ensure that list items in `additional_css` are unique and then combine statements while
+    # separating with `\n`; use an empty string if list is empty or value is None
+    if has_additional_css:
+        additional_css_unique = OrderedSet(additional_css).as_list()
```
machow, line 155: this ensures that if `google_font()` is used multiple times, that it doesn't keep adding css

machow (approved):
LGTM; from pairing, we have an issue open for exposing a general mechanism for google font-like behaviors, and will punt it down the road

Issue #424, opened 2024-08-30
Rendered table has a missing last row with Chrome webdriver


HagaiHargil:
## Prework

- [X] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [X] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description
Hi,

GT has been great so far - thanks! Recently, the tables it outputs have a semi-missing last line:

<img width="389" alt="Screenshot 2024-08-30 150555" src="https://github.com/user-attachments/assets/2336baa3-127a-47b7-947b-274ef6a6b9a3">

(I cut the upper half of the image since the data is sensitive)

This is the last row of the table - and you can also see that the table "knows" that it's the last one since the image ends a bit below that. This happens on Windows and Linux (WSL).

## Reproducible example

The code I'm running is a simple `GT(df)...save(fname)` (see below) and as I said it worked until a week ago. Sadly - I have no idea what has changed. `git diff` doesn't provide any clues, but regardless the code and data are still legit in the sense that they should be rendered correctly.

I was somehow able to make Firefox run on this work laptop (inside Docker) and saw that changing `web_driver` to `firefox` solves the issue, but still not having it work on Chrome (the default) on both Windows and Linux feels worthy of a bug report (and annoys me slightly).

```python
    table = (
        GT(df)
        .tab_header(title=title + " (values in um)", subtitle=subtitle)
        .tab_stub(rowname_col="Radius")
        .tab_stubhead(label="Radius range")
        .cols_label(
            errors_x="X Mean Error",
            std_err_x="X St. Dev.",
            errors_y="Y Mean Error",
            std_err_y="Y St. Dev.",
        )
        .tab_spanner(label="X", columns=["errors_x", "std_err_x", "X 99% Rep."])
        .tab_spanner(label="Y", columns=["errors_y", "std_err_y", "Y 99% Rep."])
        # .fmt_number(["std_err", "mean_err"], decimals=2)  # unicode error, unrelated to this issue
    )
    table.save(str(fname), web_driver="firefox")  # "chrome" fails
```

These are things that I checked:

1. The HTML output (`as_raw_html()`) outputs the correct table.
2. Using either a Polars or Pandas DF doesn't change anything.
[...]

PR #425, opened 2024-08-30
fix: save with latest google chrome

machow:
This PR is to address #424 

The challenge was that our resizing set the browser outer dimensions, but this does not account for space taken up by browser UI elements (which vary by browser). We had a "crud factor" set that basically accounted for this space by blissful accident.

`.save()` now explicitly tries to calculate the space between outer and inner browser dimensions, with a much smaller crud factor (to create some small padding around the table. This is important because table borders without crud factor might cause issues).

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -360,7 +360,7 @@ def _save_screenshot(
     # the window can be bigger than the table, but smaller risks pushing text
     # onto new lines. this pads width and height for a little slack.
     # note that this is mostly to account for body, div padding, and table borders.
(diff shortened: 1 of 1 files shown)


machow:
@rich-iannone do you mind checking the uploaded screenshots when you review? They're tucked intot he "Test for saving table image..." gh action...

image

machow:
When it's time, let's squash and merge this so the orphan commit isn't in our history

rich-iannone (approved):
LGTM!

Issue #395, opened 2024-07-08
Document tab_spanner's unit syntax functionality


machow:
For `cols_label()` we have these notes:

![image](https://github.com/posit-dev/great-tables/assets/2574498/50cc0541-fa7e-464b-9f2d-0525db3fd76c)

We should add a similar explanation to tab spanner

PR #426, opened 2024-08-30
docs: indicate that `tab_spanner()` allows for use of units notation in its `label=` arg

rich-iannone:
This is a simple documentation improvement for the `tab_spanner()` method. The `cols_label()` allows for the use of units notation syntax and has documentation to explain that. This was missing for `tab_spanner()` though it has equivalent functionality. This PR explains that units notation can be used in `tab_spanner()`

Fixes: https://github.com/posit-dev/great-tables/issues/395

### DIFF
--- a/great_tables/_spanners.py
+++ b/great_tables/_spanners.py
@@ -40,6 +40,16 @@ def tab_spanner(
     `level` as space permits) and with `replace`, which allows for full or partial spanner
     replacement.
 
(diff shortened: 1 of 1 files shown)


machow (approved):
LGTM, thanks! (from pairing, we'll add a guide page for better explaining text handling / units https://github.com/posit-dev/great-tables/issues/439)

PR #431, opened 2024-09-05
Standardize the `GTSelf` signature

jrycw:
It seems we have two different signatures, `data: GTSelf` and `self: GTSelf`, to describe `GTSelf`. I propose we standardize the signature to `self: GTSelf`, as it is more commonly used in the repo.


machow (approved):
LGTM -- thanks for this nice clean up 😊

PR #432, opened 2024-09-05
docs: update import statement in the `GT.data_color()` example

jrycw:
I propose changing the import statement in the `GT.data_color()` example to follow a more commonly used pattern.


rich-iannone (approved):
Thank you!

Issue #348, opened 2024-05-15
Update signature for tab_spanners to indicate label can be md(), html() result


machow:
https://github.com/posit-dev/great-tables/pull/345 supported having more than a single spanner. As a part of that, we added some quick logic for labels that are md() and html() result objects.

Let's be sure to...

* update the tab_spanners label annotation to reflect these objects are possible
* update the logic to use isinstance checks rather then hasattr.

(Alternatively we could use structural subtyping and support anything with a specific attribute, but I'd lock it down for now!)

PR #434, opened 2024-09-05
fix: use `isinstance()` instead of `hasattr()`

rich-iannone:
In https://github.com/posit-dev/great-tables/issues/348 we wanted to ensure that the signature of `tab_spanner()` has been updated with types that indicate `md()` and `html()` could be used in the `label=` argument. This was done in a previous PR.

A second requirement was update the logic to use `isinstance()` checks rather than `hasattr()`. That change was made in this PR.

Fixes: https://github.com/posit-dev/great-tables/issues/348

### DIFF
--- a/great_tables/_spanners.py
+++ b/great_tables/_spanners.py
@@ -130,7 +130,7 @@ def tab_spanner(
     if id is None:
         # The label may contain HTML or Markdown, so we need to extract
         # it from the Text object
-        if hasattr(label, "text"):
+        if isinstance(label, Text):
             id = label.text
         else:
             id = label


machow:
Can you show that #348 is resolved?

* If `md()` is passed is it handled as expected? (e.g. does the markdown get rendered)
* If `html()` is passed is it handled as expected? (e.g. does the html get rendered)
* If text is passed that looks like html, it should not get rendered as html (if this is happening, even if we close #348 IMO  we should track this bug)

machow:
I got this so wrong 😭 -- these changes are for handling id, which is the unique name for referring to the spanner. But label display is handled later on in the code (https://github.com/posit-dev/great-tables/pull/434/files#diff-b4c74a432c9ffd70ea3c383684233bfdea5f3bac27dacab684be0c3a4e9cffb5L190).

machow (approved):

PR #444, opened 2024-09-16
Update `FmtImage.to_html()`

jrycw:
Hello team,

I'd like to format the column by providing an HTTP/HTTPS URL to the `path` parameter in `GT.fmt_image()` like so:

```python
import polars as pl
from great_tables import GT

df = pl.DataFrame({"logo": ["GT_logo.svg"]})
GT(df).fmt_image(
    "logo",
    path="https://raw.githubusercontent.com/posit-dev/great-tables/main/docs/assets/",
)
```
I was expecting this outcome:
image

However, I'm encountering a `TypeError: sub() missing 1 required positional argument: 'string'`.

It appears that the error originates from this line:
https://github.com/posit-dev/great-tables/blob/016c67f7cf75272ac4bff0b678383a8f8993c985/great_tables/_formats.py#L3529

If I understand correctly, this line is intended to do two things: (1) remove trailing spaces, if any, and (2) remove the trailing `/`, if present. To fix this, the line could be updated as follows:

```python
norm_path = re.sub(r"/\s*$", "", self.path)
```
This would:
- Pass all three required arguments to `re.sub()`.
- Change the `+` modifier to `*` to allow for zero or more spaces.

That said, I believe using `self.path.rstrip().removesuffix("/")` as proposed in this PR is more straightforward and easier to understand.

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -3492,7 +3492,6 @@ class FmtImage:
     SPAN_TEMPLATE: ClassVar = '<span style="white-space:nowrap;">{}</span>'
 
     def to_html(self, val: Any):
-        import re
         from pathlib import Path
 
         # TODO: are we assuming val is a string? (or coercing?)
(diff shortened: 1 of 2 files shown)


machow:
@rich-iannone do you mind taking a look? Especially on double checking`norm_path`'s job there? Seems like a nice addition!

(may be useful for us to leave a note here on what would happen in the gt r package if someone used an http path)

rich-iannone (approved):
LGTM!

Issue #439, opened 2024-09-12
Create docs for text handling/rendering


rich-iannone:
Documentation is currently a bit limited w.r.t. text input for a wide variety of methods (e.g., `cols_label()`, `tab_header()`, etc.).

We seem to be missing:

- term for text handling (e.g., `md()`, `html()`, inline units)
- term for inline units
- explanations for where these can be used (could start with a simple list of methods and expand as necessary)

We ought to have a section in the Get Started guide that generally explains the value of inline units in a table (e.g., why use inline units at all?).

machow:
One more piece from recent discussions, it might help to have a one sentence explanation os what a "unit" is, and what "units" are (e.g. is everything inside the `{{}}` shorthand a unit? or is it "units" and a unit is a smaller piece)

PR #446, opened 2024-09-17
Add an example demonstrating `define_units()` usage

jrycw:
Hello team,

As we know, `define_units()` powers `GT.fmt_units()` behind the scenes, but it seems to lack a clear example for users to quickly grasp its usage. 

I’d like to propose an example that demonstrates how to use `define_units()` to render a string with unit annotations as a subtitle in the table header. This example highlights that `define_units()` can be used independently and within components that don’t yet support unit annotations (such as the table header). 

Below is the final table for reference.
image

### DIFF
--- a/great_tables/_helpers.py
+++ b/great_tables/_helpers.py
@@ -1051,6 +1051,71 @@ def define_units(units_notation: str) -> UnitDefinitionList:
         )
     )
     ```
+
+    Examples
+    --------
+
+    Let’s demonstrate a use case where we utilize `define_units()` to render an equation as
+    the subtitle in the table header, which currently doesn’t accept unit notation as input.
+
+    We'll start by creating a Polars DataFrame representing the calculations of the equation
+    $y= a_2x^2 + a_1x + a_0$.
+
+    ```{python}
+    #| code-fold: true
+
+    import polars as pl
+    from great_tables import GT, html, define_units
+
+    df = pl.DataFrame(
+        {"x": [1, 2, 3], "a2": [2, 3, 4], "a1": [3, 4, 5], "a0": [4, 5, 6]}
+    ).with_columns(
+        y=(
+            pl.col("a2").mul(pl.col("x").pow(2))
+            + pl.col("a1").mul(pl.col("x"))
+            + pl.col("a0")
+        )
+    )
+
+    df
+    ```
+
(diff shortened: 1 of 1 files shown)


rich-iannone (approved):
LGTM!

rich-iannone:
@machow this looks great to me. Please merge whenever convenient!

machow:
@rich-iannone afaict parsing the curly braces in inputs like `"{{a_1}}"` is not a job for `define_units()` (for example, we didn't list it out in [the rules](https://docs-preview-define-units--gt-python.netlify.app/reference/define_units#specification-of-units-notation).

It seems like define_units() handling of curly braces may be a bug (and inconsistent)?:

```python
from great_tables import define_units

# removes curly braces ----
define_units("{{a_1}}").to_html()
# > 'a<span style="white-space:nowrap;"><sub style="line-height:0;">1</sub></span>'

# Notice that the second piece in curly braces (b_2) still has the braces around it ----
define_units("{{a_1}} {{b_2}}").to_html()
# > 'a<span style="white-space:nowrap;"><sub style="line-height:0;">1}}</sub></span> {{b<span style="white-space:nowrap;"><sub style="line-height:0;">2</sub></span>'
```

What is the intended behavior of define_units? It seems like it should not parse braces, since that's afaict `UnitStr.from_str()` territory. So...

* `UnitStr.from_str()` parses pieces out of braces,
* then it passes those piece to `define_units()`

Does this sounds right, and this `define_units()` behavior is a bug?

jrycw:
Thanks for pointing this out! I originally thought it was a cool feature and didn’t notice that.

machow:
I definitely think there is a cool feature tucked away in `.cols_label()`! You can use `{{}}` as a shorthand for using `define_units()`. For example, `cols_label(some_col="Some Col {{h[_0^3]}}")`

But I wonder if we skipped an important piece when wiring up units. Based on this PR, it seems like exposing a function similar to `html()` and `md()` might help connect the dots?

Maybe something like this?:

* explicit: `cols_label(some_col = text("Some Col ", unit("h[_0^3]")))`
* shorthand: `cols_label(some_col = "Some Col {{h[_0^3]}}")`

This would provide a bridge into other structure things:

* `tab_header(title = text("Some title", unit("a_1"))`

Could be nicer ways to handle though?! (Related to #439)

PR #447, opened 2024-09-17
Favor `OrderedSet().as_list()`

jrycw:
Related to #407.

It appears that both `list(OrderedSet())` and unpacking `OrderedSet()` rely on the `__iter__()` method. I’d like to suggest using `OrderedSet().as_list()` instead of depending on the dunder method, as this approach is clearer and more stable, even though the implementation is unlikely to change.


machow (approved):
Seems reasonable, thanks for such a steady stream of improvements 😄

machow:
@rich-iannone do you want to merge?

PR #450, opened 2024-09-17
Refactor import statements in `_formats.py`

jrycw:
This PR focuses on cleaning up the import statements in `_formats.py`:

* Moved `import re` and `from dataclasses import dataclass` to the top, as they were previously placed near the end of the file.
* Removed the redundant local import of `import re`, now handled at the top level.
* Eliminated the local import of `from pathlib import Path` within a function, as `Path` is already imported at the top level and is required for type hints in other functions.

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import math
+import re
+from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


machow (approved):
LGTM! thanks for tidying this up

rich-iannone (approved):
Thanks!

Issue #418, opened 2024-08-17
Allow images to be used as column headers


minimav:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal

This can be achieved in a couple of ways already, although they both are very hacky. One is by using placeholder column names and replacing them appropriately in the HTML output created by `as_raw_html` with base64 encoded `<img>` tags (essentially the approach that is used in `fmt_image`).

A couple of approaches that might work for a proper implementation this:

* A `.fmt_column_header_image` method that mimicks `fmt_image` and reuses `FmtImage` if possible.

```python
df = pl.DataFrame([
    {"column_name": 1.0}
])

GT(df).fmt_column_header_image(
     column_name=<path to image>,
     # fmt_image kwargs
     height=...
)
```

* Allow option in `.col_labels` to enable use of an image instead of a renaming label, or infer from argument type

```python
df = pl.DataFrame([
    {"regular_name": 1.0, "image_column": "something"}
])

GT(df).col_labels(
    regular_name="some other name",
    # infer from Path type?
    image_column=Path(...),
)
```

I would be happy to try implementing something like this. Thank you 😄

jrycw:
Thank you for bringing this issue to our attention. I'll close it now, as the [vals.fmt_image()](https://posit-dev.github.io/great-tables/reference/vals.fmt_image.html#great_tables.vals.fmt_image) function is available to achieve the desired result.

PR #451, opened 2024-09-18
Add `val_fmt_image()` to enable image rendering in various components

jrycw:
Related to #418.

It seems that users are interested in rendering images across different components. To support this, I'd like to propose introducing a `val_fmt_image()` function. With this function, we can achieve more flexibility, such as:
```python
import polars as pl
from polars import selectors as cs
from great_tables import GT, vals, html
from importlib_resources import files


img_paths = files("great_tables") / "data/metro_images"

df = pl.DataFrame(
    {
        "col1": list("123"),
        "col2": list("123"),
        "col3": list("123"),
        "group": ["grp_a", "grp_a", "grp_b"],
        "row": ["row_1", "row_2", "row_3"],
    }
)

img1, img2, img3, img4, img5, img6 = vals.fmt_image(
    list("123456"), path=img_paths, file_pattern="metro_{}.svg"
)

(
    GT(df)
    .cols_label(**{"col1": html(img1), "col2": html(img2), "col3": html(img3)})
    .tab_stub(groupname_col="group", rowname_col="row")
    .tab_stubhead(html(img4))
    .tab_header(title=html(img5))
    .tab_source_note(html(img6))
    .fmt_image(cs.starts_with("col"), path=img_paths, file_pattern="metro_{}.svg")
)
```
image

The idea is to wrap the output from `val_fmt_image()` in the `html()` function.

### DIFF
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
+from pathlib import Path
 
 from great_tables import GT
 from great_tables.gt import GT, _get_column_of_values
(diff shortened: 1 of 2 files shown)


rich-iannone:
This will be a great addition, thank you! If you can, could you add some basic tests for this function in a new `test_formats_vals.py` file? I know we aren't directly testing the other `val_fmt_*` functions right now but this will get the process started.

jrycw:
Sure. Since we have actual SVG files in the project, I thought it might be a good idea to directly test `val_fmt_image()` using those images. Feel free to update the tests if they need further refinement.

Yasin197:
This would be an amazing feature, when do you think it will be released?

rich-iannone (approved):
LGTM!

rich-iannone:
Sorry for taking awhile to get back to this PR. Just reviewed again, did some testing, and everything is really good here!

PR #452, opened 2024-09-18
Unify `GT.cols_label()` and `GT.cols_width()` to accept both `cases` and `**kwargs`

jrycw:
This PR unifies `GT.cols_label()` and `GT.cols_width()` to accept both `cases` and `**kwargs`. 

Currently, these two functions have different parameter structures, which can be confusing and often requires consulting the documentation. I hope this change simplifies their usage :)

### DIFF
--- a/great_tables/_boxhead.py
+++ b/great_tables/_boxhead.py
@@ -11,7 +11,9 @@ if TYPE_CHECKING:
     from ._types import GTSelf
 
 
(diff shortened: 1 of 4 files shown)


rich-iannone (approved):
LGTM

machow:
Thanks, this is great! I wonder if in the future, there might be some good uses of polars expressions inside cases, e.g.

`cols_width({cs.starts_with("abc_"): "50px"})`

Although maybe there's a more polars way to go from selector -> literal? 🤔

PR #453, opened 2024-09-18
docs: add a preview section for the built-in datasets

jrycw:
I’d like to propose adding a preview section to allow users to quickly glance at the built-in datasets. I believe this would help users choose the appropriate dataset for exploring the table.

### DIFF
--- a/great_tables/data/__init__.py
+++ b/great_tables/data/__init__.py
@@ -311,12 +311,13 @@ This is a dataset with 13,545 rows and 5 columns.
 Preview
 -------
 ```
-  country_name country_code_2 country_code_3  year  population
-0        Aruba             AW            ABW  1960       54608
-1        Aruba             AW            ABW  1961       55811
-2        Aruba             AW            ABW  1962       56682
-3        Aruba             AW            ABW  1963       57475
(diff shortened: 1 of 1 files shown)


rich-iannone (approved):
LGTM!

machow:
I wonder if something like [polars.DataFrame.glimpse](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.glimpse.html) could be useful here? I realize these are pandas DataFrames, but glimpse scales so nicely width number of columns 😭

jrycw:
Thanks for sharing this—it’s new to me. 

I like how it uses the row to display information more concisely. I’m fine with either option :).

jrycw:
I've pushed another commit using `polars.DataFrame.glimpse`, and it looks promising—I’m leaning toward this option. 

My only concern is that the preview includes Polars-specific data types, which might be a small hurdle for users more familiar with Pandas.

jrycw:
@rich-iannone and @machow, have we reached a consensus on which format to use for the preview section yet?

machow:
From pairing w/ Rich -- the glimpse approach looks great, thanks for taking the time to explore different approaches here!

Issue #440, opened 2024-09-12
Encoding issue on windows when saving a table


bart-r-willems:
## Description

When saving a table with the `save` message, an exception occurs in Windows.

## Reproducible example

- Run pandas/jupyter on a windows maching
- create a simple table
- try saving e.g. `df.pipe(GT).save('test.pdf')`

**The bottom of the stack trace shows this**

    File D:\Python\Python312\Lib\site-packages\great_tables\_export.py:310, in save(self, file, selector, scale, expand, web_driver, window_size, debug_port, _debug_dump)
        303 with (
        304     tempfile.TemporaryDirectory() as tmp_dir,
        305     wdriver(options=wd_options) as headless_browser,
        306 ):
        307 
        308     # Write the HTML content to the temp file
        309     with open(f"{tmp_dir}/table.html", "w") as temp_file:
    --> 310         temp_file.write(html_content)
        312     # Open the HTML file in the headless browser
        313     headless_browser.set_window_size(window_size[0], window_size[1])

    File D:\Python\Python312\Lib\encodings\cp1252.py:19, in IncrementalEncoder.encode(self, input, final)
         18 def encode(self, input, final=False):
    ---> 19     return codecs.charmap_encode(input,self.errors,encoding_table)[0]

    UnicodeEncodeError: 'charmap' codec can't encode character '\u2212' in position 7916: character maps to <undefined>

Note that the error occurs due to CP1252 encoding which is the default. I fixed it by adding `encoding='utf-8'` ib line 309 of the _export.py file:

        # Write the HTML content to the temp file
        with open(f"{tmp_dir}/table.html", "w", encoding='utf-8') as temp_file:
            temp_file.write(html_content)

(Sorry I'm not comfortable with pull requests I don't want to ruin anything)

## Expected result

A file `test.pdf` should be created

## Development environment

- Operating System:  Windows 10, Python 3.12.4
- **great_tables** Version: 0.11.0

HansMellman:
Encountering the same issue when trying to utilize a Unicode character in the heading section (in lieu of an image). Unable to figure out a workaround yet without just removing the character.

bart-r-willems:
I just tested it, with my fix in place that works without issues (windows
10, python 3.12.4)

HansMellman:
You sir, are wonderful. Thank you for that - did the trick perfectly! Hopefully they'll be able to implement the fix natively.

jrycw:
[...]

Issue #391, opened 2024-07-03
table with negative data fails to save as image when using .fmt_number


Mike-Purtell:
## Description

Saving image of a table to png file fails when the table has negative values, and .fmt_number is used.

## Reproducible example - Verified on complex use cases, and the simple example posted here. Notice that the file extension is .txt, please change to .py or paste into a notebook to run this code.

[gt_bug_2024_07_03_MP.txt](https://github.com/user-attachments/files/16091016/gt_bug_2024_07_03_MP.txt)

## Development environment
Win11, great_tables 0.9.0, python 3.11.5 with Anaconda/Jupyter Lab, polars 0.20.31

## Expected result
Expect that table with negative data can use .fmt_number to clean the table, and then can be saved as an image file. This failed. .

machow:
Hey, thanks for raising--I'm having some trouble viewing the .txt file. Do you mind pasting in the python code directly?

Mike-Purtell:
Hello Michael,

No problem for me to post the python code, i will do so in a few minutes.

Thank you for looking into this issue.

mike purtell

Mike-Purtell:
**Here is the python code I wrote to demonstrate the reported bug:**


import great_tables
from great_tables import  GT
import polars as pl
def save_gt(df, filename):
    my_gt = (GT(df).tab_header(title = f'{filename}', subtitle = f'subtitle')
        
        # TO TEST THIS BUG, RUN THIS CODE WITH and WITHOUT .fmt_number 
        # save table to image fails when .fmt_number with negative values is used
        .fmt_number(
            columns=df.columns,
            decimals=1,
            use_seps=True, 
            sep_mark=','
            )
    )
    try:
        my_gt.save(filename, window_size=(6, 6))
        print(f'\n ###########  SUCCESSFULLY WROTE {filename}  ###########\n')
    except:
        print(f'\n ###########  FAILED TO WRITE {filename}  ###########\n')
    return
        
df_pos = pl.DataFrame(
    {
        'A': [x for x in list(range(3))],
        'B': [x*0.5 for x in list(range(3))],
        'C': [x*01.5 for x in list(range(3))],
    }
)
# make df_neg by multiplying all values of df_pos by -1
df_neg = df_pos.with_columns(pl.all()*pl.lit(-1))
display(df_neg, df_pos)
save_gt(df_neg, 'df_neg.png')
save_gt(df_pos, 'df_pos.png')

jrycw:
[...]

PR #454, opened 2024-09-18
Update `GT.save()`

jrycw:
Related issues: #391, #440

This PR aims to refine `GT.save()` in three areas:

1. Use `Path().suffix` to check the file extension.
2. Unpack `window_size` directly instead of using indexing.
3. Add an `encoding` parameter, defaulting to `utf-8` for writing temporary files.

I'm not entirely sure if `utf-8` is the ideal fix for all users, but I believe it's the best default for most. Let's start with `utf-8` and adjust over time if needed.

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -185,6 +185,7 @@ def save(
     web_driver: WebDrivers = "chrome",
     window_size: tuple[int, int] = (6000, 6000),
     debug_port: None | int = None,
+    encoding: str = "utf-8",
     _debug_dump: DebugDumpOptions | None = None,
 ) -> None:
     """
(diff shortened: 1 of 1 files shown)


jrycw:
I'm curious why codecov bot isn't running for this PR?

rich-iannone:
> I'm curious why codecov bot isn't running for this PR?

I looked into our project configuration in the codecov site and saw nothing that requires attention. This might be a rare miss from codecov (maybe due to services being down). I'll see if this can be reported to them however.

machow (approved):
Thanks, this LGTM.

rich-iannone (approved):
LGTM!

Issue #414, opened 2024-08-06
Problem with the encoding


Galoist:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

Cyrillic (arabic too, I assume, other alphabets as well) isn't shown properly.
<img width="193" alt="image" src="https://github.com/user-attachments/assets/f0ed26ae-c89e-4a6b-b903-9aa0fa6ed1f4">

## Reproducible example

- [x] Post a minimal reproducible example (MRE) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
```python
import pandas as pd
from great_tables import GT

GT(pd.DataFrame([{"a": "привіт"}, {"a": "світ"}])).show()
```
  - [x] **Runnable**: post enough code and data so any onlooker can create the error on their own computer.
  - [x] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
  - [x] **Readable**: format your code according to the [Style Guide for Python Code](https://peps.python.org/pep-0008/).

## Expected result

The text in other languages should be shown properly

## Development environment

- Operating System: [e.g., Linux, macOS, Windows] Mac os
- **great_tables** Version: [e.g., 0.0.1]
great_tables = ">=0.10.0,<0.11"
pandas = ">=2.2.2,<3"
python = ">=3.12.4,<4"

rich-iannone:
Thanks for reporting this and thanks for the reproducible example! We'll look into ensuring this works with all scripts.

rich-iannone:
With the merged PR of https://github.com/posit-dev/great-tables/pull/458, this is now fixed for the next release of Great Tables.

PR #458, opened 2024-09-23
fix: use full html page in show() for correct utf-8 display

machow:
This PR addresses #414 by producing a full html page when using `.show()` to open a browser window. Previously, we only produced the table fragment, which for some reason caused the browser (or server) to screw up the encoding.

(The full html includes `<meta charset="utf-8"/>` which makes the encoding explicit)

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -97,22 +97,23 @@ def show(
 
     """
 
-    html = self._repr_html_()
-
     if target == "auto":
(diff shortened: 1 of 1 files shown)


machow:
shoot -- I need to fix a bit of code in our old blog post :/

rich-iannone (approved):
LGTM!

rich-iannone (approved):
Looks good, thanks!

Issue #435, opened 2024-09-11
gt.opt_row_striping unexpected keyword argument row_striping_include_table_body


aronatkins:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

Attempting to disable row striping presents an error.

## Reproducible example

```python
from great_tables import GT
import polars

data = polars.DataFrame({
    "first": ["alice", "bob"],
    "last": ["smith", "smith"],
})
table = GT(data)
table.opt_row_striping(False)
```

```
Traceback (most recent call last):
  File "PROJECT_DIR/bug.py", line 9, in <module>
    table.opt_row_striping(False)
  File "PROJECT_DIR/env/lib/python3.11/site-packages/great_tables/_options.py", line 619, in opt_row_striping
    return tab_options(self, row_striping_include_table_body=row_striping)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: tab_options() got an unexpected keyword argument 'row_striping_include_table_body'
```

## Expected result

What should have happened? Please be as specific as possible.

Row striping is disabled without error.

## Development environment

- Operating System: macOS
- **great_tables** Version: 0.11.0

## Additional context

Directly calling `gt.tab_options(row_striping_include_table_body=False)` presented the same error.

jrycw:
It seems that this option is not officially released (it is hidden from `GT.tab_options()`).

aronatkins:
@jrycw - The option is present in the published documentation, which is how I found it.
https://posit-dev.github.io/great-tables/reference/GT.html#great_tables.GT.opt_row_striping

aronatkins:
It's not important for this issue, but I was trying to build tables in both R and Python. The R package also has a public function: https://gt.rstudio.com/reference/opt_row_striping.html

rich-iannone:
Sorry you bumped into this problem. I will work on getting row striping generally working in Great Tables for the upcoming release.

rich-iannone:
[...]

PR #461, opened 2024-09-24
feat: implement row striping options

rich-iannone:
This PR addresses https://github.com/posit-dev/great-tables/issues/435 by making row striping options functional in rendered tables. An example of using row striping involves three new options in `tab_options()`:

```python

(
    GT(exibble[["num", "char", "row", "group"]], rowname_col="row", groupname_col="group")
    .tab_options(
        row_striping_include_table_body=True,
        row_striping_include_stub=True,
        row_striping_background_color="lightblue"
    )
)
```

image

Alternatively, the `opt_row_striping()` method allows for a simplified shortcut for `row_striping_include_table_body=True` so the user doesn't have to look through `tab_options()` for this basic functionality:

```python

(
    GT(exibble[["num", "char", "row", "group"]], rowname_col="row", groupname_col="group")
    .opt_row_striping()
)
```

image

Note that the default background color for striping is a *very* light gray (by design).

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -516,30 +516,28 @@ def create_body_component_h(data: GTData) -> str:
 
             if is_stub_cell:
 
-                th_classes = ["gt_row", "gt_left", "gt_stub"]
+                el_name = "th"
+
+                classes = ["gt_row", "gt_left", "gt_stub"]
 
                 if table_stub_striped and odd_i_row:
-                    th_classes.append("gt_striped")
-
-                # Ensure that th_classes becomes a space-separated string
-                th_classes_str = " ".join(th_classes)
-
-                body_cells.append(
-                    f"""    <th {cell_styles}class="{th_classes_str}">{cell_str}</th>"""
-                )
+                    classes.append("gt_striped")
 
             else:
 
-                td_classes = ["gt_row", f"gt_{cell_alignment}"]
+                el_name = "td"
+
+                classes = ["gt_row", f"gt_{cell_alignment}"]
 
                 if table_body_striped and odd_i_row:
-                    td_classes.append("gt_striped")
+                    classes.append("gt_striped")
 
-                td_classes_str = " ".join(td_classes)
+            # Ensure that `classes` becomes a space-separated string
+            classes = " ".join(classes)
 
-                body_cells.append(
-                    f"""    <td {cell_styles}class="{td_classes_str}">{cell_str}</td>"""
-                )
+            body_cells.append(
+                f"""    <{el_name} {cell_styles}class="{classes}">{cell_str}</{el_name}>"""
+            )
 
         prev_group_label = group_label
         body_rows.append("  <tr>\n" + "\n".join(body_cells) + "\n  </tr>")
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -444,6 +444,12 @@ def create_body_component_h(data: GTData) -> str:
     if stub_var is not None:
         column_vars = [stub_var] + column_vars
 
+    # Is the stub to be striped?
+    table_stub_striped = data._options.row_striping_include_stub.value
+
+    # Are the rows in the table body to be striped?
+    table_body_striped = data._options.row_striping_include_table_body.value
+
     body_rows: list[str] = []
 
     # iterate over rows (ordered by groupings)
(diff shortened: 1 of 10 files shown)


tests/__snapshots__/test_options.ambr
```diff
@@ -701,3 +701,163 @@
   
   '''
 # ---
+# name: test_tab_options_striping_snap
```
machow, line 704: This is too big. Can you reduce it to a snapshot that is mostly limited to the piece being tested, or test it another way?

tests/test_options.py
```diff
@@ -387,6 +387,36 @@ def test_opt_table_font_raises():
...
+    assert gt_tbl_1._options.row_striping_include_table_body.value == True
+    assert gt_tbl_2._options.row_striping_include_table_body.value == False
+
+
+def test_tab_options_striping():
+
+    gt_tbl_opt = GT(exibble, id="test").opt_row_striping()
+    gt_tbl_tab_options = GT(exibble, id="test").tab_options(row_striping_include_table_body=True)
+
+    assert gt_tbl_opt.as_raw_html() == gt_tbl_tab_options.as_raw_html()
```
machow, line 406: Can you test this further upstream? If rendered html strings are the very last step, is there an earlier step that indicates these two approaches do the same thing (e.g. checking the options values directly, etc..)?

tests/test_options.py
```diff
@@ -387,6 +387,36 @@ def test_opt_table_font_raises():
...
+
+
+def test_tab_options_striping_snap(snapshot):
```
machow, line 409: See comment on snapshot. It'd be okay to test the css and html separately if it makes it easier to strip down. I wonder if you could easily test the css using an assertion, and snapshot test just the html body (with fewer rows of data; the minimum needed to test--3? or w/e makes sense given there are groupings)

machow (changes requested):

tests/test_options.py
rich-iannone: I'll check the options values directly here (comparing the two approaches).

tests/test_options.py
rich-iannone: I'm going to make the input df with much smaller dimensions and only snapshot the table body (everything inside `<tbody>`). I don't think there's a need to snapshot the CSS part.

great_tables/_utils_render_html.py
```diff
@@ -503,10 +515,30 @@ def create_body_component_h(data: GTData) -> str:
                 cell_styles = ""
 
             if is_stub_cell:
-                body_cells.append(f"""    <th class="gt_row gt_left gt_stub">{cell_str}</th>""")
+
```
machow, line 518: This is looking good, so just a small nit here 😅. Do you mind consolidating so there is a bit less duplicated code? Seems like the following could happen:

* change variables like `td_classes` and `th_classes` to have the same name. Similar with `th_classes_str`, etc..
* add a variable like el_name that is either "th" or "td"
* move all other logic (e.g. the appending of "gt_striped", the " ".join, the body_cells.append) out of the if / else, so it is only written once (e.g. move it to line 543 and unindent)
  - Use el_name in what should now be a single piece of code creating the final th or td string.

It looks like if something is a stub, that only affects the element name and some of the classes to be set. (But other logic is being included)

tests/test_options.py
```diff
@@ -387,6 +388,45 @@ def test_opt_table_font_raises():
...
+    assert gt_tbl_opt_stri._options.row_striping_include_stub.value == False
+
+
+def test_tab_options_striping_snap(snapshot):
+
+    gt_tbl = GT(
+        exibble[["row", "group", "char"]].head(4), rowname_col="row", groupname_col="group"
+    ).tab_options(
+        row_striping_include_table_body=True,
+        row_striping_include_stub=True,
```
machow, line 420: Can you test these separately, so that we know if one options is specified, it does not set both things?

machow (changes requested):

great_tables/_utils_render_html.py
rich-iannone: Thanks, will make these changes!

tests/test_options.py
rich-iannone: Good idea. There is now two distinct snapshots where one tests striping only of the stub, the other only of the body cells.

machow (approved):
Yo, thanks for all these tweaks, it's beautiful!

rich-iannone:
Thanks @machow! Always a pleasure.

PR #460, opened 2024-09-24
docs: update `superbowl` example to align with the new version of `Polars`

jrycw:
I've noticed the CI is failing due to the `superbowl` example. This PR updates the example to be compatible with the latest version of `Polars`.

PR #462, opened 2024-09-25
docs: fix deprecation warning in the `superbowl` example

jrycw:
@machow , it seems that the [deprecation warning](https://github.com/posit-dev/great-tables/pull/460#issuecomment-2371277167) is coming from the first code block in the `index.qmd`.


machow (approved):
LGTM, thanks for being so good with all the polars related stuff :)

PR #464, opened 2024-09-27
Include `google_font()` helper fn in API reference

rich-iannone:
Currently the `google_font()` helper function is missing from the package API reference in the site. This PR simply adds it in so it's visible (in time for the next release, `v0.12.0`).


machow (approved):
LGTM

PR #470, opened 2024-09-28
Update docs and code to support `GoogleFont` in `opt_table_font()` and add tests

jrycw:
This PR updates the documentation to mention that the `font` parameter of `opt_table_font()` now accepts a `GoogleFont` object. It also consolidates the internal code and introduces additional tests.


rich-iannone (approved):
Looks great, thank you!

PR #471, opened 2024-09-28
Update `__repr__()` for `GoogleFont`, `UnitStr` and `UnitDefinitionList`

jrycw:
I’d like to propose using `type(self).__name__` to represent the class name, rather than hardcoding the class name directly.


machow (approved):
Agreed, thanks for cleaning this up

PR #473, opened 2024-09-29
fix: resolve global `locale` not being respected in `GT.fmt_*()` functions

jrycw:
This PR addresses a potential bug where the global `locale` setting is not respected in `GT.fmt_*()` functions that accept `locale` as a parameter.

Consider the following example:
```python

(
    GT(exibble[["currency"]])
    .with_locale("ja")
    .fmt_currency("currency")
)
```
image

It appears that the global `locale="ja"` setting is applied, but the output doesn't reflect this, and here’s what I expected:

image

The issue seems to stem from most of the formatting functions not fetching the global `locale` correctly. The line `locale = x._locale._locale if locale is None else locale` in `_resolve_locale()` is crucial to resolving this.

Therefore, I propose applying `_resolve_locale()` at the start of these formatting functions that accept the `locale` parameter. Additionally, I suggest moving the `_validate_locale()` call before `_normalize_locale()` within `_resolve_locale()` for better logic flow.

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -279,11 +279,6 @@ def fmt_number(
     Take a look at the functional version of this method:
     [`val_fmt_number()`](`great_tables._formats_vals.val_fmt_number`).
     """
-
-    # Stop if `locale` does not have a valid value; normalize locale and resolve one
-    # that might be set globally
-    _validate_locale(locale=locale)
-    locale = _normalize_locale(locale=locale)
     locale = _resolve_locale(self, locale=locale)
 
     # Use locale-based marks if a locale ID is provided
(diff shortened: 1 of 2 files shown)


jrycw:
If `locale` only applies to `GT.fmt_*()` functions, the current fix might be sufficient. However, if we eventually need to apply `locale` to other functions, this type of validation could become tedious.

If that becomes the case, I would propose using a decorator for validation, for example:
```python
from functools import wraps


class GTData:
    def __init__(self, locale: str | None = None):
        self._locale = locale


gtdata = GTData()


def _resolve_locale(x: GTData, locale: str | None = None) -> str | None:
    locale = x._locale if locale is None else locale
    # add validators or modifiers afterwards. For example:
    locale = locale.replace("_", "^_^")

    return locale


def localer(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        if locale := kwargs.get("locale"):
            kwargs |= {"locale": _resolve_locale(args[0], locale=locale)}
        return func(*args, **kwargs)

    return wrapper


@localer
def fake_fmt_currency(
    self: GTData, fake1="fake1", *, locale: str | None = None
) -> GTData:
    print(f"{locale=}")


print(fake_fmt_currency(gtdata))  # locale=None
print(fake_fmt_currency(gtdata, locale="new_locale"))  # locale='new^_^locale'
```

In the example code, I assume the `locale` parameter is passed as a keyword-only argument. Otherwise, we may need to use the `inspect` module to create a more generalized validator.

jrycw:
The tests were a bit tricky to write, but hopefully I got them right! :)

rich-iannone (approved):
LGTM!

Issue #467, opened 2024-09-27
Document individual loc.*() functions, include in API reference


machow:
There are 11 loc classes now (via #341). Let's make sure they're all documented with at least 1 example in their docstring, and added to the API reference.

rich-iannone:
With https://github.com/posit-dev/great-tables/pull/474, this is complete.

PR #341, opened 2024-05-14
Allow for granular section restyling via  convenience api

timkpaine:
# Summary

This is a work-in-progress PR for allowing more granular styling of the various components via the API currently used for cell styling (e.g. `loc.body()`). The motivation is that it's often desired to style e.g. all column labels at once. It looks like this was already started in part (there are many TODOs such as [these](https://github.com/posit-dev/great-tables/blob/7eae1f269a7d1f787de7a2c9198d775933794fd5/great_tables/_utils_render_html.py#L103)). 

This PR is nowhere near done yet, so please don't approve CI running (it will just make noise).

Early Example:
```python
from great_tables import GT, style, loc
from superstore import superstore

(
    GT(superstore(10))
    .tab_style(style=style.text(color="red"), locations=loc.column_labels())
)
```

Screenshot 2024-05-14 at 13 42 14

# Checklist

- [x] I understand and agree to the [Code of Conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/).
- [x] I have followed the [Style Guide for Python Code](https://peps.python.org/pep-0008/) as best as possible for the submitted code.
- [ ] I have added **pytest** unit tests for any new functionality.

PR #474, opened 2024-09-30
docs: Add docstrings for new location methods

rich-iannone:
Given that there are now 11 loc classes (new ones added through https://github.com/posit-dev/great-tables/pull/341), we need to ensure that there is sufficient documentation for these (addressing https://github.com/posit-dev/great-tables/issues/467). We need at least one example in each of their docstrings. And we have to ensure these are all added to the API reference.


machow:
Yooo this is looking so good!!!!

This is a small thought -- I wonder if changing...

* from "A location specification for targeting the..."
* to "Target the.."

might help keep the index a bit punchier. The downside might be the word "location" doesn't get used, but it's in the section description, so may be okay 😬?!

image

rich-iannone:
^^^ great idea and I made that change!

PR #475, opened 2024-09-30
docs: do not document GT members inline on its reference page

machow:
This PR removes the redundant method documentation on the GT class page.


rich-iannone (approved):
LGTM, thanks!

Issue #476, opened 2024-10-01
docs: rendering issue for `GT.fmt_*()` functions


jrycw:
It seems we’re encountering a documentation rendering issue with `GT.fmt_*()` functions. For example, see the issue with [GT.fmt_integer()](https://posit-dev.github.io/great-tables/reference/GT.fmt_integer.html#adapting-output-to-a-specific-locale-.doc-section-.doc-section-adapting-output-to-a-specific-locale).

![image](https://github.com/user-attachments/assets/88eca858-ddd6-4486-9ace-6d4beeaeec6c)

rich-iannone:
@machow I can reproduce this in a local docs site build using Quarto `1.5.57` (latest prerelease, same as in CI) *and* after updating `quartodoc` with

```
pip install git+https://github.com/machow/quartodoc.git@a98bf47004a76578e6893b4ba6d9447322137a3c
```

as was recently done for CI (in https://github.com/posit-dev/great-tables/commit/871ca1ca749a1e6c0d97f20907f8be73b07d136c).

*Before* updating `quartodoc` on my system, the local build of the docs showed a correctly rendered heading on the `great-tables/docs/_site/reference/GT.fmt_integer.html` page.

rich-iannone:
FWIW I built the site locally again with Quarto `1.6.18` (latest prerelease) with the `a98bf`-commit build of `quartodoc` and the issue persists.

machow:
Shoot, we're now using the latest quartodoc commit, which tries to add html class attributes to titles. However, I think something about this custom section header in the docstring is messing with that :/. The classes are correctly being added to sections like "Examples".

I'm building the docs locally and will take a look!

machow:
shoot, quartodoc does not expect titles to have special characters (like the backtick) in them, so the markdown to add classes to the title has backticks included in it :/

```
## Returns {.doc-section .doc-section-returns}

...

## Adapting Output To A Specific `Locale` {.doc-section .doc-section-Adapting-output-to-a-specific-`locale`}

NOTE THE `locale` above ----^
...

## Examples {.doc-section .doc-section-examples}

```

jrycw:
Thanks to @machow for the quick fix! I’ll go ahead and close the issue.

jrycw:
@machow , I’m not sure if this is related to the recent fix, but it looks like some descriptions in the docs are now showing rendering issues with backslashes appearing before double quotes (`"`). For example:

[...]

PR #477, opened 2024-10-01
docs: fix in headers causing their css classes get printed out

machow:
Addresses

* #476


machow:
Note that an additional effect of the quartodoc fix is that the offending section is also no long rendered in title case!

https://pr-477--gt-python.netlify.app/reference/gt.fmt_integer#adapting-output-to-a-specific-locale

image

rich-iannone (approved):
Thanks!

Issue #416, opened 2024-08-11
Allow custom web driver (options) when saving


minimav:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal

When saving a table currently the options for the the web driver that selenium uses are strings found [here](https://github.com/posit-dev/great-tables/blob/7365aaa54bfe5616fc1d772d1c39b101e7c3bb77/great_tables/_export.py#L169), for which some default options are set within the `save` [function](https://github.com/posit-dev/great-tables/blob/7365aaa54bfe5616fc1d772d1c39b101e7c3bb77/great_tables/_export.py#L269-L300). In some situations it is helpful to allow further customisation of the web driver e.g. I was trying to save a table from within a Docker container where even with a browser installed in the Docker image correctly (Chrome in my case), I needed to add extra web driver options in order to get saving to work.

There are a couple of ways in which you could add such functionality:

* Allow the user to pass an instantiated web driver via the `web_driver` kwarg alongside the current string options which result in the default options being used, i.e. pass in an object like `wdriver(options=wd_options)` that is created [here](https://github.com/posit-dev/great-tables/blob/7365aaa54bfe5616fc1d772d1c39b101e7c3bb77/great_tables/_export.py#L305C9-L305C36) in the `save` function
* Add a separate kwarg for additional web driver options that would get added to the web driver options like is already done for the default options `wd_options.add_argument(...)`

There will likely be other ways of achieving the same outcome too.

I'd be happy to work on a PR for this if you think this is a good addition. Cheers :)

thuiop:
Seconding this need. I use Firefox installed with snap (which is awful but this is my Ubuntu work PC and it came like that), and it requires some specific setup for Selenium, which cannot be achieved currently.

machow:
Extending the `web_driver` argument makes sense to me -- I can work on a PR to implement

thuiop:
Great, thanks!

PR #478, opened 2024-10-02
feat: allow passing a webdriver instance to save

machow:
addresses #416 by allowing a webdriver instance to be passed to `save(web_driver=...)`. This also addresses #479 by coercing file=Path(...) to a string.

Note that:

* Because we do context handling when we create web drivers, I wrapped any passed web drivers in a no-op context handler.
* Some arguments (like debug_port) that require setting options do not apply when a webdriver instance is passed.

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -197,7 +197,7 @@ class _NoOpDriverCtx:
 
 def save(
     self: GT,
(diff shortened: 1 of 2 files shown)


rich-iannone (approved):
LGTM!

Issue #178, opened 2024-02-06
epic: Implement rendering `GT.to_latex()`


machow:
Currently, we only render to HTML. It would be great to support latex.

jrycw:
I'll close this issue as the feature has been implemented and merged.

Issue #75, opened 2023-12-11
Option to export tables to LaTeX code


aeturrell:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal

It would be fantastic if there was an option to export tables to LaTeX as this would unlock a wide range of further uses in presentations, papers, and reports. For prior art, the original **gt** package in R does this, and **pandas** has a table export option.

Thanks for the great work on great tables!

rich-iannone:
Yes! We are planning for this, and it'll be something that's done in early-to-mid 2024. The reason for it being scheduled a bit later is so we could first get the core functionality into this package working well.

aeturrell:
Amazing news, thank you.

apoorvkh:
Just wondering, do you think this will be implemented by the next version? Or a longer timeline?

Would love to ditch my [`pandas` dependency](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_latex.html) and rely on `polars` and `great-tables` instead!

machow:
We should have the more fine-grained styling options implemented pretty soon (https://github.com/posit-dev/great-tables/pull/341), which will be useful to have in place before implementing latex support. Lemme check with @rich-iannone in our next meeting on timing 😅

apoorvkh:
Great, thanks so much for the update!

fzyzcjy:
+1 that would be super helpful!

jrycw:
I'll close this issue as the feature has been implemented and merged.

fzyzcjy:
Congratulations!

aeturrell:
This is fantastic news, thank you for implementing. I will add to [Coding for Economists](https://aeturrell.github.io/coding-for-economists/intro.html) in due course! ([Issue here](https://github.com/aeturrell/coding-for-economists/issues/83))

PR #481, opened 2024-10-02
feat: allow rendering of tables to LaTeX

rich-iannone:
This feature PR implements the rendering of tables to LaTeX, addressing https://github.com/posit-dev/great-tables/issues/75 and https://github.com/posit-dev/great-tables/issues/178. This will be exposed through the `as_latex()` method on the `GT` class. If in Quarto and the render target is LaTeX, tables will be automatically rendered as LaTeX tables.

Preview: https://pr-481--gt-python.netlify.app/a-latex_examples/index.pdf.


great_tables/gt.py
```diff
@@ -71,6 +71,20 @@
     create_heading_component_h,
     create_source_notes_component_h,
 )
+from great_tables._utils_render_latex import (
```
machow, line 74: Can you move these out of `gt.py` along with the definition `_render_as_latex()`? You could still import `_render_as_latex()` from somewhere and add it as a method.

But as it exists, this code is...

* importing `as_latex`, which...
* calls `_render_as_latex()`, that is defined in this file, which...
* calls the `create_*()` functions defined in `_utils_render_latex`

So the path is going...

* `_export.as_latex()` -> `gt.py` -> `_utils_render_latex()`

Ideally it seems like it should just be going...

* `_export.as_latex()` -> `_utils_render_latex()`

As then we just assign the function to GT.as_latex()

great_tables/_gt_data.py
```diff
@@ -177,6 +183,50 @@ def render_formats(self, data_tbl: TblData, formats: list[FormatInfo], context:
 
         return self
 
+    def migrate_unformatted_to_output(
```
machow, line 186: let's move this out to another file (and make a function), so it isn't a method exposed to users

great_tables/_text.py
```diff
@@ -26,24 +26,53 @@ def _md_html(x: str) -> str:
...
+        if isinstance(x, Md):
+            return _md_html(x.text)
+        elif isinstance(x, Html):
+            return x.text
+        elif isinstance(x, str):
+            return _html_escape(x)
+        elif isinstance(x, Text):
+            return x.text
+        elif isinstance(x, UnitStr):
+            return x.to_html()
```
machow, line 54: What if we made all text type x context rendering functions methods?

For example:

* text type - Md, context - html: `Md.to_html()`
* text type - Md, context - latex: `Md.to_latex()`
* text type - Text, context - latex: `Text.to_latex()`
  - This could just return the raw text

This way, no matter whether you get Md, Text, or HTML, you can just call `.to_html()` or `.to_latex()`, so we don't need to do specific isinstance checks on them.  (the `.to_<context>()` becomes an interface).

great_tables/_formats.py
machow, line 1: Please add comments to this file, explaining what changes were made. The text diff is large, but from pairing it sounds like the effective changes are small, so just quick comments (in the github PR code) explaining what was changed would help a lot.

machow (changes requested):

great_tables/gt.py
rich-iannone: Thanks! Completed during code pairing.

great_tables/_formats.py
```diff
@@ -26,7 +27,7 @@
     to_list,
     _get_column_dtype,
 )
-from ._text import _md_html
+from ._text import _md_html, escape_pattern_str_latex
```
rich-iannone, line 30: The new function `escape_pattern_str_latex()` is used here for escaping string literals in the `pattern=` arg. Only LaTeX for now, we could later generalize to include the `"html"` context later.

great_tables/_formats.py
```diff
@@ -285,66 +286,129 @@ def fmt_number(
     sep_mark = _get_locale_sep_mark(default=sep_mark, use_seps=use_seps, locale=locale)
     dec_mark = _get_locale_dec_mark(default=dec_mark, locale=locale)
 
-    # Generate a function that will operate on single `x` values in the table body
```
rich-iannone, line 288: The `fmt_number_fn()` function is extracted from the body of the `fmt_number()` function.

This is a general pattern that will be used throughout this file. The extracted function is migrated to a new function that has a `context=` argument (for `"html"` and `"latex"` contexts). This function will be used in a `partial()` statement for each of the supported contexts (all within a `fmt()` call).

great_tables/_formats.py
```diff
@@ -285,66 +286,129 @@ def fmt_number(
...
-        is_negative = _has_negative_value(value=x)
-
-        if compact:
-            x_formatted = _format_number_compactly(
-                value=x,
+    # TODO: fix type errors for members of `FormatFns`
+    return fmt(
+        self,
+        fns=FormatFns(
+            html=partial(
```
rich-iannone, line 293: While it's hard to see in the diff, the `FormatFns()` call will always use three named parameters: `html`, `latex`, and `default`. Each of these will have a `partial` statement so that variables are injected as default values. The args of `partial()` are always of this pattern/order:

- the function `fmt_*_context()`
- `data=self`
- other args for the `fmt_*_context()` function
- `context=`, where the context string is hard-coded, matching the named parameter in `FormatFns()`

The `fmt_*_context()` has an `x` argument which is untouched by `partial()`. When the function is later invoked, `x` will be the list of values (e.g., from a column).

great_tables/_formats.py
```diff
@@ -285,66 +286,129 @@ def fmt_number(
...
+            drop_trailing_dec_mark=drop_trailing_dec_mark,
+            use_seps=use_seps,
+            sep_mark=sep_mark,
+            dec_mark=dec_mark,
+            force_sign=force_sign,
+        )
+
+    # Implement minus sign replacement for `x_formatted`
+    if is_negative:
+        minus_mark = _context_minus_mark(context=context)
```
rich-iannone, line 399: The `_context_minus_mark()` call here switches on the `context` value, providing the correct glyph for the output type. Other functions that follow this pattern are named as `_context_*_mark()`.

great_tables/_formats.py
```diff
@@ -285,66 +286,129 @@ def fmt_number(
...
+        x_formatted = _replace_minus(x_formatted, minus_mark=minus_mark)
+
+    # Use a supplied pattern specification to decorate the formatted value
+    if pattern != "{x}":
+
+        # Escape LaTeX special characters from literals in the pattern
+        if context == "latex":
+            pattern = escape_pattern_str_latex(pattern_str=pattern)
```
rich-iannone, line 407: The function `escape_pattern_str_latex()` is used throughout the formatting functions here to modify the `pattern` string for the `"latex"` context only (for this PR). These three lines will be included in every `if pattern != "{x}":` conditional below.

great_tables/_formats.py
```diff
@@ -663,112 +778,175 @@ def fmt_scientific(
...
 
+        if small_pos:
+            # If the value is small enough to not require the (x 10^n) notation, then
+            # the formatted value is based on only the `m_part`
+            x_formatted = m_part
         else:
-            # Define the exponent string based on the `exp_style` that's not the default
-            # value of 'x10n'
+            # Get the set of exponent marks, which are used to decorate the `n_part`
+            exp_marks = _context_exp_marks(context=context)
```
rich-iannone, line 910: The `_context_exp_marks()` function is new here since scientific notation formatting is different in HTML and LaTeX. This follows the convention of getting the correct text fragment(s) for the rendering context.

great_tables/_formats.py
```diff
@@ -663,112 +778,175 @@ def fmt_scientific(
...
 
-        return x_formatted
+        x_formatted = m_part + exp_str + n_part
+
+    # Use a supplied pattern specification to decorate the formatted value
+    if pattern != "{x}":
+
+        # Escape LaTeX special characters from literals in the pattern
+        if context == "latex":
+            pattern = escape_pattern_str_latex(pattern_str=pattern)
```
rich-iannone, line 945: Using `escape_pattern_str_latex()` as before to escape the `pattern` specifically for LaTeX.

great_tables/_formats.py
```diff
@@ -919,72 +1097,139 @@ def fmt_percent(
...
+        rows=rows,
+    )
 
-        # Scale `x` value by a defined `scale_by` value
-        x = x * scale_by
 
-        # Determine properties of the value
-        is_negative = _has_negative_value(value=x)
-        is_positive = _has_positive_value(value=x)
+def fmt_percent_context(
```
rich-iannone, line 1158: The pattern continues, which is:

- extract internal formatting function to one that generalizes to different output contexts (naming it w/ suffix `_context`)
- replace `fmt()` statement in `fmt_*()` (e.g., `fmt_percent()`) with one that covers three contexts (`html`, `latex`, and `default` <- same as `html`) with `partial()`ed versions of the `fmt_*_context()` function

great_tables/_utils_render_latex.py
```diff
@@ -0,0 +1,1001 @@
...
+    "px": 1.0,
+    "pt": 4 / 3,
+    "in": 96.0,
+    "cm": 37.7952755906,
+    "emu": 1 / 9525,
+    "em": 16.0,
+}
+
+
+class WidthDict(TypedDict):
```
machow, line 27: Can you change this away from

* WidthDict: A class full of lists that will all become `len(range(n))`

To a list of width objects:

* Use a class of non-list entries. E.g. a dataclass called EntryWidth, with type, unspec, etc..
* A list of len n, where each entry is an EntryWidth instance
* Totally legit to define a container dataclass like XxWidths, that
  - holds the list[EntryWidth] and tbl_width
  - name w/e feels more appropriate

great_tables/_utils_render_latex.py
```diff
@@ -0,0 +1,1001 @@
+from __future__ import annotations
```
machow, line 1: It seems like there are some larger untested blocks. Can you take a look at this file in codecov, and test parts of some of the bigger blocks?

If rendering row / group names is unsupported, can you ensure it raises an error? (that path appears untested)

https://app.codecov.io/gh/posit-dev/great-tables/pull/481/blob/great_tables/_utils_render_latex.py

machow (changes requested):
It seems like comments from my earlier review are unresolved (I've commented on them again to be sure they surface). I looked through and left some quick feedback. Once it's addressed, I can do a deeper dive!

great_tables/_export.py
```diff
@@ -166,6 +168,30 @@ def as_raw_html(
...
+    gt
+        A GT object.
+
+    Returns
+    -------
+    str
+        An LaTeX fragment containing a table.
+
+    """
+    built_table = self._build_data(context="latex")
```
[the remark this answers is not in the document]
rich-iannone: I think the comment on `great_tables/_text.py` refers to an older version. Could you check the latest version of `great_tables/_text.py`? I'm using `.to_*()` methods there.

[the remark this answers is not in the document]
machow: Oh shoot -- I see them now, sorry for the mix up 😓

great_tables/_utils.py
```diff
@@ -7,7 +7,7 @@
 import re
 from collections.abc import Generator
 from types import ModuleType
-from typing import Any, Iterable
+from typing import Any, Iterable, Callable
```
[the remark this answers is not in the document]
rich-iannone: Now removed.

great_tables/_export.py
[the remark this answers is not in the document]
rich-iannone: It's okay! Just moved `migrate_unformatted_to_output()` to `utils.py` for now. And it's a function now. There are some type annotation issues that I flagged in a TODO comment (circular dependency issue when importing the parameter types). This could perhaps be resolved by placing in a different file (will experiment later).

great_tables/quarto.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+
+def check_quarto() -> bool:
```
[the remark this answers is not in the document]
rich-iannone: Thanks for the suggestion here. I went with `is_quarto_render()`, which is much better than `check_quarto()`.

great_tables/_text.py
```diff
@@ -52,3 +81,47 @@ def _process_text_id(x: str | Text | None) -> str:
...
+    match the pattern. The processed elements are then recombined to obtain a selectively processed
+    string.
+
+    Args:
+        string (str): The string to process.
+        pattern (str): The pattern to split the string by.
+        func (Callable[[str], str]): The function applied to elements that do not match the pattern.
+
+    Returns:
+        str: The selectively processed string.
```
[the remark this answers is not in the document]
rich-iannone: Now changed.

great_tables/_text.py
```diff
@@ -52,3 +81,47 @@ def _process_text_id(x: str | Text | None) -> str:
...
+def escape_pattern_str_latex(pattern_str: str) -> str:
+
+    pattern = r"(\{[x0-9]+\})"
+
+    return process_string(pattern_str, pattern, _latex_escape)
+
+
+def process_string(string: str, pattern: str, func: Callable[[str], str]) -> str:
+    """
+    Process a string selectively based on a pattern.
```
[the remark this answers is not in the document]
rich-iannone: Now much more clear what the function does.

great_tables/_text.py
```diff
@@ -52,3 +81,47 @@ def _process_text_id(x: str | Text | None) -> str:
...
+
+    This function splits a string based on a pattern and applies a function to elements that do not
```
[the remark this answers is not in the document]
rich-iannone: Did some rewriting here and it's quite a bit clearer now.

great_tables/_utils.py
```diff
@@ -9,7 +9,9 @@
 from types import ModuleType
 from typing import Any, Iterable
 
-from ._tbl_data import PdDataFrame
+
```
machow, line 12: You can work around circular imports for type annotations by doing

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import ...
```

great_tables/_text.py
```diff
@@ -12,38 +12,65 @@
...
-        return x.to_html()
-    else:
-        raise TypeError(f"Invalid type: {type(x)}")
+    escape_fn = _html_escape if context == "html" else _latex_escape
+
+    if isinstance(x, str):
+
+        return escape_fn(x)
+
+    elif isinstance(x, (Md, Text, Html, UnitStr)):
```
machow, line 69: From pairing, we could create a parent class for these pieces (e.g. Md and Text inherit from BaseText, etc..)

great_tables/_utils_render_latex.py
machow: * The type hint for `type:` should be `ColInfoTypeEnum`
* The type hint for `column_align` should be _gt_data.ColumnAlignment

machow: It looks like pieces like `type:` are pulled out specifically, so we later can essentially manually do a `boxhead[i].visible` check. I wonder if just storing the original ColInfo object for each element would let us do these checks directly (and then we don't need to store just the `type` piece).

machow (changes requested):
From pairing -- this is looking good! Here are the notes we wrote down!

great_tables/_formats.py
```diff
@@ -663,112 +778,175 @@ def fmt_scientific(
...
-        sep_mark: str = sep_mark,
-        dec_mark: str = dec_mark,
-        force_sign_m: bool = force_sign_m,
-        force_sign_n: bool = force_sign_n,
-    ):
-        # If the `x` value is a Pandas 'NA', then return the same value
-        if is_na(self._tbl_data, x):
-            return x
+    # TODO: fix type errors for members of `FormatFns`
+    return fmt(
```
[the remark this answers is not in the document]
rich-iannone: This is now done.

great_tables/_formats.py
```diff
@@ -919,72 +1097,139 @@ def fmt_percent(
...
-        dec_mark: str = dec_mark,
-        force_sign: bool = force_sign,
-        placement: str = placement,
-        incl_space: bool = incl_space,
-    ):
-        # If the `x` value is a Pandas 'NA', then return the same value
-        if is_na(self._tbl_data, x):
-            return x
+    # TODO: fix type errors for members of `FormatFns`
+    return fmt(
```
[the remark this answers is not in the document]
rich-iannone: Done.

great_tables/_formats.py
```diff
@@ -1151,79 +1396,144 @@ def fmt_currency(
...
```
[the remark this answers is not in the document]
rich-iannone: Done.

great_tables/_formats.py
```diff
@@ -1290,50 +1600,85 @@ def fmt_roman(
...
-        case: str = case,
-    ):
-        # If the `x` value is a Pandas 'NA', then return the same value
-        if is_na(self._tbl_data, x):
-            return x
-
-        # Get the absolute value of `x` so that negative values are handled
-        x = abs(x)
+    # TODO: fix type errors for members of `FormatFns`
+    return fmt(
```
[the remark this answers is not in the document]
rich-iannone: Done.

great_tables/_utils.py
```diff
@@ -237,3 +239,53 @@ def seq_groups(seq: Iterable[str]) -> Generator[tuple[str, int], None, None]:
...
+    # Get all visible cells in the table
+    all_visible_cells = _get_visible_cells(data=data_tbl)
+
+    # Get the difference between the visible cells and the formatted cells
+    all_unformatted_cells = list(set(all_visible_cells) - set(all_formatted_cells))
+
+    # TODO: this currently will only be used for LaTeX (HTML escaping will be performed
+    # in the future)
+    if context == "latex":
```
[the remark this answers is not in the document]
rich-iannone: Now done.

great_tables/_utils_render_latex.py
```diff
@@ -0,0 +1,800 @@
...
+        This dictionary is generated by the `create_width_dict_l()` function.
+
+    Returns
+    -------
+    str
+        The LaTeX code that signifies the start of the table.
+    """
+
+    # Is the longtable environment being used?
+    latex_use_longtable = data._options.latex_use_longtable.value
```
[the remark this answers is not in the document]
rich-iannone: Options are now args in the LaTeX rendering method and function.

great_tables/_gt_data.py
rich-iannone: Now moved to `_utils.py`.

great_tables/quarto.py
```diff
@@ -0,0 +1,8 @@
...
+
+    # Check for the presence of the QUARTO_BIN_PATH environment variable
+    # to determine if the current environment is a Quarto environment.
+    return "QUARTO_BIN_PATH" in os.environ
```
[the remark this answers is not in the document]
rich-iannone: Both things done. From checking with the Quarto team, this is currently the best way to check that a Quarto render is occurring. Added a docstring here to explain the purpose of the function.

great_tables/_utils_render_latex.py
rich-iannone: From some refactoring, this whole piece was removed (it's never actually used in the code). It will be used once we have the stub and row groups rendered, but right now it's unused code.

great_tables/_utils_render_latex.py
rich-iannone: Now, pretty much everything added in this PR has tests.

rich-iannone: Errors are raised for row / group names usage. And that's formally covered in unit tests as well.

great_tables/_boxhead.py
```diff
@@ -5,14 +5,14 @@
 from ._locations import resolve_cols_c
 from ._utils import _assert_list_is_subset
 from ._tbl_data import SelectExpr
-from ._text import Text
+from ._text import BaseText
```
machow, line 8: Please don't put sweeping refactor style changes in feature PRs

rich-iannone: Now fixed up during code review.

machow (changes requested):
From pairing -- we ran the examples from https://posit-dev.github.io/great-tables/examples/ .

Main pieces:

* Explicitly raise or warn for currently unsupported behaviors (e.g. styles)
* formatters sometimes run but their output generate latex errors
  - fmt_nanoplot
  - fmt_markdown
  - fmt_image
  - fmt_url
  - fmt_integer (?) -- this seems to raise a python error
* the `md()` and `html()` functions produce output that causes latex errors (but maybe this is okay?)

rich-iannone:
* Explicitly raise or warn for currently unsupported behaviors (e.g. styles)

Now we warn when there is any definition of `tab_style()` in LaTeX output.

* formatters sometimes run but their output generate latex errors
  - fmt_nanoplot
  - fmt_markdown
  - fmt_image
  - fmt_url
  - fmt_integer (?) -- this seems to raise a python error

- fmt_integer() is fine. No error on second look
- fmt_url() is not in API.
- fmt_image now provides a warning and formatting is skipped
- fmt_markdown raises a NotImplementedError in LaTeX render
- fmt_nanoplot also raises a NotImplementedError

* the `md()` and `html()` functions produce output that causes latex errors (but maybe this is okay?)

- md() still raises a NotImplementedError
- html() will warn and perform LaTeX escaping on the string

machow:
From pairing, LGTM -- thanks for this ENORMOUS push, excited for latex tables!

Issue #466, opened 2024-09-27
Test and harden selection in `loc.*()` functions


machow:
There are 5 loc functions that can do selection based on the data. While they're wired up in #341, we should do more thorough testing of them:

- [ ] : `loc.row_groups(groups=...)`
- [ ] : `loc.stub(rows=...)`
- [ ] : `loc.column_labels(columns=...)`
- [ ] : `loc.column_spanners(columns=...)`
- [x] : loc.body()

PR #484, opened 2024-10-04
Fix locations: row and group selection

machow:
This PR addresses #466 by ensuring that group and row selection work across three scenarios:

* position number
* name string
* expression


machow:
Note that column spanners currently can't do selection by polars selectors (and the concept of selection by position doesn't make sense for them, since they stuck vertically)

machow:
Restructured the Get Started guide sections, to split into structure, format, style, theme (and selection help)

image

rich-iannone (approved):
LGTM!

PR #485, opened 2024-10-05
Add `GT.write_raw_html()` as a helper function for easier HTML output

jrycw:
Hello team,

I've noticed that some users expect an easier way to interact with our tables. While `GT.as_raw_html()` is great, it doesn’t fully meet those expectations. To address this, I’d like to propose adding `GT.write_html()` as a helper function, simplifying the process of writing the table’s HTML directly to a file without needing to use `open()`.

I believe this is a useful addition, and if necessary, we could even rename it to `GT._write_html()` for unofficial use.


jrycw:
I've found that utilizing the parameters of `GT.as_raw_html()` might be a better approach. I suggest we keep the `make_page` and `all_important` parameters undocumented, similar to how it's done in `GT.as_raw_html()`.

machow:
Thanks for this @jrycw -- @rich-iannone and I are pairing on this right now, and thinking about the use of both `.as_*()` and `.write_*()` in the Great Tables API.

We noticed that polars `.write_*()` methods will return a string, if not filename is specified. WDYT of that approach. So in that case, we would deprecate the `as_*()` methods in favor of `write_*()` methods.

Basically, it seems like there are 3 possible approaches:

* **polars**: `.write_*()` methods return a string when no filename given
* **two kinds of methods**: each render target has a `.as_*()` and `.write_*()` method.
* **polars approach, but with `.as_*()`:** use `.as_*()` instead of `.write_*()`

Would love to get your input!

jrycw:
Hello Team, 

I’m inclined to go with the first option as the higher-level abstraction while retaining `.as_*()` for internal use.

jrycw:
To issue a deprecation message to users, we could take a similar approach to what [Polars uses](https://github.com/pola-rs/polars/blob/py-1.13.0/py-polars/polars/_utils/deprecation.py#L47-L64). We might consider decorating `as_raw_html()` in the GT object, like this:

```python
class GT(GTData):
    ...

    as_raw_html = deprecate_function("Use `GT.write_html()` instead.", version="0.15.0")(as_raw_html) 
    write_html = write_html
```

This method avoids directly decorating `as_raw_html()`, which could trigger the message unintentionally when using `write_html()`. Once we’re ready for full deprecation, we can simply remove the line: `as_raw_html = ...`.

rich-iannone:
With some discussion with @machow we decided on using this option (from the list of three further above)

- two kinds of methods: each render target has a `.as_*()` and `.write_*()` method.

The reasoning is that we don't have a lot of conversion methods so we don't have to be too economical on the number of export methods. Plus having output type stability would make more clear what the intended effect of the method is. 

So the exported methods needed are:

- `as_raw_html()`
- `write_raw_html()`
- `as_latex()`
- `write_latex()`

The last one `write_latex()` could be done here or in another PR (up to you).

jrycw:
I suspect the doc build failure might be related to the change in Polars (`pl.DataFrame.deserialize`).

docs/_quarto.yml
```diff
@@ -214,11 +214,11 @@ quartodoc:
       desc: >
         There may come a day when you need to export a table to some specific format. A great method
         for that is `save()`, which allows us to save the table as a standalone image file. You can
-        also get the table code as an HTML fragment with the `as_raw_html()` method.
+        also get the table code as an HTML fragment with the `write_raw_html()` method.
```
rich-iannone, line 217:
```suggestion
        also get the table code as an HTML fragment with the `*_raw_html()` methods.
```

docs/_quarto.yml
```diff
@@ -214,11 +214,11 @@ quartodoc:
...
       contents:
         - GT.save
         - GT.show
-        - GT.as_raw_html
+        - GT.write_raw_html
```
rich-iannone, line 221:
```suggestion
        - GT.as_raw_html
        - GT.write_raw_html
```

rich-iannone (changes requested):
Looks great! Two small suggestions would make it perfect.

great_tables/_export.py
```diff
@@ -559,3 +559,47 @@ def _dump_debug_screenshot(driver, path):
...
+
+
+def write_raw_html(
+    gt: GT,
+    filename: str | Path | None = None,
+    encoding: str = "utf-8",
+    newline: str | None = None,
+    make_page: bool = False,
+    all_important: bool = False,
+) -> str | None:
```
machow, line 571: @rich-iannone you mentioned type stability, but this does not have the stability you mentioned. Is this what you want in the interface?

rich-iannone: Thanks for noting this. @jrycw could the signature be this instead:

```
def write_raw_html(
     gt: GT,
     filename: str | Path,
     encoding: str = "utf-8",
     newline: str | None = None,
     make_page: bool = False,
     all_important: bool = False,
 ) -> None:
```

Then these lines would need to be removed:

```
if filename is None:
         return html_content
```

Since we are requiring a `filename=` value, the last sentence of `filename` parameter description in the docstring needs to removed. Also, the description in Returns needs to change to fit a return of `None`.

great_tables/_export.py
```diff
@@ -559,3 +559,43 @@ def _dump_debug_screenshot(driver, path):
...
+        The name of the file to save the HTML. Can be a string or a `pathlib.Path` object.
+    encoding
+        The encoding used when writing the file. Defaults to 'utf-8'.
+    newline
+        The newline character to use when writing the file. Defaults to `os.linesep`.
+    Returns
+    -------
+    str | None
+        If `filename` is `None` (default), returns the HTML output as a string. Otherwise, writes
+        the HTML to the specified file path and returns `None`.
```
rich-iannone, line 592:
```suggestion
    None
        An HTML file is written to the specified path and the method returns `None`.
```

rich-iannone: ^^ is a docs tweak to fit with the revised interface.

great_tables/_export.py
machow: I'm down for whatever seems reasonable -- in this case it sounds like the intended spec is...

**Polars**:
  - `to_*()` methods return objects.
  - `write_*()` methods return None if writing to file, string otherwise.

**Great Tables**:
  - `as_*()` methods return objects.
  - `write_*()` methods always return None (and filename is mandatory).

machow:
@rich-iannone feel free to merge after re-reviewing

rich-iannone (approved):

rich-iannone:
This looks great now. Will merge.

great_tables/_export.py
jrycw: @rich-iannone Thanks! 😳

PR #487, opened 2024-10-05
feat: add pyarrow.Table support

amol-:
# Summary

Allows to directly use ``pyarrow.Table`` objects without having to cast them to pandas (which is an optional dependency of pyarrow). This is also convenient as more and more libraries are supporting a ``to_arrow()`` method and thus would allow to use data from those libraries too.

# Checklist


amol-:
While I added tests for ``tbl_data`` it wasn't immediately clear to me if e2e tests for each table library exist and how to add one.  I updated the tests where ``@parametrize`` was used to test across multiple backends

great_tables/_tbl_data.py
```diff
@@ -207,27 +246,46 @@ def _set_cell(data: DataFrameLike, row: int, column: str, value: Any):
...
     col_indx = data.columns.get_loc(column)
     data.iloc[row, col_indx] = value
+    return data
 
 
 @_set_cell.register(PlDataFrame)
-def _(data, row: int, column: str, value: Any) -> None:
+def _(data, row: int, column: str, value: Any) -> PlDataFrame:
     data[row, column] = value
+    return data
```
machow, line 260: Does they PyArrowTable version return a copy of the data? For polars and pandas, I think returning None is useful for [command-query separation](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation) (where `_set_cell` is a command, so shouldn't return a value).

But it seems totally fair if PyArrowTable can't  be mutated. (I'm guessing at some point we'll want to change `_set_cell` to set entire columns, anyway?!)

amol-: Correct, in PyArrow data is immutable, so it is not possible to mutate in-place the Table (at least from public APIs point of view) and by design of Arrow it is considered bad practice to mutate it in general.

I understand the point of command-query separation, my approach has been to retain as much as possible of the existing polars/pandas implementation of the functions (to focus PR changes to Arrow), and thus the implementations ended up mixing the two behaviours between the various backends, but I think that the cleanest approach would be to move all of them to return a copy instead of mutating it in-place. That seems to be more future-proof as it doesn't constraint the codebase to only work with mutable structures.

machow:
Thanks for opening this PR! It seems like adding pyarrow support is a good chance for us to clean up our tests of different backends a bit, and make sure the tests in tbl_data cover the range of use cases.

If we could add a snapshot test (would be okay to put at the bottom of `test_tbl_data.py` I think), I'd be on board with merging this. We can note that it's in development, while tbl_data tests get cleaned up. (I'm happy to add the snapshot to this PR)

amol-:
> If we could add a snapshot test (would be okay to put at the bottom of `test_tbl_data.py` I think), I'd be on board with merging this. We can note that it's in development, while tbl_data tests get cleaned up. (I'm happy to add the snapshot to this PR)

Happy to add a test, not a problem 👍

great_tables/_tbl_data.py
amol-: PS: Given that the scope of great-tables is rendering tables for human consumption, we should be talking about relatively small data, so I think that even involving `Dataframe.copy` on pandas should introduce an acceptable cost

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 exclude: "(.*\\.svg)|(.*\\.qmd)|(.*\\.ambr)|(.*\\.csv)|(.*\\.txt)|(.*\\.json)"
 repos:
   - repo: https://github.com/pycqa/flake8
-    rev: 6.0.0
+    rev: 7.1.1
```
amol-, line 4: I had to update flake8 because the used version seemed to have a bug complaining about the content of f-strings:

```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

great_tables/_tbl_data.py:803:70: E231 missing whitespace after ':'
great_tables/_tbl_data.py:809:33: E221 multiple spaces before operator
```

great_tables/_tbl_data.py
amol-: https://github.com/posit-dev/great-tables/pull/487/commits/d40040b00d5738ae09e42c1ba71b28f6ca5efea0 consolidates the behavior of the functions

amol-:
> > If we could add a snapshot test (would be okay to put at the bottom of `test_tbl_data.py` I think), I'd be on board with merging this. We can note that it's in development, while tbl_data tests get cleaned up. (I'm happy to add the snapshot to this PR)
> 
> Happy to add a test, not a problem 👍

Test added in https://github.com/posit-dev/great-tables/pull/487/commits/d3cdd1c06cc92e5691ed52ac6767ad3047553993

But I couldn't find a way to avoid duplicating the snapshot that didn't involve writing a custom AmberSnapshotExtension

great_tables/_utils.py
```diff
@@ -136,7 +136,7 @@ def _object_as_dict(v: Any) -> Any:
         return list(_object_as_dict(i) for i in v)
     if isinstance(v, dict):
         return dict((k, _object_as_dict(val)) for (k, val) in v.items())
-    if type(v) == type(_object_as_dict):  # FIXME figure out how to get "function"
+    if type(v) is type(_object_as_dict):  # FIXME figure out how to get "function"
```
amol-, line 139: This is to satisfy newer flake8: https://github.com/amol-/great-tables/actions/runs/11259168945/job/31307498137

```
great_tables/_utils.py:139:8: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```
but indeed it was a good idea anyway

machow: It seems like maybe this check was trying to do either `if callable(v):` or maybe `inspect.isfunction(v)` 😓

machow (changes requested):
Thanks again for this -- I'd be on board merging this with an experimental warning! The one thing I think that might be important is if we could revert back the changes to the pandas and polars implementation of `_set_cell`, just to keep it scoped onto PyArrow.

I should be able to tweak the use of `_set_cell` to be column focused down the road, which should hopefully work better with frame libraries 😓 

(Happy to make any of these changes and merge if useful!)

great_tables/_gt_data.py
```diff
@@ -173,7 +174,7 @@ def render_formats(self, data_tbl: TblData, formats: list[FormatInfo], context:
                 # TODO: I think that this is very inefficient with polars, so
                 # we could either accumulate results and set them per column, or
                 # could always use a pandas DataFrame inside Body?
-                _set_cell(self.body, row, col, result)
+                self.body = _set_cell(self.body, row, col, result)
```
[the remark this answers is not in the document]
amol-: sounds a reasonable approach

amol-:
@machow I think all requested changes should have been addressed

machow:
Hey, thanks for putting all this work in on the PR! I'm just getting back from being OOO, and can work on merging next week!

machow:
Ok, thanks for infinite patience here -- I'm planning to take a look on Monday

machow (approved):
LGTM! Thanks @amol- for pushing this through -- it's really neat to see how things wire up to pyarrow! (And for us to have to think about our `_set_cell` behaviors...)

machow:
@rich-iannone do you want to merge?

PR #496, opened 2024-10-19
Add `save_pdf` method that directly renders table HTML to PDF

lschr:
# Summary

Generate a real vector graphics PDF without depending on a web browser.

The `weasyprint` package is used to generate the PDF, followed by `pdfplumber` to determine the bounding box of table elements, and finally `pypdf` writes the cropped PDF.

This could also be used to replace the `save` method, e.g. using `PyMuPDF` to render the resulting PDF into a raster image. Should this be desired, `PyMuPDF` could (probably) also be used for cropping, removing the need for `pypdf`.

# Checklist

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -208,10 +208,12 @@ quartodoc:
     - title: Export
       desc: >
         There may come a day when you need to export a table to some specific format. A great method
-        for that is `save()`, which allows us to save the table as a standalone image file. You can
-        also get the table code as an HTML fragment with the `as_raw_html()` method.
+        for that is `save()`, which allows us to save the table as a standalone image file. To render
+        to a vector graphics PDF file, use `save_pdf()`. You can also get the table code as an HTML
+        fragment with the `as_raw_html()` method.
       contents:
         - GT.save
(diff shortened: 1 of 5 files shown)


lschr:
The Windows test fails because weasyprint depends on Pango. How could/should this be handled?

- Point the user to [weasyprint documentation](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#windows) (or replicate), i.e., have the user install pango via MSYS2.
- Try to implement a function to download and use MSYS2 libraries via [msys2dl](https://pypi.org/project/msys2dl/). If such a function is possible, the user would have to call it once, which is probably more convenient than installing MSYS2 and pango manually.
- Other suggestions?

machow:
Hey, thanks for working on this -- it seems really nice to have an option to go straight from HTML to PDF, without a web browser.

I noticed that the full extent of Great Tables is calling `GT.as_raw_html()`, so that this work is largely a handy way to go from a wide range of html to pdf.

WDYT of this being in its own library rather than in Great Tables? It seems useful, but the added maintenance burden would be challenging inside Great Tables.

PR #499, opened 2024-10-28
Improve `GT.save()` usability

jrycw:
Related PR: #496.

This PR introduces several improvements to enhance the usability of `GT.save()`:

* Propose exposing `**params` to allow advanced users to customize save parameters directly via [Pillow](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save).
* Streamline the `.png` extension check to a single line.
* Construct the URL directly, similar to `FmtImage._get_image_uri()`, without using `tempfile.TemporaryDirectory()`.
* Replace the final `time.sleep(0.05)` with `WebDriverWait(driver, 1).until()` (we may need to determine the optimal timeout for most use cases).
* Remove the `.png` saving branch that relied on `selenium`; all files are now saved using `Pillow`.
* Suggest modifying `GT.save()` to return itself, allowing users to save intermediate tables. For example:
  ```python
  GT(exibble).save("initial.png").tab_header("title").save("with_title.png")
  ```

Additionally, I noticed a potential speed boost (approximately 40-50% faster on my machine) when calling `headless_browser=wdriver(options=wd_options)` directly, bypassing the context manager. However, I'm unsure about the safety implications of this approach.

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -369,6 +369,7 @@ def _save_screenshot(
 
     from PIL import Image
 
+    # convert to other formats (e.g. pdf, bmp) using PIL
     Image.open(fp=BytesIO(el.screenshot_as_png)).save(fp=path)
 
 
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -188,7 +188,6 @@ def save(
     debug_port: None | int = None,
     encoding: str = "utf-8",
     _debug_dump: DebugDumpOptions | None = None,
-    **params,
 ) -> GTSelf:
     """
     Produce a high-resolution image file or PDF of the table.
@@ -227,14 +226,6 @@ def save(
         Whether the saved image should be a big browser window, with key elements outlined. This is
         helpful for debugging this function's resizing, cropping heuristics. This is an internal
         parameter and subject to change.
-    **params
-        Additional parameters supported by
-        [Image.save()](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save)
-        in Pillow. For instance, when saving the table as a
-        [PNG](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#png), you can
-        adjust the `compress_level` to balance between speed and compression. The `compress_level`
-        ranges from 0 to 9, where 1 offers the best speed, 9 offers maximum compression, and 0
-        applies no compression. The default value is 6.
 
     Returns
     -------
@@ -286,7 +277,7 @@ def save(
         encoded = base64.b64encode(html_content.encode(encoding=encoding)).decode(encoding=encoding)
         headless_browser.get(f"data:text/html;base64,{encoded}")
 
-        _save_screenshot(headless_browser, scale, file, debug=_debug_dump, **params)
+        _save_screenshot(headless_browser, scale, file, debug=_debug_dump)
 
     if debug_port and web_driver not in {"chrome", "firefox"}:
         warnings.warn("debug_port argument only supported on chrome and firefox")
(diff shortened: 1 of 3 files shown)


jrycw:
Hello team,  

In my last commit, I experimented with isolating the web driver preparation logic into `_utils_selenium.py`. Please feel free to review and decide whether to keep or discard this change.

great_tables/_utils_selenium.py
```diff
@@ -0,0 +1,100 @@
...
+    ) -> bool | None:
+        self.driver.quit()
+
+
+class _ChromeWebDriver(_BaseWebDriver):
+    def __init__(self, debug_port: int | None = None):
+        self.debug_port = debug_port
+        self.wd_options = webdriver.ChromeOptions()
+        super().__init__()
+        self.driver = webdriver.Chrome(self.wd_options)
```
machow, line 41: Could we consolidate the driver initialization across all these subclasses, by adding `cls_wd_options=ChromeOptions`, and `cls_driver=Chrome` as class attributes?

That way, the BaseClass could essentially require subclasses to specify those pieces. Then, the init would not have to be overridden by child classes.

This makes the child classes mostly about:

* specifying Driver and Option classes (as class attributes)
* customizing add_arguments method

jrycw: @machow , thanks for the quick response.  

Using `cls_driver` seems like a good idea, but I’m concerned that `cls_wd_options` might lead to issues. For example, what if we need to create two `Chrome` instances with different `ChromeOptions` in the future?

jrycw: @machow , I think I understand now. Does the latest commit align with what you had in mind?

rich-iannone:
This is looking good! I had a look at the output images on all 4 webdrivers (at 3 different scale values) to make sure outputs are similar to `main`. In case this is interesting I've included the images in this zip file:

[image-tests.zip](https://github.com/user-attachments/files/17806470/image-tests.zip)

I've found no reduction in quality with the changes here. Chrome is the only one that is problematic (cuts off content at the bottom) but that's also on `main` and an open issue anyway.

great_tables/_utils_selenium.py
machow: Yeah, thanks -- that looks really nice!

great_tables/_export.py
```diff
@@ -239,17 +222,25 @@ def save(
     debug_port
         Port number to use for debugging. By default no debugging port is opened.
     encoding
-        The encoding used when writing temporary files.
+        The character encoding used for the HTML content.
     _debug_dump
         Whether the saved image should be a big browser window, with key elements outlined. This is
         helpful for debugging this function's resizing, cropping heuristics. This is an internal
         parameter and subject to change.
+    **params
```
machow, line 230: Is it okay if we remove the `**params` piece of this PR? I'm onboard with everything else. Because `**params` only applies to when we go from a `.png` -> something else, it exposes PIL as part of the user API.

If we leave it out, we can swap out PIL down the road. As an alternative could we tell users that we're using Pillow? We could even tell them about PIL open and save if needed (so they can go from .png -> anything PIL supports on their own if more customization is needed).

machow (changes requested):
This PR is so helpful! Thanks for taking the time to work on .save(), which afaict is one of the most important parts of our API. My one request was that we remove `**params` so that PIL is a internal detail.

Excited to merge all these PRs, sorry for slowing your roll 😭

great_tables/_export.py
jrycw: Thanks for the explanation. I agree that we should remove `**params`, making `Pillow` an internal dependency rather than a public one. 

Regarding hints, I believe that diligent readers likely already recognize we use `Pillow` [under the hood](https://posit-dev.github.io/great-tables/reference/GT.save.html#details) and would consult its documentation for further customization if needed. 

This aligns with our goal of providing a user-friendly way to save generated tables rather than focusing on creating highly customized table figures in various formats🤔.

rich-iannone (approved):
LGTM!

machow:
Shoot, I just noticed the change to passing a base64 encoded url directly to the headless browser. I think there could be some challenges with this approach (and url length limitations). In general, modern browsers seem to have large limits, but it's a tricky territory.

Opened an issue to track:

https://github.com/posit-dev/great-tables/issues/518

PR #505, opened 2024-11-08
fix: use polars.__version__ in version check

lukemanley:
# Summary

Polars has a few different installations options including:
* pip install polars
* pip install polars-u64-idx
* pip install polars-lts-cpu

The last one was not covered by the existing version check. This PR changes the version check to use `polars.__version__` which should work for all installation types. 

- Ref: #

# Checklist

### DIFF
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -4,7 +4,6 @@ import warnings
 import re
 
 from functools import singledispatch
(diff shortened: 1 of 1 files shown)


machow (approved):
afaict this is the right approach in this situation since in polars, many distribution names may correspond with the module `polars`. (and version() uses distribution name)

Issue #491, opened 2024-10-10
`cols_width()` not stretching the cell values


mthomas-ketchbrook:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

The `cols_width()` function does not appear to be stretching the cell text to the full width of the column.

## Reproducible example

- [x] Post a minimal reproducible example (MRE) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
  - [x] **Runnable**: post enough code and data so any onlooker can create the error on their own computer.
  - [x] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
  - [x] **Readable**: format your code according to the [Style Guide for Python Code](https://peps.python.org/pep-0008/).

```
from great_tables import GT, style, exibble

exibble_mini = exibble[["num", "char", "date", "datetime", "row"]].head(5)

exibble_mini['long_text'] = 'Here is a really long sentence that I want to ensure is being handled appropriately by the cols_width argument, but I cannot get it to work unfortunately. I appreciate you taking the time to read this issue.'

(
    GT(exibble_mini)
    .cols_hide(columns = 'datetime')
    .cols_width(
        cases={
            "num": "50px",
            "char": "50px",
            "date": "50px",
            "row": "50px",
            "long_text": "400px"
        }
    )
)
```

This returns a GT that looks like this:

![Image](https://github.com/user-attachments/assets/38fa8db6-cc86-4583-91d3-cd6dc3d47a47)

## Expected result

I would have expected the text in the cells in the `long_text` column to have stretched horizontally and wrap every 400px.  I would have expected the final output to look something more like this:

![Image](https://github.com/user-attachments/assets/bdcc2d52-e216-4b29-b173-2c722ecf7d59)

## Development environment

- Operating System: Linux 
- **great_tables** Version: 0.13.0

[...]

PR #509, opened 2024-11-14
Fix: `cols_width()` should exclude hidden columns

rich-iannone:
This fixes https://github.com/posit-dev/great-tables/issues/491, and the problem is that hidden columns are included in the `<colgroup>` tag when only visible columns should be included.

With the fix, the following table renders correctly (adapted from the issue's table):

```python

exibble_mini = exibble[["num", "char", "date", "datetime", "row"]].head(1)

exibble_mini["long_text"] = (
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
)


tbl = (
    GT(exibble_mini)
    .cols_hide(columns="datetime")
    .cols_width(
        cases={
            "num": "50px",
            "char": "50px",
            "date": "400px",
            "row": "50px",
            "long_text": "400px",
            "datetime": "20px"}
    )
    .tab_options(
        table_body_vlines_style="solid",
        table_body_vlines_width="3px",
        table_body_vlines_color="red"
    )
)

tbl
```

image

The colgroup here is now:

```
<colgroup>
  <col style="width:50px;"/>
  <col style="width:50px;"/>
  <col style="width:400px;"/>
  <col style="width:50px;"/>
  <col style="width:400px;"/>
</colgroup>
```

which matches the number of columns and has the correct widths for each of the columns. Including or excluding a column width definition for a hidden column in the GT code makes no difference to the final result.

There was also another bug addressed with column widths and it involves not writing a column width in `<colgroup>` for a stub column. This PR fixes that by retrieving the column width (if any) for a stub column and placing it at the beginning of the series in `<colgroup>`. Here's a revised version of the above code that uses the `"date"` column as the row label column in the stub:

```python
...

tbl = (
    GT(exibble_mini, rowname_col="date")
...
```

image

Here's the resulting `<colgroup>`:

```
<colgroup>
  <col style="width:400px;"/>
  <col style="width:50px;"/>
  <col style="width:50px;"/>
  <col style="width:50px;"/>
  <col style="width:400px;"/>
</colgroup>
```

Fixes: https://github.com/posit-dev/great-tables/issues/491

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -432,6 +432,16 @@ class Boxhead(_Sequence[ColInfo]):
     def _get_column_widths(self) -> list[str | None]:
         return [x.column_width for x in self._d if x.type == ColInfoTypeEnum.default]
 
+    # Get a column width for a specific var value
+    def _get_column_width_by_var(self, var: str) -> str | None:
+
+        column_width = [x.column_width for x in self._d if x.var == var]
+
+        # Convert the single column width value in the list to a string
+        column_width = str(column_width[0])
+
+        return column_width
+
     # Get a list of visible columns
     def _get_default_columns(self) -> list[ColInfo]:
         default_columns = [x for x in self._d if x.type == ColInfoTypeEnum.default]
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -428,9 +428,9 @@ class Boxhead(_Sequence[ColInfo]):
 
         return self.__class__(out_cols)
 
-    # Get a list of column widths
+    # Get a list of visible column widths
     def _get_column_widths(self) -> list[str | None]:
-        return [x.column_width for x in self._d]
+        return [x.column_width for x in self._d if x.type == ColInfoTypeEnum.default]
 
     # Get a list of visible columns
     def _get_default_columns(self) -> list[ColInfo]:
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -719,7 +719,7 @@ def _get_table_defs(data: GTData) -> dict[str, Any]:
 
     # Get all the widths for the columns as a list where None values mean that the width is
     # not set for that column
-    # TODO: ensure that the stub column is set first in the list
+    # TODO: ensure that the stub column (if present) is included and set first in the list
     widths = data._boxhead._get_column_widths()
 
     # If all of the widths are defined as px values for all columns, then ensure that the width
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -719,8 +719,16 @@ def _get_table_defs(data: GTData) -> dict[str, Any]:
 
     # Get all the widths for the columns as a list where None values mean that the width is
     # not set for that column
-    # TODO: ensure that the stub column (if present) is included and set first in the list
     widths = data._boxhead._get_column_widths()
+
+    # Get the stub column variable
+    stub_var = data._boxhead._get_stub_column()
+
+    # If there is a stub column, then get the width for that column and insert it at the beginning
+    if stub_var is not None:
+
+        stub_width = data._boxhead._get_column_width_by_var(var=stub_var.var)
+        widths.insert(0, stub_width)
 
     # If all of the widths are defined as px values for all columns, then ensure that the width
     # values are strictly respected as absolute width values (even if table width already set)


great_tables/_utils_render_html.py
```diff
@@ -719,9 +719,17 @@ def _get_table_defs(data: GTData) -> dict[str, Any]:
...
-    # TODO: ensure that the stub column is set first in the list
     widths = data._boxhead._get_column_widths()
 
+    # Get the stub column variable
+    stub_var = data._boxhead._get_stub_column()
+
+    # If there is a stub column, then get the width for that column and insert it at the beginning
+    if stub_var is not None:
+
+        stub_width = data._boxhead._get_column_width_by_var(var=stub_var.var)
```
machow, line 730: Could we add a method like `Boxhead.canonical_columns()` or `.final_columns()`, that returns a list[ColInfo] for all final visible columns (including stub, and group column if correct option is set), in the order they will be rendered in?

This way, Boxhead becomes the source of truth on the final columns. I realize there's some ambiguity between Stub and Boxhead columns. But it seems okay for the final columns across these two things to live in one place, and the Boxhead seems like it might be the best fit?

In this case, the code could just call something like...

```
columns = data._boxhead.canonical_columns(options)

# widths for all columns that will be represented as columns
# e.g. groupings are rows by default, so wouldn't be included,
# but if the option to make them a column is set, then they would
widths = [col.width for col in columns]
```

machow (changes requested):

great_tables/_gt_data.py
```diff
@@ -396,6 +396,47 @@ def reorder(self, vars: list[str]) -> Self:
...
+        # Remove hidden columns from the boxhead
+        visible_columns = [x for x in self._d if x.visible]
+
+        # If there is a row group column, place it at the start of the boxhead
+        stub_column = [x for x in self._d if x.type == ColInfoTypeEnum.stub]
+
+        if len(stub_column) != 0:
+            # Get index of stub column and use that to move it to the start of the boxhead list
+            stub_index = self._d.index(stub_column[0])
+            visible_columns = [self._d[stub_index]] + [
```
machow, line 414: Don't you already have `stub_column`, why is it being fetched again (as `self._d[stub_index]`)?

great_tables/_gt_data.py
```diff
@@ -396,6 +396,47 @@ def reorder(self, vars: list[str]) -> Self:
 
         return self[new_order]
 
+    # Get a mutated version of the Boxhead that:
+    # - removes hidden columns
+    # - places the stub column in the correct position
+    # - places the row group column in the correct position (if option is set for that)
+    def final_columns(self, options: Options) -> Self:
```
machow, line 403: What if rather than...

* get visible columns
* get stub
* if stub exists, remove from visible columns and create a new list of columns with stub at the front
* get group column
* if group column exists do something similar (shifting around columns)

You constructed the final column list directly, so something like:

* get stub column if it exists and append to list
* get row group column if it exists (and option dictates its a column) and append to list
* extend list with remaining columns (those marked ColInfoTypeEnum.default?)

This way the code reads in the same order as the final columns returned

great_tables/_utils_render_html.py
```diff
@@ -717,10 +717,11 @@ def _get_table_defs(data: GTData) -> dict[str, Any]:
     # Get the table's width (which or may not have been set)
     table_width = data._options.table_width.value
 
-    # Get all the widths for the columns as a list where None values mean that the width is
-    # not set for that column
-    # TODO: ensure that the stub column is set first in the list
-    widths = data._boxhead._get_column_widths()
+    # Get all the widths for the columns (this includes the stub column)
+    boxh_final = data._boxhead.final_columns(options=data._options)
```
machow, line 721: Can we return a list of columns for the result of `final_columns`? It seems likely this boxhead with columns dropped (rather than hidden) violates some assumption of our data model.

machow (changes requested):

great_tables/_utils_render_html.py
rich-iannone: Now done. Thanks!

great_tables/_gt_data.py
```diff
@@ -396,6 +396,16 @@ def reorder(self, vars: list[str]) -> Self:
...
 
+    def final_columns(self, options: Options) -> list[str]:
+        row_group_column = (
+            self._get_row_group_column() if options.row_group_as_column.value else None
+        )
+        stub_column = self._get_stub_column()
+        default_columns = self._get_default_columns()
+        combined = [row_group_column] + [stub_column] + default_columns
+
+        return [x.var for x in combined if x is not None]
```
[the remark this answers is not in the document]
rich-iannone: Good idea. Now implemented.

machow:
This looks great, thanks!

PR #510, opened 2024-11-16
Fix missing exception raise in `_val_is_numeric()` and `_val_is_str()`

jrycw:
Hello Team,  

While reviewing these two functions, I noticed that a `ValueError` exception is created but not raised.  
I suspect this might be a bug, but please feel free to correct me if this behavior is intentional.


machow (approved):
ah thanks, lgtm! @rich-iannone do you want to merge?

Issue #278, opened 2024-04-11
Add ability to format as accounting values (in `fmt_*()` methods where it makes sense)


rich-iannone:
We currently don't have the ability to format numeric values as accounting-type values (i.e., handling of negative values different than using the minus sign). This should ideally work inside:

- `fmt_number()`
- `fmt_integer()`
- `fmt_currency()`
- `fmt_percent()`

As for how it would look, this is a good start:

<img width="625" alt="accounting_values" src="https://github.com/posit-dev/great-tables/assets/5612024/28cfe1a1-0ed3-448c-b35e-9538ca1ef55d">

JulianCologne:
Great Idea! 😃 

Not sure if this is just to show the format but the "percent" **value** looks wrong 😄 .
Most programms (like excel, powerbi, ...) multiply by 100 and not divide by 10 😮 .

For example you perform a calculation resulting in `0.15` and formatted as percent yields `15%` (`1 = 100%`)

rich-iannone:
I rescaled the number before applying the percent formatting (to put it in a better range).

PR #513, opened 2024-11-20
Feat: accounting notation for `fmt_number()`, `fmt_percent()`, `fmt_integer()` and `fmt_currency()`

rich-iannone:
This introduces the `accounting=` arg to `fmt_number()`, `fmt_percent()`, `fmt_integer()` and `fmt_currency()`, allowing for accounting notation (e.g., `"-343.27"` -> `"(343.27)"`. This new functionality also extends to the same formatters in the `vals` module.

Here is an example table demonstrating usage:

```python
from great_tables import GT
import polars as pl

df = pl.DataFrame(
    {"number": [-1.2, 23.6], "percent": [-0.0523, 0.363], "integer": [2323, -23213], "currency": [-24334.23, 7323.253]}
).with_columns(number_acc = pl.col("number"), percent_acc = pl.col("percent"), integer_acc = pl.col("integer"), currency_acc = pl.col("currency"))

(
    GT(df)
    .fmt_number(columns="number")
    .fmt_percent(columns="percent")
    .fmt_integer(columns="integer")
    .fmt_currency(columns="currency")
    .fmt_number(columns="number_acc", accounting=True)
    .fmt_percent(columns="percent_acc", accounting=True)
    .fmt_integer(columns="integer_acc", accounting=True)
    .fmt_currency(columns="currency_acc", accounting=True)
    .tab_spanner(label="Default formatting", columns=[0, 1, 2, 3])
    .tab_spanner(label="Accounting notation", columns=[4, 5, 6, 7])
)
```

image

Fixes: #278

### DIFF
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -55,6 +55,7 @@ def val_fmt_number(
     drop_trailing_zeros: bool = False,
     drop_trailing_dec_mark: bool = True,
     use_seps: bool = True,
+    accounting: bool = False,
     scale_by: float = 1,
     compact: bool = False,
     pattern: str = "{x}",
@@ -108,6 +109,9 @@ def val_fmt_number(
         The `use_seps` option allows for the use of digit group separators. The type of digit group
         separator is set by `sep_mark` and overridden if a locale ID is provided to `locale`. This
         setting is `True` by default.
+    accounting
+        An option to use accounting style for values. Normally, negative values will be shown with a
+        minus sign but using accounting style will instead put any negative values in parentheses.
     scale_by
         All numeric values will be multiplied by the `scale_by` value before undergoing formatting.
         Since the `default` value is `1`, no values will be changed unless a different multiplier
@@ -151,6 +155,7 @@ def val_fmt_number(
         drop_trailing_zeros=drop_trailing_zeros,
         drop_trailing_dec_mark=drop_trailing_dec_mark,
         use_seps=use_seps,
+        accounting=accounting,
         scale_by=scale_by,
         compact=compact,
         pattern=pattern,
@@ -168,6 +173,7 @@ def val_fmt_number(
 def val_fmt_integer(
     x: X,
     use_seps: bool = True,
+    accounting: bool = False,
     scale_by: float = 1,
     compact: bool = False,
     pattern: str = "{x}",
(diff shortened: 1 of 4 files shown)


great_tables/_formats_vals.py
```diff
@@ -108,6 +109,9 @@ def val_fmt_number(
         The `use_seps` option allows for the use of digit group separators. The type of digit group
         separator is set by `sep_mark` and overridden if a locale ID is provided to `locale`. This
         setting is `True` by default.
+    accounting
+        An option to use accounting style for values. Normally, negative values will be shown with a
```
machow, line 113: WDYT of changing these docstrings to something like "Whether to use accounting style, which wraps negative numbers in parentheses instead of using a minus sign."

machow (approved):
I left one suggestion for a minor tweak to the docstring language, but feel free to take or leave it! (not the language is repeated in several places)

tests/__snapshots__/test_formats.ambr
```diff
@@ -9,6 +9,49 @@
...
+    <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="percent_acc">percent_acc</th>
+    <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="integer_acc">integer_acc</th>
+    <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="currency_acc">currency_acc</th>
+  </tr>
+  </thead>
+  <tbody class="gt_table_body">
+    <tr>
+      <td class="gt_row gt_right">−1.20</td>
+      <td class="gt_row gt_right">−5.23%</td>
+      <td class="gt_row gt_right">2,323</td>
```
machow, line 34: It seems like this is not testing that fmt_integer uses accounting format on negative numbers

machow (changes requested):

tests/__snapshots__/test_formats.ambr
rich-iannone: The first four columns of this table are formatted but with `accounting=False` (minus signs persist) , the last four have `accounting=True` so we do get the negative values in parens there.

great_tables/_formats_vals.py
rich-iannone: Sounds good, now done!

tests/test_formats.py
```diff
@@ -68,6 +68,37 @@ def test_format_snap(snapshot):
     assert_rendered_body(snapshot, new_gt)
 
 
+def test_format_accounting_snap(snapshot):
+
+    df = pl.DataFrame(
+        {
+            "number": [-1.2, 23.6],
+            "percent": [-0.0523, 0.363],
+            "integer": [2323, -23213],
```
rich-iannone, line 76: Ensure that negative numbers are in first row, positive in second.

machow (approved):
Once the integer column in the test is made consistent with the others (i.e. negative number first) this LGTM

PR #520, opened 2024-11-23
Enhance `GT.fmt_image()` to support http/https schema in the `columns` parameter

jrycw:
Related PRs: #444, #451.

Currently, `GT.fmt_image()` supports formatting images from URLs by combining the `columns` and `path` parameters, as shown below:  
```python
import polars as pl
from great_tables import GT, vals, html

img_path = "https://raw.githubusercontent.com/posit-dev/great-tables/refs/heads/main/great_tables/data/metro_images/"

df = pl.DataFrame({"lines": ["tram_T6.svg", "tram_T7.svg", "tram_T8.svg"]})
GT(df).fmt_image("lines", path=img_path)
```  
image

<hr>

This PR proposes an enhancement to allow image formatting using only the `columns` parameter, enabling the following code to execute without errors:  
```python
df2 = df.with_columns(pl.lit(img_path).add(pl.col("lines")).alias("lines"))
```  
image

```python
GT(df2).fmt_image("lines")  # It produces the same table as the original approach.
```
This modification should benefit users who retrieve a list of URLs through a database query and store them as a column in the DataFrame. A possible scenario can be illustrated with the following pseudo-code:  

```python
...
urls: list[str] = db.execute("SELECT url FROM Foo")
df = pl.DataFrame({"urls": urls})
GT(df).fmt_image("urls")
```

<hr>

Additionally, this change would make it more convenient for users who want to place images in locations other than the body using `vals.fmt_image()`. For example:  
```python
header_imgs = vals.fmt_image(f"{img_path}tram_T5.svg")
GT(df2).fmt_image("lines").tab_header(title=html(header_imgs[0]))
```  
image

<hr>

I believe this improvement will enhance the user experience and better align with the documentation, which states that [complete http/https URLs should be accepted](https://posit-dev.github.io/great-tables/reference/GT.fmt_image.html#great_tables.GT.fmt_image). That said, I'm open to any suggestions or feedback!

### DIFF
--- a/great_tables/_formats.py
+++ b/great_tables/_formats.py
@@ -28,7 +28,7 @@ from ._tbl_data import (
     _get_column_dtype,
 )
 from ._text import _md_html, escape_pattern_str_latex
-from ._utils import _str_detect, _str_replace
+from ._utils import is_valid_http_schema, _str_detect, _str_replace
 from ._utils_nanoplots import _generate_nanoplot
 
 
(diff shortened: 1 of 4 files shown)


jrycw:
By the way, the new `ruff` settings appear to be working smoothly, and it seems our project runs on Python 3.13 without any issues.

machow (approved):
This LGTM, @rich-iannone do you want to merge?

PR #521, opened 2024-11-24
Standardize imports by converting absolute imports to relative imports

jrycw:
This PR aims to ensure consistency by converting all absolute imports to relative imports across most of our modules, as the majority already use relative imports. However, for the `_data_color` folder, I felt it was more natural to retain absolute imports, so those have been left unchanged.


machow (approved):
This LGTM

PR #523, opened 2024-11-25
feat: add `fmt_flag()` method

rich-iannone:
This adds the `fmt_flag()` method, which allows for the rendering of flag icons based on 2- or 3-letter country codes in table cells. The flag SVG data has been added within CSV data so there are no new dependencies with this PR.

Here's an example of the `fmt_flag()` method in use:

```python
from great_tables import GT
from great_tables.data import countrypops
import polars as pl

countrypops_mini = (
    pl.from_pandas(countrypops)
    .filter(pl.col("year") == 2021)
    .filter(pl.col("country_name").str.starts_with("S"))
    .sort("country_name")
    .head(10)
    .drop(["year", "country_code_3"])
)

(
    GT(countrypops_mini)
    .fmt_integer(columns="population")
    .fmt_flag(columns="country_code_2")
    .cols_label(
        country_code_2 = "",
        country_name = "Country",
        population = "Population (2021)"
    )
    .cols_move_to_start(columns="country_code_2")
)
```

image


great_tables/data/x_country_names.csv
machow, line 1: This file is  1.74 MB, which is a bit large to carry around in a Python library. Is the purpose of this file so people can specify flag names in a variety of locales? I wonder if 1.74MB in order to support specification of flag names across locales is a bit heavy (could there be a separate library for translating country names in different locales?).

rich-iannone: Yeah, I think a separate library might be better for this extra feature. I'll slim this down.

machow (approved):
This LGTM. I skimmed through the code and looked at the docs, but didn't read through the rendered HTML too closely (if it's close to / based on gt, guessing it is solid!)

Issue #91, opened 2023-12-15
Enable polars users to easily access to package datasets


machow:
Currently, great tables includes over a dozen datasets in its `.data` submodule:

```python
# all datasets live in submodule
from great_tables.data import airquality, exibble, towny

# the exibble dataset can be fetched from the top-level
# which allows us to quickly churn out examples
from great_tables import exibble
```

However, these datasets are pandas DataFrames, so polars users need to convert them:

```python
import polars as pl

pl.from_dataframe(exibble)
```

This isn't too bad. But maybe it could be better? This issue will discuss various ways we could approach loading data for both pandas and polars users.

## Possible approaches

1. **Leave as is**. Polars folks use `pl.from_pandas()` to convert.
3. **Use options object to configure DataFrame constructor**.
   * E.g. `set_options(data_frame = pl.DataFrame)`.
   * E.g. `set_options(data_frame="polars")`.
4. **Use functions to fetch each dataset**. They could take a constructor argument.
   * E.g. `exibble(pl.DataFrame)`, or
   * E.g. `exibble("polars")`, or
   * E.g. `exibble()     # uses set_options() to get DataFrame`
5. **pandas and polars each gets its own data module**. 
   * E.g. `from great_tables.data.polars import airquality`, OR
   * E.g. `from great_tables.data_pl import airquality`, OR
   * E.g. `from some_data_package import airquality`

## Desirable outcomes

* Easy to perform
* Helpful DataFrame completions in IDE

### Easy to perform

For example, if data is simply imported and `set_options()` is used, then people will need to do some code in-between imports. This feels a cludgy.

Here's an example:

```python
from great_tables import set_options

set_options(data_frame = "polars")

from great_tables.data import airquality
```

At the same time, calling functions gets kind of annoying:

```python
from great_tables import GT, set_options
from great_tables.data import airquality

set_options(data_frame = "polars")

# annoying to call, but imports can be up top
# you can change the option and call airquality again
# to get data based on the current data_frame option
df_airquality = airquality()

# maybe the workaround is calling inside GT
GT(airquality())
```

### Helpful DataFrame completions in IDE

I'm not sure how to implement something like `set_options()` and a data fetcher like `airquality()`? Is there a way to type it, so tools like pyright know that when an option is set to a specific value, that airquality() returns a specific type of DataFrame?

```python
from great_tables import set_options
from great_tabels.data import airquality

set_options(data_frame = "polars")
airquality.<tab>                                           # shows polars methods
```

[...]

PR #525, opened 2024-11-28
Refactor the `data.__init__.py` module

jrycw:
Hello team,  

This PR aims to address the `Pandas` dependency in reading datasets by introducing a unified `Dataset` API. The proposed approach allows users to retrieve datasets in a user-specified dataframe format. For example:  
* To get a `Pandas` dataframe for the `sza` dataset, use `Dataset.sza.to_pandas()`.  
* To get a `Polars` dataframe for the same dataset, use `Dataset.sza.to_polars()`.  
* To get a `PyArrow` dataframe for the same dataset, use `Dataset.sza.to_pyarrow()` (implementation of `_convert_to_pyarrow()` and `to_pyarrow()` is needed to support this).  

This way, users can use autocomplete to select both the dataset and the desired dataframe type.

To facilitate the transition:  
* Each dataset name begins with an uppercase letter (e.g., `Sza`).  
* A lowercase variable (e.g., `sza`) is provided as a `Pandas` dataframe, created using `to_pandas()`.  

If we decide to completely remove `Pandas` as a dependency in the future, the following tasks will be required:  
* Remove the warning message.  
* Remove the lowercase variables representing `Pandas` dataframes, and rename the dataset classes to lowercase.  
* Update `__all__`.
* Update documentation and tests to reflect the new API.  
* Address code sections marked with `# remove pandas` for further cleanup.  

I’m confident there are other excellent approaches to tackle this issue, so please feel free to modify or reject this PR as needed.

### DIFF
--- a/great_tables/data/__init__.py
+++ b/great_tables/data/__init__.py
@@ -6,7 +6,6 @@ except ModuleNotFoundError:
         "This will change in the future. Most of great_tables can be used without pandas."
     )
 
-
 from typing import Any
 from pathlib import Path
 from importlib_resources import files
@@ -36,13 +35,15 @@ __all__ = (
 DATA_MOD = files("great_tables.data")
 
 
-class _DataFrameConverter:
-    """ """
-
-    filepath: Path
-    dtype: dict[str, Any]
-    # `dtype` should contain Python types and rely on the hook `_convert_to_*()` to transform
-    # these types into the appropriate formats for different DataFrame implementations.
+class _Dataset:
+    def __init__(self, filepath: Path, dtype: dict[str, Any], doc: str = ""):
+        """
+        `dtype` should contain Python types and rely on the hook `_convert_to_*()` to transform
+        these types into the appropriate formats for different DataFrame implementations.
+        """
+        self.filepath = filepath
+        self.dtype = dtype
+        self.__doc__ = doc
 
     @staticmethod
     def _convert_to_pandas(v: Any) -> str:
@@ -52,13 +53,12 @@ class _DataFrameConverter:
         typs: dict[Any, str] = {int: "Int64", float: "Float64", str: "object"}
         return typs.get(v, "object")
 
-    @classmethod
-    def to_pandas(cls):
+    def to_pandas(self):
         import pandas as pd
 
-        dtype: dict[str, str] = {k: cls._convert_to_pandas(v) for k, v in cls.dtype.items()}
-        df = pd.read_csv(cls.filepath, dtype=dtype)
-        df.__doc__ = cls.__doc__
+        dtype: dict[str, str] = {k: self._convert_to_pandas(v) for k, v in self.dtype.items()}
+        df = pd.read_csv(self.filepath, dtype=dtype)
+        df.__doc__ = self.__doc__
         return df
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


jrycw:
The `Dataset` API can be used as follows:  
```python
from great_tables import GT
from great_tables.data import Dataset

exibble = Dataset.exibble

# Pandas users
df_pd = exibble.to_pandas()
GT(df_pd).show()


# Polars users
df_pl = exibble.to_polars()
GT(df_pl).show()


# PyArrow users
...
```

great_tables/data/__init__.py
```diff
@@ -1,1355 +1,1488 @@
...
+        return df
+
+    @staticmethod
+    def _convert_to_pyarrow(v: Any): ...
+
+    @classmethod
+    def to_pyarrow(cls): ...
+
+
+class Countrypops(_DataFrameConverter):
```
machow, line 87: Currently, this approach works by having the class _DataFrameConverter...

* specifying a filepath class attribute, etc..
* subclassing it to define those attributes

One risk of this approach is that a leans a bit hard on inheritance (see composition over inheritance). WDYT of an approach where things like Countrypops a instances of `_DataFrameConverter`, or something similar?

In this case you could add init arguments to `_DataFrameConverter`, like...

```python
class _DataFrameConverter:
    def __init__(self, filepath, dtype):
        self.filepath = filepath
        self.dtype = dtype

Countrypops = _DataFrameConverter(
    DATA_MOD / "01-countrypops.csv",
    dtype = {
        "country_name": str,
        "country_code_2": str,
        "country_code_3": str,
        "year": int,
        "population": int,
    }    
)
"""A docstring"""
```

great_tables/data/__init__.py
```diff
@@ -1,1355 +1,1488 @@
...
-    "magnetic_dipole_uncert": "float64",
-    "electric_quadrupole": "float64",
-    "electric_quadrupole_uncert": "float64",
-    "atomic_mass": "float64",
-    "atomic_mass_uncert": "float64",
-    "mass_excess": "float64",
-    "mass_excess_uncert": "float64",
-}
+
+class _DataFrameConverter:
```
machow, line 39: In case it's useful, I ended up creating a tiny DataFrame like implementation in reactable-py, just so I could feed example data for demos

https://github.com/machow/reactable-py/blob/main/reactable/simpleframe.py

I think a nice advantage of your approach here though is that it doesn't read the csv until you use one of the `to_*()` methods!

machow (changes requested):
Hey, thanks for working on this! This is a huge step towards not depending on a specific DataFrame library for our examples!

One thing I wonder is whether it might help cut out boilerplate to use a style with a bit more composition, rather than class inheritance.

great_tables/data/__init__.py
jrycw: @machow , thank you for the excellent suggestion! Since we're moving towards favoring composition, renaming `_DataFrameConverter` to `_Dataset` seems more appropriate, and passing the docstrings as arguments aligns well with this approach. In addition, I've taken the liberty of implementing the `Pyarrow` conversion and added a `__repr__()` method, though I'm uncertain about its utility—feedback on this would be appreciated.

For the internal datasets, I recommend retaining the current parsing approach with `Pandas`. It appears to be more robust, and I've noticed that handling certain cases with `Polars`, such as `x-airquality.csv` and `x_locales.csv`, is a bit tricky. Since these datasets are used internally, we should have room to explore better solutions for these edge cases.

great_tables/data/__init__.py
jrycw: The code looks great! It seems you'll need to keep the data as an internal variable within `simpleframe` to enable `get` and `set` actions. However, for our dataset use case, we should be able to store just the `filepath` and `dtype` rather than holding the entire dataset in memory.

jrycw:
The CI failure seems unrelated to our codebase.
```
Run browser-actions/setup-edge@v1
Setup Edge stable
Attempting to download Edge stable...
Error: Artifact not found of Edge stable for platform linux amd6[4](https://github.com/posit-dev/great-tables/actions/runs/12219134318/job/34085425230#step:6:5)
```

machow:
@rich-iannone do you mind taking a look at this? In particular, would love to get your take on...

* What variable names should we have in `great_tables.data`. E.g. currently there is...
  - `data.Islands` for the `_Dataset` instance, which has `to_polars()`, `to_pandas()` methods
  - `data.islands` (note the lowercase `i`) for the `pd.DataFrame
* What should the default be now, and in the long term? (see below)

Here are some possible approaches we could take to data.

## Backwards compatible: keep pandas the default, provide new variables for `_Dataset` objects

* e.g. `data.Islands` and `data.islands` both importable
* I'd prefer we eventually just have 1 thing. We could always make a `_Dataset` instance run some `to_*()` method by default, when passed to `GT()`, if e.g. polars is installed, etc...

## Agnostic: shift to `_Dataset` objects (e.g. `Islands`)

* deprecate pandas-by-import:
  -  when users import variables like `data.islands`
  - give a warning that its deprecated
  - suggest user imports `data.Islands` and runs the `.to_pandas()` method

## Agnostic: shift to `_Dataset` and eventually rename (e.g. `Islands` -> `islands`)

* deprecate pandas-by-import, with alternative import path: when users import variables like `data.islands`
  -  when users import variables like `data.islands`
  - give a warning that its deprecated
  - suggest user imports `data.v1.islands`
  - provide `to_pandas()`, `to_polars()`, `to_...()`, on `data.islands` -- as a subclass of pandas or something

rich-iannone:
@machow out of the three options you provided, I prefer the third one: "shift to `_Dataset` and eventually rename (e.g. `Islands` -> `islands`)".

machow:
Noice -- another way to do the third thing could be something like this?:

* `data.islands`: current behavior, deprecate -- tell them to do v0 import.
* `data.v0.islands`: old behavior
* `data.v1.islands`: new behavior
* X months from now, `data.islands`: new behavior (and we are fully dataframe agnostic!)

jrycw:
Include the previously discussed issue #91 as a reference.

rich-iannone:
From pairing with @machow, we landed on using:

- a `SimpleFrame` for `data.dataset`
- providing the `to_polars()` and `to_pandas()` methods on `data.dataset` (e.g., `data.dataset.to_polars()`)

The release process would then be
- next release: ensure that datasets have `to_polars()` and `to_pandas()` methods
- breaking release: implement `SimpleFrame` for datasets (along with the above methods)

machow:
@jrycw if it's okay with you, I can take up this issue--since it's a good chance to generalize simple frame from reactable to here. Thanks for doing so much exploration on this -- the `.to_*()` methods are super helpful to see, in terms of what the different libraries need to work!

If it's okay, I'll close this and open something with simple frame vendored into Great Tables

jrycw:
@machow Sure! `SimpleFrame` seems like a great choice. I wonder if I would have gone for it had I known about this option earlier.

machow:
I think this PR as is made a lot of sense -- it pushed us to discuss where we wanted example data to be a year from now. So I definitely think it was the right choice in terms of getting things moving!

Issue #529, opened 2024-12-02
[pyOpenSci review] Add note about required use of Pandas for datasets


rich-iannone:
From the pyOpenSci review comment:

Also, in order to run through all the examples using from great_tables.data import islands you have to have pandas installed. The error ModuleNotFoundError: Currently, importing great_tables.data requires pandas. This will change in the future. Most of great_tables can be used without pandas. that comes up is great, but it would be better to have it clear in the documentation at https://posit-dev.github.io/great-tables/get-started/#a-basic-table-using-great-tables that it needs to be installed to use this example data set at this point to lower the barrier of entry for users.

So either remove this dependency or make note of it in the documentation.

https://github.com/pyOpenSci/software-submission/issues/202#issuecomment-2468592164

PR #549, opened 2024-12-03
docs: add text about Pandas req for use of datasets

rich-iannone:
This addresses a pyOpenSci reviewer comment. In the API reference section of the docs site, there is a note about a requirement for a Pandas installation when using any of the datasets from the `datasets` module.

Fixes: https://github.com/posit-dev/great-tables/issues/529


machow (approved):
LGTM!

Issue #531, opened 2024-12-02
[pyOpenSci review] Add a clear reference to CITATION in the README or DOCS


rich-iannone:
From the pyOpenSci comment:

Governance and Engagement is clear, but I did not find a clear reference to CITATION in the README or DOCS

https://github.com/pyOpenSci/software-submission/issues/202#issuecomment-2468592164

PR #551, opened 2024-12-03
docs: add CITATION.cff file, link to CITATION text in README

rich-iannone:
This addresses a pyOpenSci reviewer comment related to citation information. The recommendation was to include a CITATION file and link to it from the README.md file (or other docs). Here, we have added the CITATION.cff file at the top level. A *Citation* section has been added to README.md file along with the bibtex citation text.

Fixes: https://github.com/posit-dev/great-tables/issues/531

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,29 @@
+cff-version: 1.2.0
+message: 'If you wish to cite the "great-tables" package use:'
+type: software
+license: MIT
+title: "great-tables: Make awesome display tables using Python."
+version: 0.14.0
+abstract:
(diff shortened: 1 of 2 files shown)


machow (approved):
LGTM

jrycw:
@rich-iannone and @machow, I noticed that we need to manually update these files with every release, which can be a bit of a hassle. Would it be possible for the CI to handle these updates automatically for each new release?

rich-iannone:
@jrycw I was just thinking this the other day! This is definitely something we should do otherwise the citation info will inevitably get stale.

Issue #547, opened 2024-12-02
[pyOpenSci review] exclude TYPE_CHECKING statement from coverage report


rich-iannone:
From the pyOpenSci review:

In the code, there is some code using the if TYPE_CHECKING statement. This could be excluded from the coverage report since this is False at runtime.

https://github.com/pyOpenSci/software-submission/issues/202#issuecomment-2489584108

rich-iannone:
There's a possible global solution here: https://github.com/nedbat/coveragepy/issues/831#issuecomment-517778185

PR #556, opened 2024-12-10
chore: exclude `if TYPE_CHECKING:` lines from coverage reporting

rich-iannone:
This addresses a pyOpenSci reviewer comment related to test coverage reporting. The recommendation was to exclude `if TYPE_CHECKING:` lines from coverage reporting since the condition is always False at runtime. This is done here globally through the addition of a `.coveragerc` file.

Fixes: https://github.com/posit-dev/great-tables/issues/547

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,0 @@
-[report]
-exclude_lines =
-    if TYPE_CHECKING:
-    pragma: no cover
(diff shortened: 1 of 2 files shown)


rich-iannone:
I can verify that this does work:

image

PR #559, opened 2024-12-12
docs: add note about pandas req in Get Started

rich-iannone:
This adds a note about the requirement for Pandas in the Get Started guide.


machow (approved):
If this shows up as expected in the docs, then LGTM!

PR #562, opened 2024-12-13
docs: update image and text around datasets provided

rich-iannone:
This PR updates the accounting of datasets in the  README and docs site landing page.


rich-iannone:
From pairing, we decided to merge right away.

Issue #389, opened 2024-07-01
Value based styling for pivoted tables? 


igorcalabria:
## Question

What would you like to know?

Hi, not sure if I've missed something but is there an easy way to apply conditional styles (https://posit-dev.github.io/great-tables/get-started/basic-styling.html) to pivoted tables (ie. Dates as columns)? 

I know it's possible to pass a polars column selector in the location body, but I haven´t found a proper way to select rows per column automatically.

Something like this 
```python
 tab_style(
        style=style.text(color="white"),
        locations=loc.body(
            columns=cs.numeric(),
            rows=pl.col("each-column-from-the-selector").eq(0)
        )
    )
```

The alternative is applying styles on individual columns but that's a bit cumbersome when columns are dynamic

jrycw:
@igorcalabria If I understand correctly, your idea should work as expected.

```python
import polars as pl
from great_tables import GT, loc, style
from great_tables.data import gtcars
from polars import selectors as cs

(
    GT(pl.from_pandas(gtcars).head())
    .tab_style(
        style=style.text(color="red"),
        locations=loc.body(
            columns=cs.numeric(),
            rows=pl.col("year").eq(2017.0)
        )
    )
)
```
![image](https://github.com/posit-dev/great-tables/assets/67060418/187571b4-3d7c-4209-8a76-6dd16d623fd3)

jrycw:
Oh, I think I understand what you're looking for now. Maybe something like this:
```python
import polars as pl
from great_tables import GT, loc, style
from great_tables.data import gtcars
from polars import selectors as cs

df_mini = pl.from_pandas(gtcars).head()
gt = GT(df_mini)


for column in df_mini.select(cs.numeric()).columns:
    gt = gt.tab_style(
        style=style.text(color="red"),
        locations=loc.body(columns=column, rows=pl.col(column).gt(600)),
    )

gt
```
![image](https://github.com/posit-dev/great-tables/assets/67060418/02297ad3-fb6e-432a-ab2b-98403197957b)

igorcalabria:
Yeah that's pretty much it. A closer example is something like this 
```python
import polars as pl
from great_tables import GT, loc, style
from great_tables.data import gtcars
from polars import selectors as cs

df_mini = pl.from_pandas(gtcars).head().pivot(on="year", index=["trim", "mfr"], values="hp")
gt = GT(df_mini,rowname_col="trim", groupname_col="mfr")
[...]

PR #566, opened 2024-12-16
Support `mask=` argument in `LocBody`

jrycw:
Related Issue: #389.

Hello team,  

This is a rough idea for supporting the passing of `Polars` expressions to the `mask=` argument of `LobBody`.

The procedures are outlined as follows:  
1. Exclude the `stub` and `group` columns if necessary.  
2. Use `df.select()` to process the `Polars` expression and obtain a boolean-like matrix.  
3. Use two for-loops to iterate through the cells and select the desired cells based on whether their values are `True`.  

This idea is preliminary but should provide a foundation for further discussions.  

```python
import polars as pl
from great_tables import GT, loc, style
from polars import selectors as cs

df_mini = pl.from_pandas(gtcars).head().pivot(on="year", index=["trim", "mfr"], values="hp")
(
    GT(df_mini,rowname_col="trim", groupname_col="mfr")
    .tab_style(
        style=style.text(color="red"),
        locations=loc.body(mask=cs.numeric().gt(600))
    )
)
```
image

```python
print(df_mini.select(cs.numeric().gt(600)))
```
```
shape: (3, 4)
┌────────┬────────┬────────┬────────┐
│ 2017.0 ┆ 2015.0 ┆ 2014.0 ┆ 2016.0 │
│ ---    ┆ ---    ┆ ---    ┆ ---    │
│ bool   ┆ bool   ┆ bool   ┆ bool   │
╞════════╪════════╪════════╪════════╡
│ true   ┆ null   ┆ null   ┆ null   │
│ null   ┆ false  ┆ false  ┆ true   │
│ null   ┆ false  ┆ null   ┆ null   │
└────────┴────────┴────────┴────────┘
```

### DIFF
--- a/great_tables/_locations.py
+++ b/great_tables/_locations.py
@@ -831,30 +831,33 @@ def resolve_mask_i(
     excl_group: bool = True,
 ) -> list[tuple[int, int, str]]:
     """Return data for creating `CellPos`, based on expr"""
+    import polars as pl
+
     if not isinstance(expr, PlExpr):
         raise ValueError("Only Polars expressions can be passed to the `mask` argument.")
+
+    frame: PlDataFrame = data._tbl_data
+    frame_cols = frame.columns
 
     stub_var = data._boxhead.vars_from_type(ColInfoTypeEnum.stub)
     group_var = data._boxhead.vars_from_type(ColInfoTypeEnum.row_group)
     cols_excl = [*(stub_var if excl_stub else []), *(group_var if excl_group else [])]
 
-    frame: PlDataFrame = data._tbl_data
-    df = frame.select(expr)
+    masked = frame.select(expr)
+    for col_excl in cols_excl:
+        masked = masked.drop(col_excl, strict=False)
 
-    newer_row_attr, older_row_attr = "with_row_index", "with_row_count"
-    row_number_colname = "__row_number__"
-    row_attr = newer_row_attr if hasattr(df, newer_row_attr) else older_row_attr
+    # Validate that `masked.columns` exist in the `frame_cols`
+    assert set(masked.columns).issubset(set(frame_cols))
 
-    # Add row numbers after `df.select()`, as the `__row_number__` column type is `UInt32`.
-    df_with_row_number = getattr(df, row_attr)(name=row_number_colname)
-
-    select_columns = [col for col in df.columns if col not in cols_excl]
+    # Validate that row lengths are equal
+    assert masked.select(pl.len()).item(0, "len") == frame.select(pl.len()).item(0, "len")
 
     cellpos_data: list[tuple[int, int, str]] = []  # column, row, colname for `CellPos`
-    for row_dict in df_with_row_number.iter_rows(named=True):
-        for col_idx, colname in enumerate(select_columns):
-            if row_dict[colname]:  # select only when `row_dict[colname]` is True
-                row_idx = row_dict[row_number_colname]
+    for row_idx, row_dict in enumerate(masked.iter_rows(named=True)):
+        for colname, value in row_dict.items():
+            if value:  # select only when `value` is True
+                col_idx = frame_cols.index(colname)
                 cellpos_data.append((col_idx, row_idx, colname))
     return cellpos_data
 
--- a/great_tables/_locations.py
+++ b/great_tables/_locations.py
@@ -839,8 +839,6 @@ def resolve_mask_i(
     excl_group: bool = True,
 ) -> list[tuple[int, int, str]]:
     """Return data for creating `CellPos`, based on expr"""
-    import polars as pl
-
     if not isinstance(expr, PlExpr):
         raise ValueError("Only Polars expressions can be passed to the `mask` argument.")
 
@@ -861,7 +859,7 @@ def resolve_mask_i(
         raise ValueError("The `mask` may reference columns not in the original DataFrame.")
 
     # Validate that row lengths are equal
-    if not (masked.select(pl.len()).item(0, "len") == frame.select(pl.len()).item(0, "len")):
+    if masked.height != frame.height:
         raise ValueError("The DataFrame length after applying `mask` differs from the original.")
 
     cellpos_data: list[tuple[int, int, str]] = []  # column, row, colname for `CellPos`
--- a/great_tables/_locations.py
+++ b/great_tables/_locations.py
@@ -832,7 +832,7 @@ def resolve_rows_i(
     )
 
 
-def resolve_mask_i(
+def resolve_mask(
     data: GTData | list[str],
     expr: PlExpr,
     excl_stub: bool = True,
@@ -930,7 +930,7 @@ def _(loc: LocBody, data: GTData) -> list[CellPos]:
             CellPos(col[1], row[1], colname=col[0]) for col, row in itertools.product(cols, rows)
         ]
     else:
-        cellpos_data = resolve_mask_i(data=data, expr=loc.mask)
+        cellpos_data = resolve_mask(data=data, expr=loc.mask)
         cell_pos = [CellPos(*cellpos) for cellpos in cellpos_data]
     return cell_pos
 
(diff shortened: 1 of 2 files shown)


great_tables/_locations.py
```diff
@@ -823,6 +824,41 @@ def resolve_rows_i(
...
+    """Return data for creating `CellPos`, based on expr"""
+    if not isinstance(expr, PlExpr):
+        raise ValueError("Only Polars expressions can be passed to the `mask` argument.")
+
+    stub_var = data._boxhead.vars_from_type(ColInfoTypeEnum.stub)
+    group_var = data._boxhead.vars_from_type(ColInfoTypeEnum.row_group)
+    cols_excl = [*(stub_var if excl_stub else []), *(group_var if excl_group else [])]
+
+    frame: PlDataFrame = data._tbl_data
+    df = frame.select(expr)
```
machow, line 842: What if the rules for the mask were as follows?:

* the mask expressions should--when run by `frame.select(mask_expr)`--return a DataFrame that:
  - columns: has the same or fewer columns (names must match original frame, but could be in a different order).
  - rows: has the same number of rows
* after running the expression, we validate this right away
* the mask is assumed to be in the original order
* in the case we can capture the column name and row number of each cell (e.g. with `enumerate(mask_result.iter_rows())`).

jrycw: Thanks for summarizing the rules.  

It seems that for the last part, we only need a single loop to gather all the information required by `CellPos`.

jrycw: How about this version?  

For prototyping, use `assert` for validation, which can later be replaced with `raise ValueError()`.

jrycw: I've replaced `assert` with `raise ValueError()` and added more test cases.

machow:
@jrycw this is looking really good! Anything else needed for this PR? I'm happy to help with any pieces!

jrycw:
@machow One thing I’m unsure about is the naming of the function. Specifically, I’m not clear on what the `i` or `c` represents in `resolve_rows_i` and `resolve_cols_c`. Does `resolve_mask_i` make sense? Other than that, I think the function is ready for users to experiment with. The idea of `mask=` is great, and I believe it will offer a lot of flexibility to users.

rich-iannone:
> @machow One thing I’m unsure about is the naming of the function. Specifically, I’m not clear on what the `i` or `c` represents in `resolve_rows_i` and `resolve_cols_c`. Does `resolve_mask_i` make sense? Other than that, I think the feature is ready for users to experiment with. The idea of `mask=` is great, and I believe it will offer a lot of flexibility to users.

I can help!

`i` == index (int)
`c` == column (str)

this is admittedly a bit of a holdover from the R package (which also has the `resolve_cols_i()` and `resolve_rows_l()` functions). But based on this, maybe just `resolve_mask` is fine?

jrycw:
@rich-iannone Thanks for the explanation! I agree that `resolve_mask` seems like a better fit here.

machow:
From a quick glance it's looking really nice -- if it's ready for review I'm happy to take a closer look!

jrycw:
The PR should be ready for review; please proceed.

great_tables/_locations.py
```diff
@@ -823,6 +832,45 @@ def resolve_rows_i(
...
+
+    frame: PlDataFrame = data._tbl_data
+    frame_cols = frame.columns
+
+    stub_var = data._boxhead.vars_from_type(ColInfoTypeEnum.stub)
+    group_var = data._boxhead.vars_from_type(ColInfoTypeEnum.row_group)
+    cols_excl = [*(stub_var if excl_stub else []), *(group_var if excl_group else [])]
+
+    # `df.select()` raises `ColumnNotFoundError` if columns are missing from the original DataFrame.
+    masked = frame.select(expr)
```
machow, line 853: Could we replace the drop loop below with `masked = frame.select(expr).drop(cols_excl, strict=False)`? We may want to verify dropping like this doesn't error if `cols_excl` is an empty list.

great_tables/_locations.py
```diff
@@ -823,6 +832,45 @@ def resolve_rows_i(
...
+    for col_excl in cols_excl:
+        masked = masked.drop(col_excl, strict=False)
+
+    # Validate that `masked.columns` exist in the `frame_cols`
+    if not (set(masked.columns).issubset(set(frame_cols))):
+        raise ValueError("The `mask` may reference columns not in the original DataFrame.")
```
machow, line 859: Could we enumerate the extra columns in the error message? For example, something like this?:

```python
missing = set(masked.columns) - set(frame_cols)

raise ValueError(
    "The `mask` expression produces extra columns, with names not in the original DataFrame."
    f"\n\nExtra columns: {missing}"
)
```

great_tables/_locations.py
```diff
@@ -823,6 +832,45 @@ def resolve_rows_i(
...
+
+    # Validate that row lengths are equal
+    if masked.height != frame.height:
+        raise ValueError("The DataFrame length after applying `mask` differs from the original.")
```
machow, line 863: Could we say what the differences are here? Maybe similar to above...

```python
raise ValueError(
    "The DataFrame length after applying `mask` differs from the original."
    "\n\n* Original length: {frame.height}"
    "\n* Mask length: {masked.height}"
)
```

great_tables/_locations.py
```diff
@@ -823,6 +832,45 @@ def resolve_rows_i(
...
+
+    cellpos_data: list[tuple[int, int, str]] = []  # column, row, colname for `CellPos`
+    for row_idx, row_dict in enumerate(masked.iter_rows(named=True)):
+        for colname, value in row_dict.items():
+            if value:  # select only when `value` is True
+                col_idx = frame_cols.index(colname)
```
machow, line 869: Could we move this indexing to just outside the loop? (Probably not a big deal for computation time, but just in case there's a very big table). For example...

```python
col_idx_map = {frame_cols.index(colname) for colname in frame_cols.columns}

# for loop starts below...
# does col_idx = col_idx_map[colname]
```

great_tables/_locations.py
```diff
@@ -869,14 +917,21 @@ def _(loc: LocStub, data: GTData) -> set[int]:
...
-    cell_pos = [
-        CellPos(col[1], row[1], colname=col[0]) for col, row in itertools.product(cols, rows)
-    ]
+    if (loc.columns is not None or loc.rows is not None) and loc.mask is not None:
+        raise ValueError(
+            "Cannot specify the `mask` argument along with `columns` or `rows` in `loc.body()`."
+        )
 
+    if loc.mask is None:
+        rows = resolve_rows_i(data=data, expr=loc.rows)
```
machow, line 926: Can we move `cols = resolve_cols_i(...)` from line 920 to this if block? That way the guard that errors if columns and mask are both set triggers right away, and the cols piece only runs if needed.

(I could be missing an important reason for it to be at the top though)

machow (changes requested):
This is looking really great--thanks for taking the time to implement this! I left several small suggestions for some error wording, and tweaks. Always happy to pick things up from here if useful. Thanks again for all the work you put into this!

great_tables/_locations.py
jrycw: That's a good idea! I think I initially used a for-loop because I forgot that we could drop multiple columns at once.

By the way, during the modification, I noticed a couple of things:  
* If `cols_excl` is an empty list, the code might not enter this for-loop block.  
* `df.drop()` seems to handle an empty list without needing `strict=False`.

great_tables/_locations.py
jrycw: I got your idea and made some tweaks (`frame_cols` is a list, so no need to use `frame_cols.columns`, and `col_idx_map` should be a map instead of a set).

jrycw:
@machow, I truly appreciate your review—all your suggestions are spot on! Also, it’s great to see all green checks again.

machow (approved):
This looks really great, thanks! @rich-iannone do you want to merge?

Issue #571, opened 2024-12-20
uint dtypes not being classified as scalar values by _str_detect


e-miz:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
  - Contributing guidelines 404s
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.
  - I couldn't find any nanoplot issue related to this

## Description

Hi! I wanted to make a nanoplot using a column in a dataframe. I found that columns with `uint` datatypes incorrectly fail a check for being scalar datatypes, which breaks nanoplot functionality for them.

### Works

```python
single_vals_df = pl.DataFrame(
    {
        "bars": [1, 2, 3, 5],
    }
)
single_vals_df.select("bars").cast(pl.Int32).style.fmt_nanoplot(columns="bars", plot_type="bar")
```

### Does Not Work

```python
single_vals_df = pl.DataFrame(
    {
        "bars": [1, 2, 3, 5],
    }
)
single_vals_df.select("bars").cast(pl.UInt32).style.fmt_nanoplot(columns="bars", plot_type="bar")
```

This leads to an error `TypeError: 'NoneType' object is not iterable`

## Expected result

The example with a `UInt32` column should've been able to create nanoplot just like the working example. The values aren't any different, just the datatype.

## Development environment

- Operating System: WSL2 Ubuntu 22.04 LTS
- **great_tables** Version: 0.15.0

## Additional context

I think I isolated the cause to [this line](https://github.com/posit-dev/great-tables/blob/2a1143466166e1436a2e24ba0a3d70eb72d6802b/great_tables/_formats.py#L4604).

`_str_detect` uses `re.match`, which fails to match `"int"`  with `"uint"` because it checks for a match at the beginning. Replacing with `re.search` fixes the problem because it looks for the pattern anywhere in the target, instead of just the beginning. I don't know if this has repercussions elsewhere though.

P.S. Thanks for the cool library! 😄

jrycw:
@e-miz Thanks so much for reporting this issue and pinpointing the root cause!

PR #577, opened 2025-01-09
fix: support `pl.UInt` type in nanoplot

jrycw:
Fix: #571  

It seems the issue stems from the input being a type we didn’t anticipate. While more advanced types like `pl.Enum`, `pl.Categorical`, `pl.Array`, `pl.List` or `pl.Struct` could introduce additional complexity, this PR addresses the simpler part for now.  

Since `_str_detect()` is clearly an internal function, we could consider modifying it. However, if we switch to using `re.search()`, the following test cases will fail:  
```bash
- FAILED `tests/test_formats.py::test_fmt_nanoplot_polars_listcol` - assert [- snapshot] == [+ received]  
- FAILED `tests/test_utils_render_html.py::test_render_polars_list_col` - assert [- snapshot] == [+ received]  
```
If we opt not to modify `_str_detect()`, we need to handle `uint` as a numerical type, as implemented in this PR.  

Additionally, I noticed that the alignment of `pl.UInt` might require adjustment.


rich-iannone (approved):
LGTM, thanks!

PR #584, opened 2025-01-23
blog: add blog post promoting the `mask=` parameter in `loc.body()`

jrycw:
Hello team,

I'd like to propose a blog post to promote the new `mask=` parameter in `loc.body()`. Feedback is always welcome!

The post is relatively short, and perhaps the team could extract some content to incorporate into the release blog for `0.16.0`.

### DIFF
--- a/docs/blog/locbody-mask/index.qmd
+++ b/docs/blog/locbody-mask/index.qmd
@@ -1,7 +1,7 @@
 ---
 title: "Style Table Body with `mask=` in `loc.body()`"
 html-table-processing: none
-author: Rich Iannone, Michael Chow and Jerry Wu
+author: Jerry Wu
 date: 2025-01-24
 freeze: true
 jupyter: python3
@@ -26,7 +26,7 @@ from great_tables import GT, loc, style
 from great_tables.data import gtcars
 from polars import selectors as cs
 
-year_cols = ["2014.0", "2015.0", "2016.0", "2017.0"]
+year_cols = ["2014", "2015", "2016", "2017"]
 df_mini = (
     pl.from_pandas(gtcars)
     .filter(pl.col("mfr").is_in(["Ferrari", "Lamborghini", "BMW"]))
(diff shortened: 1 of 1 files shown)


jrycw:
By the way, has anyone noticed that we've hit 2000 stars? 🎉

machow (changes requested):
Yo, this article is so handy! WDYT of moving the 3rd section, where you demonstrate `mask=` up to be the top one (instead of for-loop). This way, people get the most desirable nice thing up front. I think one demo of a painful alternative would be okay, but wonder if two might start to drag a bit (@rich-iannone mentioned preferring the list to loc body demo vs the for loop).

What about this?

* mask example up top
* two mask example sections, so people get to see it shine (e.g. @rich-iannone suggested a second section showing you style nulls)
* a final section showing one painful alternative

In any event, mask is a killer feature -- so really appreciate you contributing it AND this post 😁

jrycw:
@rich-iannone and @machow, thanks for the prompt feedback. 

I ended up doing the following modifications:
* Moved the `mask=` example up, and used two calls to `GT.tab_style()` in this example to demonstrate how we can use both Polars selectors AND expressions based on user preferences. Additionally, I highlighted that dynamically identifying columns is particularly useful in pivoted operations.
* Kept the `list to loc.body()` approach and compared it with the first approach.

I hope these changes make sense to you. Please feel free to comment if something doesn't.

By the way, may I ask for the privilege to trigger preview docs (not the official deployment environment) so I can immediately check whether the style looks right?

machow (approved):
Yooo this is looking really nice. I noticed that the black fill for null values is quite strong. WDYT of making null value text color "grey" or "lightgrey"? That way, the effect is similar to fading out missing values, so they're less prominent.

image

Otherwise, the article looks really great -- thanks again for this, we'll release the new version in a sec, and are pumped to share the post!

No need to include us as authors -- you definitely did all the work here 😁

(We're looking into making previews easier for you to trigger)

rich-iannone:
We released `0.16.0` and, shortly after that, `0.16.1` to fix the incorrect dtype for the `year` column in `gtcars`: https://github.com/posit-dev/great-tables/releases/tag/v0.16.1. So if you update the package, you can change your `year_cols` line to this:

```python
year_cols = ["2014", "2015", "2016", "2017"]
```

and we'll get properly-formatted years as column names.

jrycw:
@machow and @rich-iannone, thank you for the feedback. I've updated the background color to `lightgrey` for improved readability and revised the column names accordingly.

image

rich-iannone:
Thanks so much for this post (and thank you for your patience with our suggestions). We'll likely merge this Monday morning and then do some announcements around this post and the `v0.16.0` release!

Issue #588, opened 2025-01-26
Documentation rendering issue


jrycw:
Hello team:

I've identified two issues in the documentation:  

* The `**kwargs` is not rendering properly. [For example](https://posit-dev.github.io/great-tables/reference/GT.cols_label.html):  

![Image](https://github.com/user-attachments/assets/7006d370-6c73-45c9-98d9-1f45113c9817)

* The callout block within the `Parameters block` is not displayed. For instance:  

   https://github.com/posit-dev/great-tables/blob/d9b4db68d063d031201f1ab7cf7de0b1813581a0/great_tables/_locations.py#L510-L512

   To fix this, it seems necessary to move the callout block before the `Parameters block`.

jrycw:
The second issue (callout block) has been resolved.

PR #589, opened 2025-01-27
docs: update the related docs about `mask=` parameter in `loc.body()`

jrycw:
Hello team,

This PR updates the documentation related to the `mask=` parameter in `loc.body()`. 

Additionally, it addresses a bug where error messages were not being rendered correctly.

### DIFF
--- a/docs/get-started/basic-styling.qmd
+++ b/docs/get-started/basic-styling.qmd
@@ -228,6 +228,17 @@ gt_pl_air.tab_style(
 )
 ```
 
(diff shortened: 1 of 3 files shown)


jrycw:
I suggest keeping the `mask=` experimental callout to address potential edge cases we might not have considered (related issue #588).

machow (approved):
This LGTM! @jrycw, we added you as a writer to this repo, so you should be able to open PRs from a clone of this repo itself (rather than a fork), and trigger doc build previews. Thanks again for all the incredible work you've done -- however we can make this easier, we're happy to figure out!

PR #590, opened 2025-01-27
ci: use deploy-pypi environment

machow:
This PR moves our pypi-token into a github environment, so we can require review before a github release pushes to pypi.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*  @rich-iannone @machow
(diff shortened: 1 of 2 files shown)


machow:
Seems to be working as expected:

image

rich-iannone (approved):
LGTM

machow:
Alright, I declined the release review for deploying to pypi, which works well. We should have everything into place so that...

* a CODEOWNER review is required to merge to main
* @machow  or @rich-iannone need to review triggered pypi deploys (in workflows)
* people can be given write access, so they can easily preview PR docs / CI

Issue #593, opened 2025-01-30
`GT.save()` saves only to `.png` file


SamuMazzi:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

In the documentation is stated
> file
        The name of the file to save the image to. Accepts names ending with .png, .bmp, and other
        image extensions. Also accepts the extension .pdf.

but actually right now any format that will be insterted when saving the table to file will be replaced by `.png` extension. This is probably related to #499 which edited these lines:

```py
  -  if not Path(file).suffix:
  -     file += ".png"
  +  file = str(Path(file).with_suffix(".png"))
```
with the problem that `.with_suffix` always put the desired extension.

I can also create a PR. It seems a pretty straightforward fix, but I'd like to have a confirmation about this bug, thanks.

## Reproducible example

```py
from great_tables import GT
from great_tables.data import sp500

# Define the start and end dates for the data range
start_date = "2010-06-07"
end_date = "2010-06-14"

# Filter sp500 using Pandas to dates between `start_date` and `end_date`
sp500_mini = sp500[(sp500["date"] >= start_date) & (sp500["date"] <= end_date)]

# Create a gt table based on the `sp500_mini` table data
table = (
    GT(sp500_mini)
    .tab_header(title="S&P 500", subtitle=f"{start_date} to {end_date}")
    .fmt_currency(columns=["open", "high", "low", "close"])
    .fmt_date(columns="date", date_style="wd_m_day_year")
    .fmt_number(columns="volume", compact=True)
    .cols_hide(columns="adj_close")
)

table.save("sp500_mini.pdf")
```
You'll see a file `sp500_mini.png`

## Expected result

It should have saved a file named `sp500_mini.pdf`

## Development environment

- Operating System: Linux, Ubuntu 24.04
- **great_tables** Version: 0.16.1

jrycw:
@SamuMazzi , thanks so much for reporting this.  

For now, at least two bugs need to be fixed:  
* The suffix bug mentioned in this issue.  
* Reverting to saving the file using a temporary file to feed into the browser, as suggested in #518.  

[...]

PR #599, opened 2025-02-04
fix: do not always save screenshot as png

machow:
Addresses #593, but only writing png suffix if there is no suffix on file.


rich-iannone (approved):
LGTM!

Issue #582, opened 2025-01-16
Add the `tab_spanner_delim()` method


rich-iannone:
From a [recent Discussion topic](https://github.com/posit-dev/great-tables/discussions/579) it would be a good idea to have the `tab_spanner_delim()` method. The [R implementation](https://gt.rstudio.com/reference/tab_spanner_delim.html) has quite a few reasonable use cases illustrated in the examples.

rich-iannone:
Since https://github.com/posit-dev/great-tables/pull/647 is merged this issue can be closed.

PR #604, opened 2025-02-12
feat: add `tab_spanner_delim()` method

jrycw:
Fix issue: #582.

Hello team:

This PR sets up a scaffold for `tab_spanner_delim()`.  

### Basic Logic:
1. Use `str.split` or `str.rsplit` to split column names, corresponding to `delimi=` and `limit=` parameters.  
2. Iterate through selected columns, splitting each name and recording its `level` and `root_col` (original column).  
3. Process the recorded info:  
   - If `level != 0`, call `tab_spanner()`.  
   - If `level == 0`, call `cols_label()`.  

### Example:  
```python
import polars as pl
from great_tables import GT

data = {
    "province.NL_ZH.pop": [1, 2, 3],
    "province.NL_ZH.gdp": [4, 5, 6],
    "province.NL_NH.pop": [7, 8, 9],
    "province.NL_NH.gdp": [10, 11, 12],
}

gt = GT(pl.DataFrame(data))
gt.tab_spanner_delim()
```
image

```python
gt.tab_spanner_delim(limit=1)
```
image

```python
gt.tab_spanner_delim(split="first", limit=1)
```
image

### Remaining Tasks:
* [ ] The `reverse=` logic is not implemented—need the team's help to clarify the expected behavior. 
* [ ] Docs are ported from `{gt}` but need refinement.  
* [ ] Tests are missing since the logic isn't fully settled yet.

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -110,6 +110,7 @@ quartodoc:
       contents:
         - GT.tab_header
         - GT.tab_spanner
+        - GT.tab_spanner_delim
         - GT.tab_stub
         - GT.tab_stubhead
         - GT.tab_source_note
--- a/great_tables/_spanners.py
+++ b/great_tables/_spanners.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import copy
 import itertools
-from typing import TYPE_CHECKING, TypedDict
+from typing import Literal, TYPE_CHECKING
 
 from typing_extensions import TypeAlias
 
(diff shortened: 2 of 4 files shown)


rich-iannone:
@jrycw Thank you so much for your work on this feature, it is a big one! Regarding the `reverse=` argument, I can explain its intent and how it could be implemented.

In your example you had:

```python
import polars as pl
from great_tables import GT

data = {
    "province.NL_ZH.pop": [1, 2, 3],
    "province.NL_ZH.gdp": [4, 5, 6],
    "province.NL_NH.pop": [7, 8, 9],
    "province.NL_NH.gdp": [10, 11, 12],
}

gt = GT(pl.DataFrame(data))
gt.tab_spanner_delim()
```

image

Which is a sensible way to structure column names as we go from large to small elements, from left to right (e.g., `province`, `NL_ZH`, `pop` in the first column). Some tables may have this the other way around, like this:

```python
data = {
    "pop.NL_ZH.province": [1, 2, 3],
    "gdp.NL_ZH.province": [4, 5, 6],
    "pop.NL_NH.province": [7, 8, 9],
    "gdp.NL_NH.province": [10, 11, 12],
}
```

It's just another way of structuring the elements, but in a different order (going from smallest to largest elements). Rendering that currently with `tab_spanner_delim()` returns this table:

image

But having the option to use `reverse=True` would give you the table at the top. And it might be more realistic to have a DataFrame like this:

```python
data = {
    "pop.NL_ZH": [1, 2, 3],
    "gdp.NL_ZH": [4, 5, 6],
    "pop.NL_NH": [7, 8, 9],
    "gdp.NL_NH": [10, 11, 12],
}
```

where it makes sense to the user that `pop` or `gdp` might naturally go before `NL_ZH` or `NL_NH`, so having `reverse=` is useful there.

In terms of implementation, what's done in the R program is just this:

1. split the column name by delimiter into a list
2. reverse the contents of the list if `reverse=True`
3. continue...

Hope this helps clarify the intended use of `reverse=`!

jrycw:
@machow and @rich-iannone, thanks for all the examples and explanations! I've tried to consolidate the logic you described into a class named `SpannerTransformer` (though there might be a better name). Now, the intermediate representation can be obtained from `SpannerTransformer.split()`, and the rectangle from `SpannerTransformer.get_rectangle()`.  

I've also added some tests, but I'm not entirely sure if my logic is correct—`reverse=` is still giving me a bit of trouble! :)

machow:
If it's okay -- I can pick this up and try to do the reverse piece. Rich walked me through the R gt tests a bit, and I think we might need to pair to double check some of the different behaviors this kind of method can produce

jrycw:
> If it's okay -- I can pick this up and try to do the reverse piece. Rich walked me through the R gt tests a bit, and I think we might need to pair to double check some of the different behaviors this kind of method can produce

It's awesome if you can pick it up.

Issue #594, opened 2025-01-31
Use of ID attributes


BenGale93:
## Question

I'm trying to create a document with multiple tables. Some of which have identical column names.
This results in `<th>` elements with duplicate `id` attributes. Is there a way we can turn ID generation off to avoid these collisions?

jrycw:
Thanks for bringing this up! Perhaps [GT.with_id()](https://posit-dev.github.io/great-tables/reference/GT.with_id.html#great_tables.GT.with_id) could help.

BenGale93:
Sorry, that doesn't work. As far as I can tell, that just sets the ID of the top level `<div>` that wraps the style and table.

From a bit of a dig around the source code, the issue seems to be https://github.com/posit-dev/great-tables/blob/main/great_tables/_utils_render_html.py#L183

Each `<th>` element that represents a column name is given an ID attribute with the column label.

This means, if you have multiple tables with column names in common, duplicate IDs are created.

For example:

`<th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col" id="Count">Count</th>`

jrycw:
Thanks for the reply! How about prefixing the table ID to each column? This seems to solve the issue.

BenGale93:
That would work. I'm happy to make a PR if that helps?

jrycw:
I believe a PR is welcome, but the team will make the final decision on whether to accept it.

BenGale93:
Of course!

jrycw:
@BenGale93, thanks so much for your issue and PR.

PR #607, opened 2025-02-12
fix!: ensure HTML ID attributes are unique

BenGale93:
# Summary

Previously, if you had multiple GT HTML tables inside a single document, that had column names in common, you would have repeated HTML ID attributes. This PR makes these ID attributes unique by prepending the table ID, if it exists.

- Ref: #594 

# Checklist

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -9,7 +9,7 @@ from . import _locations as loc
 from ._gt_data import GroupRowInfo, GTData, Styles
 from ._spanners import spanners_print_matrix
 from ._tbl_data import _get_cell, cast_frame_to_string, replace_null_frame
-from ._text import _process_text, _process_text_id
+from ._text import BaseText, _process_text, _process_text_id
 from ._utils import heading_has_subtitle, heading_has_title, seq_groups
 
 
@@ -39,6 +39,14 @@ def _flatten_styles(styles: Styles, wrap: bool = False) -> str | None:
     # if not wrapping the styles for html element,
     # return None so htmltools omits a style attribute
     return None
+
+
+def _create_element_id(table_id: str | None, element_id: str | BaseText | None) -> str:
+    # Given a table ID, element IDs are prepended by it to ensure the resulting HTML
+    # has unique IDs.
+    new_table_id = table_id or ""
+    processed_id = _process_text_id(element_id)
+    return f"{new_table_id}-{processed_id}" if new_table_id and processed_id else processed_id
 
 
 def create_heading_component_h(data: GTData) -> str:
@@ -150,6 +158,9 @@ def create_columns_component_h(data: GTData) -> str:
     # Initialize the column headings list
     table_col_headings = []
 
+    # Extract the table ID to ensure subsequent IDs are unique
+    table_id = data._options.table_id.value
+
     # If there are no spanners, then we have to create the cells for the stubhead label
     # (if present) and for the column headings
     if spanner_row_count == 0:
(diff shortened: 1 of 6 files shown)


rich-iannone:
This looks great, thanks! However, it looks like the formatting used for the new ID values shouldn't have a space. According to a [W3Schools page](https://www.w3schools.com/html/html_id.asp#:~:text=Note%3A%20The%20id%20name%20must,%2C%20tabs%2C%20etc.):

> The id name must contain at least one character, cannot start with a number, and must not contain whitespaces (spaces, tabs, etc.).

Could you revise to use a hyphen instead?

BenGale93:
Just updated it. I also applied the same rule to the column label and updated the test to have a column with a space in the name to demonstrate that.

great_tables/_utils_render_html.py
```diff
@@ -180,7 +191,7 @@ def create_columns_component_h(data: GTData) -> str:
                     colspan=1,
                     style=_flatten_styles(styles_column_labels + styles_i),
                     scope="col",
-                    id=_process_text_id(info.column_label),
+                    id=_create_element_id(table_id, info.column_label),
```
rich-iannone, line 194: We can easily have duplicate column labels (e.g., using `cols_label()`) but it's much less likely to have duplicate column names. So I suggest using the column name instead of the label.   

```suggestion
                    id=_create_element_id(table_id, info.var),
```

great_tables/_utils_render_html.py
```diff
@@ -225,7 +236,7 @@ def create_columns_component_h(data: GTData) -> str:
                     colspan=len(stub_layout),
                     style=_flatten_styles(styles_stubhead),
                     scope="colgroup" if len(stub_layout) > 1 else "col",
-                    id=_process_text_id(stub_label),
+                    id=_create_element_id(table_id, stub_label),
```
rich-iannone, line 239: Just noting for the future (nothing needs to be done here) but the stub doesn't really have an ID value (I suppose the column that created the stub labels could be it?) and it often won't have a label unless `tab_stubhead()` was called.

The issue with all this is that for the stubhead `<th>` element we get `id=""`, which is invalid HTML according to https://validator.w3.org/nu/.

great_tables/_utils_render_html.py
```diff
@@ -318,7 +329,7 @@ def create_columns_component_h(data: GTData) -> str:
                         colspan=1,
                         style=_flatten_styles(styles_column_labels + styles_i),
                         scope="col",
-                        id=_process_text_id(remaining_headings_labels[j]),
+                        id=_create_element_id(table_id, remaining_headings_labels[j]),
```
rich-iannone, line 332: To complete the move to using column IDs instead of labels for the element ID, I think you need to do this:

```suggestion
                        id=_create_element_id(table_id, remaining_headings_ids[j]),
```

To make this work, you'll need to add:

```python
remaining_heading_ids = [entry.var for entry in boxhead if entry.var in remaining_headings]
```

after the definition of remaining_headings_labels on line 305.

rich-iannone (changes requested):
This is so close, but I have a few suggested changes.

rich-iannone:
@BenGale93 Sorry to reverse course here a bit, but I did notice that if labels are used to compose element IDs, there's a good chance one might get duplicate ID values (since labels can be user-modified but column names are immutable and usually always unique). Could you address these changes? I tried out all of the suggested changes locally and tests pass and the element IDs look reasonable to me.

BenGale93:
@rich-iannone makes total sense. Will work on this in a few hours time. Thanks for the review!

rich-iannone (approved):
LGTM!

Issue #596, opened 2025-02-03
Issue when attempting to save


HansMellman:
## Code

```
(
    GT(qb)
    .tab_header(
        title=md("<img src='https://i.imgur.com/xJYzCF8.png' width='90' height='90' style='vertical-align: middle;'> &#x1F3C8; **Analysis by Position Group** &#x1F3C8; <img src='https://i.imgur.com/xJYzCF8.png' width='90' height='90' style='vertical-align: middle;'>"), ## add small footballs to the title. ## these characters cause issues when saving. 
        subtitle=md("_**2065 Regular Season**_")
    )
    .fmt_image(
        columns='Team',
        path=analysis_path,
        file_pattern="{}.png"
    )
    .cols_align(
        align="center", columns=None
    )
    .cols_label(
        Team = (" "),
        Position_Group = (" "),
        Intelligence = ("Intel"),
        Performance_Grade = ("Performance"),
        Current_Overall = ("Current"),
        Future_Overall = ("Future"),
        PenaltiesCommitted = ("Penalties"),
        Pedigree_Percentage = ("Pedigree"),
        Intelligence_vs_league_avg = ("Int"),
        PenaltiesCommitted_vs_league_avg = ("Pen"),
        Performance_Grade_vs_league_avg = ("Perf"),
        Current_Overall_vs_league_avg = ("Curr"),
        Future_Overall_vs_league_avg = ("Fut"),
        Experience_vs_league_avg = ("Exp"),
        Pedigree_Percentage_vs_league_avg = ("Pedigree")
    )
    .tab_options(
        container_width = "100%",
        heading_background_color="#990033",
        table_border_top_color="#990033",
        table_font_names="Calibri",
        table_font_size="20px"
    )
    .tab_style(
        style=style.fill(
            color=map_colour_analysis),
        locations=loc.body("Position_Group")
        )
    .tab_style(
        style=style.text(
            color=text_colour_analysis),
        locations=loc.body("Position_Group")
        )
    .tab_spanner(
        label=md('**Cumulative**'),
        columns=['PenaltiesCommitted', 'Pedigree_Percentage']
    )
    .tab_spanner(
        label=md('**Group Averages**'),
        columns=['Intelligence', 'Performance_Grade', 'Current_Overall', 'Future_Overall', 'Experience']
    )
    .tab_spanner(
[...]

Issue #518, opened 2024-11-22
save: investigate use of base64 encoded urls in headless browser


machow:
In PR https://github.com/posit-dev/great-tables/pull/499 we replaced saving to a temporary file with inputting base64 encoded rendered html into a headless browser.

Browser url limits are a bit tricky, but they do exist. Let's investigate potential issues we might hit by passing long urls to the headless browser.

Alternatively, it might save us time to revert back to saving a temporary file, since the gt R package has used that approach for ~6 years with little issue.

I would vote for reverting, just to align with gt R. It's likely modern browsers have large url limits, but it's tricky to verify.

jrycw:
@machow , I wasn’t aware of this issue. Please feel free to revert to the temporary file approach if needed.

phimad:
Related to this, I am having problems with special character encodings which I believe is due to the base64 encoding.

Below I provide a MRE. I have a table with special Danish characters such as Ø, Æ, Å. It shows up correctly with GT.show method, but special characters are lost when using save()-method.

import pandas as pd
from great_tables import GT

df = pd.DataFrame(
    {
        "Stubhead with Å": [
            "Rowlabel with Ø",
            "Rowlabel with Æ",
        ],
        "2023": [
            3.2,
            4.3,
        ],
        "2024": [
            5.7,
            7.8,
        ],
        "2025": [
            8.9,
            5.4,
        ],
    }
)

gt_tbl = GT(df)

gt_tbl.save("test.png")


This outputs the attached image:

![Image](https://github.com/user-attachments/assets/a7d8b187-d839-414d-b3f6-e6a9d36ffecb)

which does not show the special characters correctly. Seems it would be prudent to use the same approach for the show()-method as for the save()-method, but I am not an expert on this.

Am hoping this will be fixed soon as I really like this library :)

jrycw:
@phimad, thanks so much for reporting this issue. I’ve reformatted your code for better readability:  
```python
import pandas as pd
from great_tables import GT

df = pd.DataFrame(
    {
        "Stubhead with Å": [
            "Rowlabel with Ø",
            "Rowlabel with Æ",
        ],
        "2023": [
            3.2,
            4.3,
        ],
        "2024": [
            5.7,
            7.8,
        ],
        "2025": [
            8.9,
            5.4,
        ],
    }
)

gt_tbl = GT(df)

gt_tbl.save("test.png")
```
[...]

Issue #504, opened 2024-11-07
.save gives a very large picture with table only taking up a small fraction


thomasfrederikhoeck:
## Prework

- [ ] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [ ] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

.save gives a very large picture with table only taking up a small fraction
## Reproducible example

- [ ] Post a minimal reproducible example (MRE) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
  - [ ] **Runnable**: post enough code and data so any onlooker can create the error on their own computer.
  - [ ] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
  - [ ] **Readable**: format your code according to the [Style Guide for Python Code](https://peps.python.org/pep-0008/).

When I run the following:
```python
import polars as pl
from great_tables import GT

GT(pl.DataFrame({"My very nice column":[1,2,4],"My other very nice column": [3,4,5]})).save("plot.png")
```
it produces this image (notice all the whitespace).

![Image](https://github.com/user-attachments/assets/8988f8d7-f4d7-4889-9334-33ac2a464305)

## Expected result

I would have exected the resulting image to be "zoomed" in (no whitespace).

## Development environment

- Operating System: [e.g., Linux, macOS, Windows]
- **great_tables** Version: 0.1.13
- Windows
- selenium 4.26.1

machow:
Shoot, sorry this isn't capturing just the table. What browser and browser version are you using? There's some funkiness with how browsers reckon the size of the table that we try to work around :/.

ritviksahajpal:
I have the same issue, I am using the latest Chrome version (130.0.6723.117). Related, is there a way to set the DPI of the image saved when using GT.save()? thanks for a great package btw!

ritviksahajpal:
Getting the same error with edge as well

thomasfrederikhoeck:
@machow weirdly , I don't see the error anymore :-) I'm using Chrome 130.0.6723.117  (64-bit) on Windows.

HagaiHargil:
[...]

Issue #480, opened 2024-10-02
GT.save cuts off output


lwanger:
## Description

Bottom row of my table is cut-off in the output using GT.save. Same result with both pdf and png output files.

## Reproducible example

from pathlib import Path
import pandas as pd

data = {'A': [1,2,3], 'B': ['foo', 'bar', 'baz'], }
df = pd.DataFrame(data)
tbl = (gt.GT(df))

out_file = str(Path("table_export.pdf"))
tbl.save(file=out_file, web_driver='edge')

## Expected result

Bottom row should be shown in full.

## Development environment

- Operating System: Windows
- **great_tables** Version: 0.10
 
[table_export.pdf](https://github.com/user-attachments/files/17234041/table_export.pdf)

![Image](https://github.com/user-attachments/assets/3fd1f69e-8fca-4459-b1bb-910689919487)

HansMellman:
I've had this happen countless times, only anecdotal solution was to ensure that I wasn't throttling my internet connection with anything else at the same time - it was almost like the image was unable to render entirely before the screenshot was taken. I'm speaking as a layperson though because I have no idea if that's actually likely to affect the process or not - just my experience.

yk2684:
Yeah, this also happens to me time to time. My current solution is to use the window_size parameter and keep increasing the size until this no longer occurs. It also helps to run the code multiple times.

HansMellman:
> Yeah, this also happens to me time to time. My current solution is to use the window_size parameter and keep increasing the size until this no longer occurs. It also helps to run the code multiple times.

Rich advised me in the Discord to test using Firefox instead of Chrome and I have not actually had this issue since switching over - which browser were you using in your .save?

lwanger:
I generally use Microsoft Edge on Windows (which is built on the same engine as Chrome)

HansMellman:
> I generally use Microsoft Edge on Windows (which is built on the same engine as Chrome)

Definitely worth giving it a go with Firefox/GeckoDriver, if this issue is persistent for you.

PR #609, opened 2025-02-16
fix: resolve encoding issue in `GT.save()`

jrycw:
Related issues: #518, #596.  

The `base64` encoding does not seem to be a suitable choice for `GT.save()`, as it has caused multiple issues. Therefore, I suggest rolling back to the previous working approach.

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -431,8 +431,6 @@ def save(
     ```
 
     """
-    import base64
-
     # Import the required packages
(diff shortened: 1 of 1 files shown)


jrycw:
By the way, we still need to address the cut-off issue (#480 and #504). Maybe we should explore alternatives like `Matplotlib`?

machow (approved):
This LGTM, thanks for taking the time to work on this

rich-iannone (approved):
LGTM!

PR #618, opened 2025-02-20
docs: add examples to the documentation

jrycw:
Hello team,  

This PR improves our documentation by adding missing examples, primarily for the `vals` module. The goal is to provide simple code snippets that users can easily copy and run.

### DIFF
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -143,6 +143,14 @@ def val_fmt_number(
     -------
     list[str]
         A list of formatted values is returned.
+
+    Examples
+    --------
+    ```{python}
+    from great_tables import vals
+
+    vals.fmt_number([0.325, 777000], decimals=2)
+    ```
     """
 
     gt_obj: GTData = _make_one_col_table(vals=x)
@@ -236,6 +244,14 @@ def val_fmt_integer(
     -------
     list[str]
         A list of formatted values is returned.
+
+    Examples
+    --------
+    ```{python}
+    from great_tables import vals
+
+    vals.fmt_integer([100000.1, 2000000000.2], use_seps=False)
+    ```
     """
 
     gt_obj: GTData = _make_one_col_table(vals=x)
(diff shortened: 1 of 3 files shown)


great_tables/_spanners.py
```diff
@@ -107,16 +107,39 @@ def tab_spanner(
     )
     ```
 
+    One cool feature of `tab_spanner()` is its support for multiple levels, allowing you to group
+    columns in various ways. For example, you can create three bottom spanners by specifying `level=0`
+    and a top spanner by setting `level=1`:
+
+    ```{python}
```
rich-iannone, line 114: In this example, you could leave out all uses of the `level=` arg and still get the same result (because of the 'stacking' implementation). Maybe a better example (after this one with no use of `level=`) is to use a `level=` when stacking works against you.

Perhaps some variation of this (could use a different dataset if preferred):

```python
colnames = ["model", "hp", "hp_rpm", "trq", "trq_rpm"]

gtcars_mini = gtcars[colnames].head(10)

(
    GT(gtcars_mini)
    .tab_spanner(
        label="hp",
        columns=["hp", "hp_rpm"],
    )
    .tab_spanner(
        label="performance",
        columns=["hp", "hp_rpm", "trq", "trq_rpm", ],
    )
    .tab_spanner(
        label="trq",
        columns=["trq", "trq_rpm"],
        level=0
    )
)
```

In the above, if that last `tab_spanner()` didn't have `level=0` it would be automatically in level `2`. Instead, because there is space in level `0` above columns `"trq"` and `"trq_rpm"`, it can be placed in level `0`.

great_tables/_formats_vals.py
```diff
@@ -1000,6 +1095,11 @@ def val_fmt_image(
     -------
     list[str]
         A list of formatted values is returned.
+
+    See Also
```
rich-iannone, line 1099: Nice cross reference!

rich-iannone (changes requested):
These are welcome additions to our documentation! My only ask is to refine the extra examples for `tab_spanner()` (suggestions provided).

great_tables/_spanners.py
jrycw: @rich-iannone , thanks for the great suggestion! It helps us introduce the stacked functionality and the `level=` parameter of `tab_spanner()` to users in a clearer way.

rich-iannone (approved):
LGTM!!

Issue #615, opened 2025-02-19
0.000123456 shows as '1.23E−40' instead of '1.23E−4'


rich-iannone:
### Discussed in https://github.com/posit-dev/great-tables/discussions/614

<div type='discussions-op-text'>

<sup>Originally posted by **marceloandrioni** February 19, 2025</sup>
Sorry if it's a very basic question, but it's my first time using great_tables.

I was trying to better format a very small number using:

```
from great_tables import vals

vals.fmt_scientific(0.000123456, n_sigfig=3, exp_style="E")
['1.23E−40']
```

But the exponent keeps showing an extra zero. Does someone can point me to the correct way of doing this?

Thank you.

</div>

This issue seems to stem from the use of `exp_style=` and values not using that arg are unaffected.

jrycw:
https://github.com/posit-dev/great-tables/blob/dc8219791fbd857c85aa5e463472c9b387d2aaba/great_tables/_formats.py#L841-L848

It looks like the error is caused by the padding `0`. Would using `str.rjust()` instead of `str.ljust()` help resolve the issue?

rich-iannone:
This is definitely the problem! I'll create a PR right now.

PR #622, opened 2025-02-25
fix: scientific notation zero padding when `exp_style=` used

rich-iannone:
This fixes a reported issue with scientific notation formatting, specifically when the `exp_style=` arg is used. Using that argument with a letter (typically `"E"` or `"e"`) results in incorrectly formatted values because zero-padding is performed in the wrong direction (i.e., adding zeroes to the right).

The solution is to use `rjust()` instead of `ljust()` when padding a string with zeros to a minimum width. Tests were added here as well.

Fixes: #615


machow:
```python
from great_tables import vals

vals.fmt_scientific(0.000123456, n_sigfig=3, exp_style="E")
```

* Output before: "2.3E−30"
* Output now: "2.3E-03"

tests/test_formats.py
```diff
@@ -1115,6 +1115,51 @@ def test_format_number_with_sep_dec_marks():
     assert x == ["12.345.678,12346", "1,00000", "0,00000", "\u2212" + "12.345.678,12346"]
 
 
+# ------------------------------------------------------------------------------
+# Tests of `fmt_scientific()`
+# ------------------------------------------------------------------------------
+
+FMT_SCIENTIFIC_CASES: list[tuple[dict[str, Any], list[str]]] = [
```
machow, line 1122: Do you mind making the cases each only a single input, and a single output? That way, readers can tell which outputs of each test case matter (right now it's clear the inputs cross with kwargs, so its not clear if each output is important, or there as a side effect of crossing).

(This would just require a x_in parameter in the test signature)

machow (changes requested):

machow (approved):

Issue #616, opened 2025-02-19
Docs: Nanoplots docs have an error in the build


phobson:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

I was reading the docs on nanoplots and noticed that jupyter threw an error during the build

## Reproducible example

As of 15:00 PST on 2025-02-19 the build contains the error:
https://posit-dev.github.io/great-tables/get-started/nanoplots.html

PR #623, opened 2025-02-25
fix: update conditional statement in nanoplot code to guard against string values (single string or list containing any strings)

rich-iannone:
This fixes a conditional statement in the nanoplot such that using `val_is_str()` is avoided, opting for a more precise definition of the conditional.

Before:

```python
# will return True if *any* list provided
if expand_x is not None and _val_is_str(expand_x):
```

After:

```python
# will return True if input is a scalar string or a list with any string items
if isinstance(expand_x, str) or (
    isinstance(expand_x, list) and any(isinstance(item, str) for item in expand_x)
): 
```

Returning `True` leads directly to a `NotImplementedError` which should not be raised for a list of numbers. Using `_val_is_str()` here returns `True` for any list input, so that should no longer be used.

This fixes broken documentation in https://posit-dev.github.io/great-tables/get-started/nanoplots.html (see the long stack trace stemming from the `ValueError`), where `expand_x` is used (correctly) in the final example. 

Fixes: https://github.com/posit-dev/great-tables/issues/616


great_tables/_utils_nanoplots.py
```diff
@@ -913,7 +913,9 @@ def _generate_nanoplot(
     # there are no x values, generate equally-spaced x values according
     # to the number of y values
     if plot_type == "line" and x_vals is not None:
-        if expand_x is not None and _val_is_str(expand_x):
+        if isinstance(expand_x, str) or (
```
machow, line 916: I don't understand this code. The type annotation for expand_x in this function and fmt_nanoplot does not include strings. Which of these functions allows strings? (and which should / shouldn't?)

machow (changes requested):

great_tables/_utils_nanoplots.py
machow: Shoot, I'm realizing that this is just a guard, which raises a error by design. I think the big issue is that this PR does not cause the failing example to succeed on the guide page

machow: Ah, the issue is that `_val_is_str` raises an error on lists!

machow (approved):

Issue #433, opened 2024-09-05
Allow for manual CSS style rules in `style=` arg of `tab_style()`


rich-iannone:
While we can add quite a few cell styles using `style.text(...)` and `style.fill(...)` there may be CSS styles that would be useful to have in a table but don't have a dedicated argument. For this reason, it would be good to be able to add CSS rules as strings (e.g., `"font-variant: small-caps;"`).

PR #624, opened 2025-02-27
docs: add `style.css` entry to API reference docs

rich-iannone:
The `style.css` class has been available for awhile but it wasn't present in the API reference docs. This PR adds a basic docstring for `style.css` and makes it available in the docs.

Fixes: https://github.com/posit-dev/great-tables/issues/433

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -20,7 +20,6 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install ".[all]"
-          python -m pip install pointblank
       - uses: quarto-dev/quarto-actions/setup@v2
         with:
           tinytex: true
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -179,6 +179,7 @@ quartodoc:
         - style.fill
         - style.text
         - style.borders
+        - style.css
     - title: Helper functions
       desc: >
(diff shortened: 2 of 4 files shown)


machow:
Taking a look at the funky docstring rendering now!

machow:
I can replicate the issue locally, but am so confused, because these quartodoc commands seem to work...

```python
from quartodoc import get_object, MdRenderer

obj = get_object("great_tables.style:css", dynamic=True)
renderer = MdRenderer(table_style="description-list")

renderer.render(obj)
```

machow:
It looks like in `.github/workflows/ci-docs.yml` pointblank gets installed. However, this also uninstalls the local great tables and installs the latest one on pypi (so the docstring is from that one and not the pr 😭 ).

image

Since the pointblank blogpost was frozen, I think we can remove the pointblank install, which should hopefully fix?!

rich-iannone:
Argh, I should have looked in the logs but I didn't suspect there would be anything there. I've removed that installation line from the docs-building CI workflow.

machow:
No worries -- if pip installing pointblank was wrong then I don't want to be right

I think one of the examples might be failing due to needing an import

image

rich-iannone:
That was from the last PR, added in the import of `md` just now.

rich-iannone:
Looks like everything is good now in the built site!

machow (approved):
LGTM

PR #625, opened 2025-03-02
feat: implement `cols_reorder()` for flexible column reordering  

jrycw:
Hello team,  

This PR introduces `cols_reorder()`, allowing users to reorder columns in a specified order easily.  

While the existing `move_*()` functions are useful for repositioning a few columns, they don't provide fine-grained control over column order. I've often had to rely on `Pandas` or `Polars` for this task. The implementation of this function follows a straightforward approach, similar to the existing `move_*()` functions.

Feedback is always welcome!😊


great_tables/_spanners.py
```diff
@@ -753,3 +753,58 @@ def cols_width(self: GTSelf, cases: dict[str, str] | None = None, **kwargs: str)
...
+    gt = GT(exibble)
+    gt
+    ```
+    Suppose you want to move the `fctr` column to the beginning of the table and position the `char`
+    and `num` columns as the last two. While this can be accomplished using `cols_move()`,
+    `cols_move_to_start()`, and `cols_move_to_end()`, **Great Tables** offers a more convenient
+    function: `cols_reorder()`. By passing a list of column names to the `columns=` parameter,
+    the table will be reordered accordingly:
+    ```{python}
+    col1, col2, col3, *cols = exibble.columns  # col1="num", col2="char", col3="fctr"
```
rich-iannone, line 795: This is a really good example!

great_tables/_spanners.py
```diff
@@ -753,3 +753,58 @@ def cols_width(self: GTSelf, cases: dict[str, str] | None = None, **kwargs: str)
...
+    gt.cols_reorder([col3, *cols, col2,  col1])
+    ```
+    """
+
+    if isinstance(columns, str):
+        columns = [columns]
+
```
rich-iannone, line 802: I wonder if we should raise here, if only to provide a more user-friendly message. The user will instead see `"Reordering vars must contain all boxhead vars."` if not providing all columns, which is good for a developer-facing error message but it exposes some of the underlying implementation that the user might not know about.

jrycw: Great suggestion! Should we raise the exception inside the function or modify the error message to be more user-friendly rather than developer-focused?

jrycw: How about `"Column reordering must include all columns in the table."`?

rich-iannone: I like it!

great_tables/_spanners.py
jrycw: I've renamed the variables to use their actual column names instead of generic names like `col1`. This should improve readability.

rich-iannone (approved):

rich-iannone:
Sorry this took so very long to review, but going to merge now!

jrycw:
Thanks @rich-iannone. The docstring and examples are much cleaner than my original draft. Great work.

PR #629, opened 2025-03-03
feat: add `cols_unhide()` method

jrycw:
Hello team,  

This PR ports the [cols_unhide()](https://gt.rstudio.com/reference/cols_unhide.html) function from `{gt}` to Great Tables.  

With this addition, users now have full control over column visibility at any state.  

Note that `Boxhead.set_cols_hidden()` and `Boxhead.set_cols_unhidden()` now rely on `Boxhead._set_cols_info_type()`.  
I'm unsure if this is the best approach—suggestions and feedback are welcome.

### DIFF
--- a/great_tables/_spanners.py
+++ b/great_tables/_spanners.py
@@ -559,6 +559,65 @@ def cols_hide(self: GTSelf, columns: SelectExpr) -> GTSelf:
     return self._replace(_boxhead=new_boxhead)
 
 
+def cols_unhide(self: GTSelf, columns: SelectExpr) -> GTSelf:
+    """Unhide one or more columns.
+
+    The `cols_unhide()` method allows us to unhide one or more columns from appearing in the final
+    output table. This may be important in cases where the user obtains a `GT` instance with hidden
+    columns and there is motivation to reveal one or more of those.
+
+    Parameters
+    ----------
+    columns
+        The columns to unhide in the output display table. Can either be a single column name or a
+        series of column names provided in a list.
+
+    Returns
+    -------
+    GT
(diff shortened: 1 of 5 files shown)


rich-iannone:
To me, the approach you took seems very elegant! Thanks for this contribution, will review right now. Also, if you merge `main` I think the CI Docs workflow should pass.

rich-iannone:
I am wrong about the merging with `main` tip... The PR that fixes the CI Docs workflow hasn't yet been merged :\ This should all be fixed soon!

great_tables/_spanners.py
```diff
@@ -559,6 +559,65 @@ def cols_hide(self: GTSelf, columns: SelectExpr) -> GTSelf:
...
+        ["country_name", "year", "population"]
+    ].tail(5)
+
+    GT(countrypops_mini).cols_hide(columns="year").cols_unhide(columns="year")
+    ```
+
+    See Also
+    --------
+    The counterpart of this function,
+    [`cols_hide()`](`great_tables._spanners.cols_hide`), allows you to hide one or more columns.
```
rich-iannone, line 602: This link *might* work (hard to verify right now) but I also know that the form `great_tables.GT.cols_hide` should definitely work.

rich-iannone (changes requested):
Looks great! I think the only change is to update the interlink to match the construction used in https://github.com/posit-dev/great-tables/blob/main/docs/_quarto.yml (look at the  links used in the API reference).

machow:
From pairing w/ Rich LGTM. I think the doc build issue is related to something fixed in our great_tables.style.css PR

Issue #567, opened 2024-12-17
css-inline requirement breaks shinylive


astrowonk:
## Prework

- [ ] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [ ] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

great_tables 0.15 introduces a dependency on css_inline. This is not a pure python library and so breaks compatibility with shinylive.
## Reproducible example

[Shiny Live failing with GT 0.15](https://shinylive.io/py/editor/#code=NobwRAdghgtgpmAXGKAHVA6VBPMAaMAYwHsIAXOcpMAMwCdiYACAZwAsBLCbJjmVYnTJMAgujxMArhwA6EOfUZMA5nThQyAfTJQARgBs4LXv0HCA4gBUJcAB4ddBuHL4ChKtRu17DLDOy4eKGNlMhY5COh0TWkmAF4pDiwoZThNGn1pABMAClC-YkkyVCLNUJyZMB0nSoBKWsi5LLgaVjg6ADd2nK4SsglC4qKJFiMWDlJaxDkmWaYAAUG+mbn5-Iw1CGa6MrIV2ebW6sMcqf252bUySToIJiscuwcnBvk3tFR40XQcj5iOEbtLp0WpgAC+eHA0Hg1DUAEdpGp4OQ-GRbGR8ERSBQqMhVOoyABaY5GOJxAAMGAAjABWDDkuSoKBbYLggC6QA)

[...]

PR #634, opened 2025-03-07
fix!: move `css-inline` pkg to extra group

rich-iannone:
Given that `css-inline` does not work with pyodide and is really only ever used in `as_raw_html()`, this PR makes this library and optional dependency. It's placed in the 'extra' optional install group (along with packages for other export functions like `save()`).

Fixes: https://github.com/posit-dev/great-tables/issues/567

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -217,6 +217,10 @@ def as_raw_html(
     gt_tbl.as_raw_html(inline_css=True)
     ```
     """
+
+    if inline_css:
+        _try_import(name="css-inline", pip_install_line="pip install css-inline")
+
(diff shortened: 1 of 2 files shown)


machow (approved):
LGTM, we'll push removing the `table_css` piece (which in testing seemed to be unapplied; since the css used is in the rendered table) to another PR

machow (approved):
LGTM, we'll push removing the `table_css` piece (which in testing seemed to be unapplied; since the css used is in the rendered table) to another PR

Stranger6667:
> Given that css-inline does not work with pyodide

Not sure if still relevant, but `css-inline` now works with pyodide (since `0.14.6`)

Issue #640, opened 2025-03-18
Support pickling


malcolmbarrett:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal

I'm trying to load pre-computed gt tables in Quarto files using pickling. Unfortunately, unpickling doesn't work

```python
from great_tables import GT
from great_tables.data import islands
import pickle
islands_mini = islands.head(10)
gt_tbl = GT(islands_mini)

with open("gt.pkl", "wb") as f:
    pickle.dump(gt_tbl, f)

with open("gt.pkl", "rb") as f:
    gt_tbl2 = pickle.load(f)
```
gives me

```
TypeError: Boxhead.__new__() missing 1 required positional argument: 'data'
Cell In[42], line 11
      8     pickle.dump(gt_tbl, f)
     10 with open("gt.pkl", "rb") as f:
---> 11     gt_tbl2 = pickle.load(f)
```

It seems that supporting unpickling would require and additional method on the GT side: https://stackoverflow.com/a/37753992

PR #641, opened 2025-03-18
feat: Support pickling

jrycw:
Fixes: #640 

This PR adds support for pickling `GT`. The `__getnewargs__` method is relatively new to me, but the implementation appears to be straightforward.

Edit: Based on the error message, it seems that providing only the `data=` parameter is sufficient, but I'm unsure if the other three parameters—`auto_align=`, `rowname_col=`, and `groupname_col=`—are also required.

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -283,10 +283,10 @@ class Boxhead(_Sequence[ColInfo]):
         rowname_col: str | None = None,
         groupname_col: str | None = None,
     ):
-        self._data = data
+        pass
 
     def __getnewargs__(self):
-        return (self._data,)
+        return (self._d,)
 
     def set_stub_cols(self, rowname_col: str | None, groupname_col: str | None) -> Self:
         # Note that None unsets a column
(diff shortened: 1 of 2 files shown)


machow:
@malcolmbarrett any chance you might be able to install and check if this does what you need?

malcolmbarrett:
Works great for me! Thanks to you both

great_tables/_gt_data.py
```diff
@@ -283,7 +283,10 @@ def __init__(
         rowname_col: str | None = None,
         groupname_col: str | None = None,
     ):
-        pass
+        self._data = data
```
machow, line 286: I feel like this `__init__` quirk is related to how I put `__new__` in, but I'm not totally sure how to fix it 😅

machow (approved):
LGTM Thanks for this! It might be worth taking a look at how `__new__` is used here down the road, which was definitely probably my screw up..!

great_tables/_gt_data.py
jrycw: @machow, thanks for the feedback! I think returning `self._d` for `__getnewargs__` might be a more correct approach.

Issue #142, opened 2024-01-30
Cannot style cell borders


davidchall:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

Following the documentation at https://posit-dev.github.io/great-tables/get-started/basic-styling.html#customizing-borders, I would like to set the border style of selected table cells.

## Reproducible example

```python
import great_tables as gt
from great_tables.data import airquality

gt_air = gt.GT(airquality.head())

gt_air.tab_style(
    style=gt.style.borders(sides="top", color="red", style="dashed", weight="3px"),
    locations=gt.loc.body(rows=[1, 2])
)
```

![image](https://github.com/posit-dev/great-tables/assets/1804856/f7591244-c50f-4d48-a5e0-24d4ff90e549)

## Expected result

Screenshot from above link:

![image](https://github.com/posit-dev/great-tables/assets/1804856/bfa58767-f595-4b5a-afee-66a680ed26f8)

## Development environment

- Operating System: Windows
- **great_tables** Version: 0.2.0
- IDE: VS Code with Jupyter Notebook extension

rich-iannone:
This should now display properly in VSCode now that we've improved the `_repr_html_` part of the output. Closing now, but feel free to reopen if there's still a problem with the latest version of Great Tables.

rich-iannone:
Upon revisiting this, I think the best thing one could do when iterating on table styling (and I've been doing much more of this lately) is to use `.show("browser")`. There are several more issues with tables viewed in VS Code (forced font sizes in the header, table cell text overflow not being honored, etc.).

davidchall:
Hi @rich-iannone, thanks for revisiting this issue. Can we add this tip to the table styling docs, so others can identify the cause and recommended solution for this issue?

PR #643, opened 2025-03-21
docs: add note about using `.show("browser")` when in VS Code

rich-iannone:
A frequent point of confusion is styling tables in VS Code using Jupyter and not seeing styles being applied. While we fixed a few issues with the display of tables there, it seems impractical to try to fix them all. A better recommendation is to view the table in a browser through `.show("browser")` when using VS Code.

To get the word out about this, it was brought up that a good place to inform users about this is in the docs. In this PR, a callout note is placed near the beginning of the [Styling the Table Body](https://posit-dev.github.io/great-tables/get-started/basic-styling.html) article in the Get Started guide.

Fixes: https://github.com/posit-dev/great-tables/issues/142

### DIFF
--- a/docs/get-started/basic-styling.qmd
+++ b/docs/get-started/basic-styling.qmd
@@ -23,7 +23,7 @@ gt_pl_air = GT(pl.from_pandas(air_head))
 
 
 :::{.callout-note}
-When using Great Tables with VS Code, some forms of table styling can be suppressed by the IDE's special handling of table outputs. For this reason, it's strongly recommended that `.show("browser")` be called on GT objects when using VS Code.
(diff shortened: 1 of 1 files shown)


docs/get-started/basic-styling.qmd
```diff
@@ -21,6 +21,11 @@ gt_air = GT(air_head)
 gt_pl_air = GT(pl.from_pandas(air_head))
 ```
 
+
+:::{.callout-note}
+When using Great Tables with VS Code, some forms of table styling can be suppressed by the IDE's special handling of table outputs. For this reason, it's strongly recommended that `.show("browser")` be called on GT objects when using VS Code.
```
machow, line 26:
```suggestion
When using Great Tables with VS Code, the IDE suppresses some forms of table styling displayed in notebooks. For example, border styles might not appear. Use `.show("browser")` to see the styled GT table in a separate browser window.
```

machow (changes requested):

Issue #639, opened 2025-03-17
Allow custom datetime formats


felixgwilliams:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal

Currently, the great-tables package allows formatting dates according to one of a fixed list of formats, specified by the `date_style` or `time_style` options in the `fmt_datetime`, `fmt_date` and `fmt_time` methods. Since the format I wanted (Month + Year) is not in the list, I would like to be able to specify a custom format in these functions. 

I'm not sure what the API should be, but one idea could be to make extra arguments like `date_format_str` and `time_format_str` which are `None` by default, and override the `*_style` arguments.

For example, this part https://github.com/posit-dev/great-tables/blob/e01e682c35168211c7e74df5a16c0b025d3be849/great_tables/_formats.py#L2187

could be 

```python
if date_format_str is None:
    date_format_str = _get_date_format(date_style=date_style) 
```

I am happy to have a go at a pull request for this.

rich-iannone:
Thanks for this issue! We will incorporate a `format=` arg to allow for custom date/time formatting.

EnsuingRequiem:
It would be nice if the `format_str` parameter was also added to `fmt_date`

PR #645, opened 2025-03-21
feat: custom datetime formatting in `.fmt_datetime()`

rich-iannone:
This adds the `format_str=` parameter to the `.fmt_datetime()` so that users can use any strftime formatting string. Added tests to ensure that this works for both date and datetime inputs.

Fixes: https://github.com/posit-dev/great-tables/issues/639


great_tables/_formats.py
```diff
@@ -2215,26 +2222,36 @@ def fmt_datetime_context(
     if is_na(data._tbl_data, x):
         return x
 
-    # From the date and time format strings, create a datetime format string
-    datetime_format_str = f"{date_format_str}'{sep}'{time_format_str}"
+    if format_str is not None:
+        # If `x` is a string, assume it is an ISO datetime string and convert
+        # it to a datetime object
+        if isinstance(x, str):
```
machow: Can you pull this above the block, so it applies to the if and else case? (and then remove the redundant version on 2239 in the else)

great_tables/_formats.py
```diff
@@ -2215,26 +2222,36 @@ def fmt_datetime_context(
...
 
-    # If `x` is a string, assume it is an ISO datetime string and convert it to a datetime object
-    if isinstance(x, str):
-        # Convert the ISO datetime string to a datetime object
-        x = _iso_str_to_datetime(x)
+        x_formatted = x.strftime(format_str)
 
     else:
-        # Stop if `x` is not a valid datetime object
-        _validate_datetime_obj(x=x)
```
machow, line 2228: AFAICT this if / else should just stay the same, and the extra formatting logic should go below it

machow (approved):
LGTM, thanks!

Issue #649, opened 2025-03-26
Add `compact` to `fmt_currency`


gartinian:
## Prework

- [x] Read and abide by the **great_tables** [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/posit-dev/great_tables/issues) (both open and closed).

## Proposal
Add the `compact` parameter to `fmt_currency`, so that values such as `$12,345` can render as `$12K`.  The intention is to mirror the behavior of `compact` in other formatters, such as `fmt_integer`.

I have reviewed the code and believe I know where changes should be made.  I am happy to submit a PR if there is alignment on moving forward.

rich-iannone:
This is definitely an unintentional omission. We’d be happy to accept a PR for this!

gartinian:
Thank you for implementing!

PR #664, opened 2025-04-10
feat: add `compact=` arg to for `fmt_currency()`

rich-iannone:
This PR adds the `compact=` parameter to `fmt_currency()`, so that values such as $12,345 can render as $12.34K. The behavior of `compact=` here is the same as in other formatters like `fmt_number()` and  `fmt_integer()`.

Here's an example of its use:

```python

GT(exibble[["num", "currency"]]).fmt_currency(columns="currency", scale_by=1000, compact=True)
```

image

Fixes: https://github.com/posit-dev/great-tables/issues/649


great_tables/_formats.py
```diff
@@ -1328,17 +1335,30 @@ def fmt_currency_context(
...
-        decimals=decimals,
-        n_sigfig=None,
-        drop_trailing_zeros=False,
-        drop_trailing_dec_mark=drop_trailing_dec_mark,
-        use_seps=use_seps,
-        sep_mark=sep_mark,
-        dec_mark=dec_mark,
-        force_sign=force_sign,
-    )
+    if compact:
```
machow, line 1338: I think you could define the function to use in the if/else block, and then call it after, eg...

```python
if compact:
    f_formatter = _format_number_compactly
else:
    f_formatter = _value_to_decimal_notation

x_formatted = f_formatter(...)
```

machow (changes requested):
reviewed while pairing...

great_tables/_formats.py
rich-iannone: This has now been changed. Thanks!

Issue #526, opened 2024-11-29
DateStyle and TimeStyle are not part of the public API


BenGale93:
Hi, thanks for the great library!

I'm writing a wrapper function for `gt.fmt_datetime`, `gt.fmt_date` etc. but I need to import `DateStyle` and `TimeStyle` from `great_tables._formats` to type hint the function correctly.

Have you considered re-exporting these Literals so they are accessible from the public API? Happy to make the PR myself, as well as re-export any other type hints that are in the public API if you agree this should be done.

jrycw:
Thanks for bringing this up.

@machow and @rich-iannone, do you think we should consider exposing some literals to make it easier for users to adopt type hints?

I fully understand that the linter has difficulty understanding our `databackend` logic, and making `mypy` work smoothly is definitely a challenge for us.

rich-iannone:
@jrycw @BenGale93 Sorry for the long delay in responding to this. I *do* think we should more publicly expose literal values (for all the advantages listed above).

rich-iannone:
@BenGale93 Would it work for `DateStyle` and `TimeStyle` to be exposed in a new `great_tables.types` module?

BenGale93:
Sounds good to me

machow:
Renamed and labeled an epic, since with the release of gt-extras, it seems really useful to think about how to export types for extension packages / usecases

machow:
I would lean towards us consolidating types into a private _types.py module or something, that reflect types ext developers will get value out of, but doesn't totally guarantee stability yet

PR #672, opened 2025-04-23
fix: move literals to `types.py`

jrycw:
Hello team,

This PR provides an initial implementation for #526.

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -7,10 +7,9 @@ import webbrowser
 from functools import partial
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 from pathlib import Path
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING
 
-from typing_extensions import TypeAlias
-
+from .types import DebugDumpOptions, RenderTargets, WebDrivers
 from ._helpers import random_id
 from ._scss import compile_scss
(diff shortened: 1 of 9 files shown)


rich-iannone (approved):
LGTM!

jrycw:
Could the team help review which literals (if not all) should be included in the `__all__` in `__init__.py`?

machow:
Thanks for this PR -- we're actively discussing types right now. No need to update this PR right now. I'll be sure to come back with more fleshed out thoughts, if that's okay!

### Background

We'll be working on a extension package for Great Tables (with extra nanoplots, themes, etc.. that we don't want to load up the main package too hard with), so this piece will be super relevant.

Two big pieces for us to think about:

* Organizing types: how should type annotations be organized and communicated in types.py?
* Exposing types: do we want a public types module right now?

I feel like the [polars._types](https://github.com/pola-rs/polars/blob/main/py-polars/polars/_typing.py) submodule has been a helpful reference. It looks like they deprecated their `polars.type_aliases` submodule for now, so it might be good to marinate this feature for a bit.

machow:
We're sprinting to merge and clear many of the outstanding PRs. I'm going to close for now, but I marked the linked issue as an epic, and definitely think it'll be good for us to come back to. Thanks for taking the time to work on this, this is mostly an issue of us getting things out the door!

PR #673, opened 2025-04-24
Add truncate option to data_color

### LINKED ISSUE #430: Feature request: add argument to `data_color` that truncates values outside of the domain instead of setting them to `nan`
Currently, values that are outside the prescribed `domain` are going to be treated like nan/null values inside `data_color`:

image

(ref: https://posit-dev.github.io/great-tables/reference/GT.html#great_tables.GT.data_color)

I think it would be a useful feature to allow a flag that enables a different behavior: Instead of treating the values as nan/null, they should be truncated to the min/max values of the domain. E.g.:
```python
gt.GT(gt.data.exibble).data_color(
    columns="currency",
    palette=["red", "green"],
    domain=[0, 50],
    na_color="lightgray"
    truncate=True          # <<<<<<<<<<< This could be the new flag.
)
```

Then the entries `65100.0` and `1325.81` would get the same color as a value of `50`.

If this is a feature that makes sense, the `_rescale_numeric` could be a good place to add this to. https://github.com/posit-dev/great-tables/blob/11660a66b9291e137db93a74b985941f4a63ad90/great_tables/_data_color/base.py#L572

The change could look like this: 
```python

    if truncate: 
        scaled_vals = [min(1, max(0, x)) if not is_na(df, x) else np.nan for x in scaled_vals]    
    else:
        scaled_vals = [x if not is_na(df, x) and (x >= 0 and x <= 1) else np.nan for x in scaled_vals]
```

mahdibaghbanzadeh:
## Summary

This PR adds a new boolean parameter `truncate` to the `data_color` method. When `truncate=True`, values outside the specified domain will be assigned the color of the domain's minimum or maximum value instead of being treated as NaN. The default is set to `truncate=False`, which follows the previous functionality.

Changes include:
- Added the `truncate` parameter (default: False)
- Updated the `_rescale_numeric` function to support truncation
- Added unit tests `test_data_color_truncate` in `tests/data_color/test_data_color.py` to verify that truncation works correctly

This close #430

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import numpy as np
-from great_tables._locations import resolve_cols_c, resolve_rows_i, RowSelectExpr
-from great_tables._tbl_data import DataFrameLike, is_na, SelectExpr
+from typing_extensions import TypeAlias
+
+from great_tables._locations import RowSelectExpr, resolve_cols_c, resolve_rows_i
+from great_tables._tbl_data import DataFrameLike, SelectExpr, is_na
 from great_tables.loc import body
 from great_tables.style import fill, text
-from typing_extensions import TypeAlias
 
 from .constants import ALL_PALETTES, COLOR_NAME_TO_HEX, DEFAULT_PALETTE
 
@@ -28,6 +29,7 @@ def data_color(
     alpha: int | float | None = None,
     reverse: bool = False,
     autocolor_text: bool = True,
+    truncate: bool = False,
 ) -> GTSelf:
     """
     Perform data cell colorization.
@@ -73,6 +75,11 @@ def data_color(
     autocolor_text
         Whether or not to automatically color the text of the data values. If `True`, then the text
         will be colored according to the background color of the cell.
+    truncate
+        If `True`, then any values that fall outside of the domain will be truncated to the
+        minimum or maximum value of the domain (will have the same color). If `False`, then any
+        values that fall outside of the domain will be set to `NaN` and will follow the `na_color=`
+        color.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


great_tables/_data_color/base.py
```diff
@@ -587,10 +596,29 @@ def _rescale_numeric(
         scaled_vals = [0.0 if not is_na(df, x) else x for x in vals]
     else:
         # Rescale the values in `vals` to the range [0, 1], pass through NA values
-        scaled_vals = [(x - domain_min) / domain_range if not is_na(df, x) else x for x in vals]
-
-    # Add NA values to any values in `scaled_vals` that are not in the [0, 1] range
-    scaled_vals = [x if not is_na(df, x) and (x >= 0 and x <= 1) else np.nan for x in scaled_vals]
+        scaled_vals: list[float] = []
```
machow, line 599: WDYT of us doing this  in two stages?:

1. apply translation and scale
2. truncate (so values outside range are set to 0 or domain_range)

If useful, we could use an iterator for (1) so it doesn't create an intermediate list (though, I suspect optimizing here isn't too critical!)

Something like...

```python
filled = (np.nan if is_na(df, x) for x in vals)

# Rescale the value to the range [0, 1] ----
scaled = ((x - domain_min) / domain_range for x in filled)

if truncate:
    min_val = 0.0
    max_val = 1.0
else:
    min_val = np.nan
    max_val = np.nan

truncated = [min_val if x < 0 else max_val if x > 1 else x for x in scaled]

return truncated
```

The last list comp is a bit gnarly, but I think what's helpful for me above is it feels like it helps me see the conditions table in the loop in the PR:

scale_val < 0, truncate True
scale_val < 0, truncate False
scale_val > 1, truncate True
scale_val > 1, truncate False

Since in the code above, all 4 replacement decisions are made in the if-else truncate block.

machow (changes requested):
Thanks for this PR, and I'm so sorry for taking a while to get to it 😓. I left a suggestion for an approach that might help flatten out the if-else conditions a bit. WDYT? I'm happy to pick up and finish if useful at all!

Yasin197:
@machow hope this PR is going to be integrated into the next release of GT!

Thank you!

machow:
@rich-iannone I think this is ready if you don't mind reviewing! I didn't look in-depth at the documentation / try it, just tweaked the code

rich-iannone (approved):
LGTM and thanks!

PR #679, opened 2025-04-29
Fix library agnosticism in random locations

### LINKED ISSUE #675: Pandas is a ghost requirement?
This test fails with a ModuleNotFound error.

`def test_one_column_table_no_lib() -> None:
    sys.modules["pandas"] = None

    with pytest.raises(ModuleNotFoundError):
        import pandas

    vals = [1, 2, 3]

    _make_one_col_table(vals=vals)
`

I dug into the code and I believe the the issue is a type checking block blindly imports some libraries if an instance check is triggered. Most of the instance checks seem to happen when a method is wrapped with single dispatch. Below is the block in question in this file great_tables/_tbl_data.py.

`if TYPE_CHECKING:
    import numpy as np
    import pandas as pd
    import polars as pl
    import pyarrow as pa`

I tried to fix it myself but the scope of it felt pretty large. Is this even a bug? From my understand Pandas was not a requirement?

tylerriccio33:
This is meant to fix #675 in GT and https://github.com/posit-dev/pointblank/issues/132 in pointblank.

I'm brute forcing it since I see no other way. I'm replacing the single dispatch paradigm (in effected places) with an attempt to import the library in question.

FYI I'm 99% sure this would be solved by Narwhals if you wanted to go that route down the line.

### DIFF
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
 from pathlib import Path
+from typing import TYPE_CHECKING, Any
 
 from typing_extensions import TypeAlias
 
-from .gt import GT, _get_column_of_values
 from ._gt_data import GTData
 from ._tbl_data import SeriesLike, to_frame
+from .gt import GT, _get_column_of_values
 
 if TYPE_CHECKING:
     from ._formats import DateStyle, TimeStyle
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import re
-import warnings
+import sys
 from functools import singledispatch
-from typing import TYPE_CHECKING, Any, Callable, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Union
 
 from typing_extensions import TypeAlias
 
@@ -761,7 +761,6 @@ def _(df: PyArrowTable, x: Any) -> bool:
     return arr.is_null().to_pylist()[0] or arr.is_nan().to_pylist()[0]
 
 
-@singledispatch
 def validate_frame(df: DataFrameLike) -> DataFrameLike:
     """Raises an error if a DataFrame is not supported by Great Tables.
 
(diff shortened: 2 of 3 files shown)


tylerriccio33:
Actually, can I get another set of eyes on this issue? The scope all of the sudden seems massive. I thought it was contained to a few accidental instance checks on single dispatch methods, but it appears to be instance checking a TON of functions.

If I'm scoping this correctly, I think there are 3 options:

1. Doing a naive check like 'pandas' in str(type(df)): ...
2. Doing a try-except import check like try: __import__('pandas') ...
3. Adopting narwhals

I'm currently unaware of a way to do an instance check on a class you do not have available in your process. Happy to help out w/any of these, I would be aided by this being solved.

I'd be even happier if you were to tell me I'm wrong and the problem is far more contained :)

machow:
Hey, thanks for taking a look at this, and raising the original issue. If you have the chance, do you mind taking a look at this explanation, and letting me know if it does what you need?

The tl;dr is that you should be able to pass a polars series to the val functions, but we should better document this 😅 

https://github.com/posit-dev/great-tables/issues/675#issuecomment-2855706009

jrycw:
@tylerriccio33, You might find our [Bring Your Own DataFrame](https://posit-dev.github.io/great-tables/blog/bring-your-own-df/#how-we-made-pandas-optional) blog post helpful—feel free to check it out.

tylerriccio33:
I'm not the expert in how single dispatch decorator works in python, but it appears to be running an is instance check when to_frame is called, which I don't think is intended. Even when i add code to convert the the list to a polars series, then pass it to to_frame, it does an instance check and fails since pandas isn't available. 

I feel I may be not understanding a key concept here, and there can't fix this. I'd really like to fix this bug, I'm running into some compatibility issues at work; it's sometimes difficult to load prod containers w/pandas due to other compatibility concerns (arrow/numpy). Let me know if i'm understanding this correctly.

Pushed a new pr to show the test I'm using, which may also be the problem?

jrycw:
I came across a [PyCon 2025 talk](https://www.youtube.com/watch?v=hidy15rK2a4) on YouTube that briefly yet clearly outlines four different ways to choose actions based on instance types.

tylerriccio33:
The error is happening before the function code is called, it's happening during the evaluation process in singledispatch. In `AbstractBackend` `__subclasshook__` method, I pass a series but the only cls._backend is a Pandas series, so it cannot import the module. This feels very specific to the data backend implementation so I'm having a hard time debugging where the backend class attribute is set...

tylerriccio33:
This pr is imperfect but working. I found it wasn't working because Polar's `Series` object isn't defined or available at the top level of the module. Instead, it's down a few levels. To get around this I added the full path (series.series.Series) to the backend.

`class PlSeries(AbstractBackend):
        _backends: list[tuple[str, str]] = [("polars", "series.series.Series")]`
        
I then added additional logic to allow for a class name not defined at the top level:

`class AbstractBackend(metaclass=_AbstractBackendMeta):
    @classmethod
    def __init_subclass__(cls):
        if not hasattr(cls, "_backends"):
            cls._backends = []

    @classmethod
    def __subclasshook__(cls, subclass):
        for mod_name, cls_name in cls._backends:
            try:
                parent_candidate = _load_class(mod_name, cls_name)
                if issubclass(subclass, parent_candidate):
                    return True
            except (ModuleNotFoundError, ImportError):
                continue

        return NotImplemented`
        
And the new load_class function which is the workhorse:
`def _load_class(mod_name: str, cls_name: str) -> type:
    # follow the path of mods if it is one
    listified_mod_path: list[str] = cls_name.split(".")

    sub_module_path: list[str] = [mod_name] + listified_mod_path[:-1]

    cur_mod: ModuleType = importlib.import_module(sub_module_path[0])
    _cls_name: str = listified_mod_path[-1]

    if len(sub_module_path) == 1:  # no submodules, just the module itself
        return getattr(cur_mod, _cls_name)

    # now, we must increment the module paths to import the final class
    remaining_mods: list[str] = sub_module_path[1:]
    for mod in remaining_mods:
        cur_mod: ModuleType = getattr(cur_mod, mod)

    return getattr(cur_mod, _cls_name)`
        
Again, not the most elegant but it appears to be effective. I've struggled to wrap my head around some aspects of how the backend works, so it's possible I'm missing finer points.

Also, strictly for development purposes, I redefined some dev deps in the pyproject.toml file for uv to find, and added uv.lock to the gitignore. Feel free to reject that, just helped me code faster!

machow:
I've added a test in a PR to confirm that pandas does not get imported when the vals functions run. More context in the original issue. Do you mind taking a look, and letting me know if more tests / other testing behaviors would be useful?

https://github.com/posit-dev/great-tables/issues/675#issuecomment-2931890996

machow:
I'm going to close (see comment in #691) -- but happy to tee up anything useful in an issue

PR #689, opened 2025-06-02
tests: add no pandas dependency test of vals function

machow:
This PR adds a test of a vals function, that demonstrates it working on a Polars series, without having Pandas installed.

This is intended to provide a simple test of https://github.com/posit-dev/great-tables/issues/675, showing Pandas is not required to use vals. However, it's fairly simple, so more testing / looking around may be needed.


rich-iannone (approved):
LGTM!

PR #691, opened 2025-06-04
Add no pandas dep test make one column table

tylerriccio33:
This is a response to #689 that tests the ability to pass a non-polars series to a table. I'm curious to see if this runs in no pandas mode, I was under the impression it wasn't

### DIFF
--- a/tests/test_dependencies.py
+++ b/tests/test_dependencies.py
@@ -1,6 +1,7 @@
 # Tests to be run with pandas uninstalled.
 
 import pytest
+from great_tables._formats_vals import _make_one_col_table
 
 # mark file as nopandas
 pytestmark = pytest.mark.no_pandas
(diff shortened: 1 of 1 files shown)


tylerriccio33:
So, the no pandas fails since it prohibits a list. I think simply what I'm looking for is for this test to pass. This function is called a great many times internally, and incidentally in a number of functions I need, and perhaps others (see https://github.com/posit-dev/pointblank/issues/132).

Instead of using passing a series instead which would've required a number of code base changes across GT, I assumed the easiest way would be to just allow passing a list. Maybe i'm overthinking this. Happy to help w/changes and open prs

machow:
Are you saying that you want Great Tables to use a specific library (pandas or polars) when a list is passed? Or is it that you'd like Great Tables to be able to operate without depending on any DataFrame library?

I do think we could implement a very simple DataFrame-like class inside Great Tables, that would enable it to work without any DataFrame dependency. If that's what you want, do you mind opening a feature request?

machow:
I'm going to close for now, but please do open an issue if enabling certain parts of Great Tables to operate without any frame library dependencies would be usefully (currently it requires Pandas or Polars, and what gets used is under the user's control)

PR #695, opened 2025-06-07
Fix/spanner

### LINKED ISSUE #688: loc.spanner_labels does not work with multiple levels
## Prework
- [X] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [X] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

`great_tables.loc.spanner_labels` does not work for levels above the lowest level

## Reproducible example
```python
from great_tables import GT, exibble, loc, style
gt = (
    GT(exibble.loc[[0, 1], ["char", "fctr", "row"]])
    .tab_spanner("spanner1_level0", ["fctr"], level=0)
    .tab_spanner("spanner2_level0", ["char"], level=0)
    .tab_spanner("spanner_level1", ["char", "fctr"], level=1)
)
color1 = "#a6cee3"
color2 = "#b15928"
(
    gt.tab_style(style.fill(color1), loc.spanner_labels(ids=["spanner1_level0"]))
    .tab_style(style.fill(color2), loc.spanner_labels(ids=["spanner2_level0"]))
    .tab_style(style.fill(color2), loc.spanner_labels(ids=["spanner_level1"]))
)
```
- [X] Post a minimal reproducible example (MRE) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
  - [X] **Runnable**: post enough code and data so any onlooker can create the error on their own computer.
  - [X] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
  - [X] **Readable**: format your code according to the [Style Guide for Python Code](https://peps.python.org/pep-0008/).

## Expected result

spanner_level1 should be filled with same color as spanner2_level0, but is not filled at all.

## Development environment

- Operating System: Linux
- **great_tables** Version: 0.17.0

## Additional context

Image

ChristopherRussell:
# Summary

Fixes #688.

`loc.spanner_labels()` styling was not working for spanner labels at levels above the lowest level. Styles applied to higher-level spanners using `tab_style()` with `loc.spanner_labels()` were ignored.

## Cause
Higher-level spanner elements in the HTML output were missing `id` attributes, which prevented the styling logic from targeting them.

## Solution
- Added `id` attributes to all spanner levels in `create_columns_component_h()`
- Fixed filtering logic to correctly apply styles to higher-level spanners
- Updated test snapshots to reflect the new (correct) HTML structure

## Testing
- Visually verified multi-level spanner styling now works in issue example and one other (see below).
- All existing tests passed except one, which required an update to the snapshot to include `id` attributes where they were previously missing.

- Ref: #688 

# Checklist

# New example:

Here is another example which demonstrates the bug fix working.

```python
from great_tables import GT, exibble, loc, style
(
    GT(exibble, rowname_col="row", groupname_col="group")
    .tab_spanner("A", ["num", "char", "fctr"])
    .tab_spanner("B", ["fctr"])
    .tab_spanner("C", ["num", "char"])
    .tab_spanner("D", ["fctr", "date", "time"])
    .tab_spanner("E", spanners=["B", "C"])
    .tab_stubhead(label="Group")
    # Add styling to test multi-level spanner support
    .tab_style(style=style.fill(color="#b15928"), locations=loc.spanner_labels(ids=["E"]))
    .tab_style(style=style.fill(color="#2166ac"), locations=loc.spanner_labels(ids=["A"]))
    .tab_style(style=style.fill(color="#5aae61"), locations=loc.spanner_labels(ids=["D"]))
)
```

After:
Screenshot 2025-06-07 at 11 08 42

Before:
Screenshot 2025-06-07 at 11 08 19

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -277,13 +277,12 @@ def create_columns_component_h(data: GTData) -> str:
                 # If colspans[i] == 0, it means that a previous cell's
                 # `colspan` will cover us
                 if colspans[ii] > 0:
-                    # Filter by column label / id, join with overall column labels style
-                    # TODO check this filter logic
+                    # Filter by spanner label / id, join with overall column labels style
                     styles_i = [
                         x
                         for x in styles_spanner_label
-                        # TODO: refactor use of set
-                        if set(x.grpname) & set([spanner_ids_level_1_index[ii]])
+                        if spanner_ids_level_1_index[ii]
+                        and spanner_ids_level_1_index[ii] in x.grpname
                     ]
 
                     level_1_spanners.append(
@@ -366,13 +365,9 @@ def create_columns_component_h(data: GTData) -> str:
 
             for colspan, span_label in zip(colspans, spanners_row.values()):
                 if colspan > 0:
-                    # Filter by column label / id, join with overall column labels style
-                    # TODO check this filter logic
+                    # Filter by spanner label / id, join with overall column labels style
                     styles_i = [
-                        x
-                        for x in styles_column_label
-                        # TODO: refactor use of set
-                        if set(x.grpname) & set([colspan, span_label])
+                        x for x in styles_spanner_label if span_label and span_label in x.grpname
                     ]
 
                     if span_label:
@@ -391,6 +386,7 @@ def create_columns_component_h(data: GTData) -> str:
                             colspan=colspan,
                             style=_flatten_styles(styles_column_labels + styles_i),
                             scope="colgroup" if colspan > 1 else "col",
+                            id=_create_element_id(table_id, span_label),
                         )
                     )
 
--- a/tests/__snapshots__/test_utils_render_html.ambr
+++ b/tests/__snapshots__/test_utils_render_html.ambr
@@ -89,10 +89,10 @@
     <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col">
       <span>&nbsp</span>
     </th>
-    <th class="gt_center gt_columns_bottom_border gt_columns_top_border gt_column_spanner_outer" rowspan="1" colspan="3" scope="colgroup">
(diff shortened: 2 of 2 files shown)


machow:
@rich-iannone do you mind taking a look at this? From a glance it LGTM but I'm not too familiar with the use of id attributes in Great Tables

rich-iannone:
Thank you @ChristopherRussell for this PR! With your changes, styles are now captured for all spanners that have styles set. 

We don't need `id` attributes in the rendered span tags for this to work and I'd prefer if they were removed. The reason is that they can be problematic (in some contexts, like Quarto) when they are not distinct from all other `id` values in a page (this assumes a GT table is placed in a web page) since it will no longer be valid HTML5 markup.

I know that we have `id` attributes in the column label tags (and we need those for accessibility reasons, see: https://webaim.org/techniques/tables/data#id) but those too are currently problematic as they  are (they should combine the table `id`, a random string if not set, with the column header `id`; will make a separate issue for this).

I left a suggestion to omit the line that produces the `id` attribute in the HTML. Could you update the snapshot (with `make test-update`) once that change is made?

ChristopherRussell:
Thanks for finishing that off @rich-iannone, and apologies that I didn't get round to it quickly enough!

rich-iannone:
@ChristopherRussell it’s okay! Really happy to get your fix in for the next release!

jsenergy:
Many thanks for this fix @ChristopherRussell 

I would however call it incomplete due to the nature of skipping the id attribute.

My table has unique level 0 spanner labels, but then identical level 1 spanner labels.

With this fix I am still unable to tab_style a location using the unique ids of level 1 spanners, nor can i use the labels of the level 1 spanners as they are not unique.

While this is an improvement, it is still not a fix for non-unique spanner labels. 

Do you have the new issue link somewhere for me to track?

PR #697, opened 2025-06-11
docs: use improved interlinking available in quartodoc

rich-iannone:
This PR focuses on using the new autolink feature now available in quartodoc. These changes are focused only pages in the docs site (excluding blog posts).


machow:
Thanks for starting this! The two big pieces to get them autolinking is 

1. Update the interlinks extension using `quarto add machow/quartodoc`
2. Add an alias config to interlinks setting aliases: quartodoc: null. See https://machow.github.io/quartodoc/get-started/interlinks-autolink.html#basic-use

rich-iannone:
@machow thanks! I was going to ask you about this very thing.

machow (approved):
This LGTM, thanks for doing this! I scanned the docs preview, but did not check super closely each piece

rich-iannone:
This looks pretty complete now. That issue we talked about during pairing about `{.qd-no-link}` not working actually works fine  in https://github.com/posit-dev/great-tables/pull/697/commits/89f068d6bb6f1369d09a18b90738661e0f202411 (and in many pages in Pointblank).

That the line:

```
### Combining Styling with `~~GT.cols_hide()`{.qd-no-link}
```

didn't render correctly from https://github.com/posit-dev/great-tables/pull/697/commits/0485bb65645208f936c0242d0ba4a2ba70516993 (what I mentioned during pairing, but couldn't locate at the time) is not apparent to me. The site preview for that commit is no longer available but I could make a commit to re-create the potential bug (with a preview) if you think it's worth investigating.

machow:
I don't mind taking a look! Do you remember what ended up being rendered and why wasn't it correct? I'm confused why the link is doing these two things together:

* using the `~~` autolink syntax
* using .qd-no-link to disable autolinking

I'd expect exactly this code element `~~GT.cols_hide()` to be rendered, so the `~~` would appear because autolinking was disabled

machow:
Feel free to merge in the meantime!

rich-iannone:
This makes sense! I've misunderstood on a fundamental level how it used because the bigger use case (to me anyway) is that

- GT (with backticks) will link
- GT with `{.qd-no-link}` will remove the link, but still render GT in a code font as backticks do

So a non-issue after all. And I've learned something as well. Will merge!

PR #701, opened 2025-06-12
fix: striping indexing issue

### LINKED ISSUE #700: `opt_row_striping()` doesn't properly stripe grouped data
## Prework

## Description

Calling `opt_row_striping()` on a table with grouped data (specifically where the groups do not have a one-to-one relationship with the original table data, unlike the exibble example) results in stripes that correspond with the original row indices data rather than the indices based on the re-arranged groups.

## Reproducible example

```{python}
from great_tables import GT
from great_tables.data import reactions
import polars as pl

reactions_mini = (
    pl.from_pandas(reactions)
    .filter(
        (pl.col("cmpd_type") == "haloalkane (multiple)")
        | (pl.col("cmpd_type") == "haloalkane (separated)")
    )
    .select(["cmpd_name", "cmpd_smiles", "cmpd_type"])
)

gt = GT(reactions_mini, groupname_col="cmpd_type", rowname_col="cmpd_name").opt_row_striping()
```

The first several lines of the output is below:

Image

## Expected result
This is the expected first several rows:
Image

## Development environment

- Operating System: macOS, Windows
- **great_tables** Version: 0.17.0

juleswg23:
# Summary

The body generation needs a mapping of the rows to their location in the table because I believe `ordered_index` refers to a mapping before `groupname_col="[row_name]"` is applied.

As noted in the [issue](#700 ), the outcome for a table with groups in the body can be seen here:

```{python}
from great_tables import GT
from great_tables.data import reactions
import polars as pl

reactions_mini = (
    pl.from_pandas(reactions)
    .filter(
        (pl.col("cmpd_type") == "haloalkane (multiple)")
        | (pl.col("cmpd_type") == "haloalkane (separated)")
    )
    .select(["cmpd_name", "cmpd_smiles", "cmpd_type"])
)

gt = GT(reactions_mini, groupname_col="cmpd_type", rowname_col="cmpd_name").opt_row_striping()
```

image

Fixes #700 

# Checklist

### DIFF
--- a/great_tables/_utils_render_html.py
+++ b/great_tables/_utils_render_html.py
@@ -471,11 +471,12 @@ def create_body_component_h(data: GTData) -> str:
 
     ordered_index: list[tuple[int, GroupRowInfo]] = data._stub.group_indices_map()
 
-    for i, group_info in ordered_index:
+    for j, (i, group_info) in enumerate(ordered_index):
         # For table striping we want to add a striping CSS class to the even-numbered
         # rows in the rendered table; to target these rows, determine if `i` in the current
         # row render is an odd number
-        odd_i_row = i % 2 == 1
+
+        odd_j_row = j % 2 == 1
 
         body_cells: list[str] = []
 
@@ -530,7 +531,7 @@ def create_body_component_h(data: GTData) -> str:
 
                 _rowname_styles = [x for x in styles_row_label if x.rownum == i]
 
-                if table_stub_striped and odd_i_row:
+                if table_stub_striped and odd_j_row:
                     classes.append("gt_striped")
 
             else:
@@ -540,7 +541,7 @@ def create_body_component_h(data: GTData) -> str:
 
                 _rowname_styles = []
 
-                if table_body_striped and odd_i_row:
+                if table_body_striped and odd_j_row:
                     classes.append("gt_striped")
 
             # Ensure that `classes` becomes a space-separated string


rich-iannone:
@juleswg23 This is a great fix, thanks for making it happen! Could you do one last thing and add a small snapshot test? I put together this table code:

```python
from great_tables import GT
import polars as pl

pl_df = pl.DataFrame({"a": [1, 2, 3, 4, 5, 6], "b": ["a", "b", "a", "b", "a", "b"], "c": ["val"]* 6})

GT(pl_df, groupname_col="b", rowname_col="a").opt_row_striping()
```

Maybe that could be combined with `assert_rendered_body()`? Before this PR, only the rows in the 'b' group would get a stripe (all 'b' rows striped, in fact). With your PR, alternating rows (starting with the second row) get striped as one might expect, like this:

image

rich-iannone (changes requested):
Looks good! Just needs a test to prevent against any future regression.

tests/test_options.py
```diff
@@ -468,6 +469,19 @@ def test_opt_stylize_no_striping(snapshot):
...
+    pl_df = pl.DataFrame(
+        {"a": [1, 2, 3, 4, 5, 6], "b": ["a", "b", "a", "b", "a", "b"], "c": ["val"] * 6}
+    )
+
+    gt = GT(pl_df, groupname_col="b", rowname_col="a").opt_row_striping()
+
+    built = gt._build_data("html")
+    body = create_body_component_h(built)
+
+    assert snapshot == body
```
juleswg23, line 482: This assert follows in style from `test_opt_stylize_default` and `test_tab_options_striping_body_snap`.

https://github.com/posit-dev/great-tables/blob/91a40d312ea6298fb1bce0c448623ac16d6e4a63/tests/test_options.py#L479-L482

It may be worth considering calling `assert_rendered_body`, but if we make that change it should be to all rather than just this one.

PR #705, opened 2025-06-15
Allow list format in `_make_one_col_table`

tylerriccio33:
This is another effort to tackle https://github.com/posit-dev/pointblank/issues/132. In various parts of the library, a list is passed to the one column table maker, so if possible I intercept it and convert it to polars before getting passed to format_values.

I'm curious to see if the tests run, I had to monkeypatch it on the fly for development.

### DIFF
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -45,7 +45,7 @@ def _make_one_col_table(vals: X) -> GT:
     with contextlib.suppress(ImportError, ModuleNotFoundError):
         import polars as pl
 
-        vals: pl.Series = pl.Series(vals)
+        vals: pl.Series = pl.Series(values=vals, strict=False)
 
     df = to_frame(vals, name="x")
 
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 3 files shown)


tylerriccio33:
Added some additional comments and inline docs to explain what's going on. Coverage somehow went down despite hitting 100% of the diff, which is the failing ci check.

Let me know what you think :) needing pandas on some images just to produce reports has been plaguing some of my companies' prod code for a bit

machow:
If I understand, this change handles a list by:

* old behavior: try to import pandas and use that, fail otherwise.
* new behavior: try to import and use pandas, otherwise try polars.

I think a big challenge with the new behavior is that it could be surprising, since Great Tables behavior will change based on your installed libraries.

Users can already control which library gets used by passing in a pandas or polars series directly, so pointblank should be able to control what gets used by passing in a Polars or Pandas Series. Would pointblank passing in the series it wants to use resolve your issue? I think we want to lean towards explicit use of Pandas or Polars, and the list -> Pandas or fail, is a historical accident. If we did it again, we would likely just raise an error for a list (but are keeping the behavior for backwards compat).

edit: I would wrap calls [like this in pointblank](https://github.com/posit-dev/pointblank/blob/e6179656f4148ccd7b178c9749f0b5cab7b65288/pointblank/_datascan_utils.py#L22) in a Polars Series, so it's explicit that's what's going to happen. I think the fact that Great Tables round-trips through Series for doing the `vals.` calculations is super intense, and we probably don't want to encourage too much doing them on scalars, etc.. 😭

tylerriccio33:
I believe @rich-iannone  just made changes to some internal call(s) in pb that would bypass the list behavior, so it should be all good.

Issue #703, opened 2025-06-13
Indicate in topnav which section of docs user is on


machow:
When on a section of documentation, there's no indication on the topnav which section you are in. This makes it challenging to guess where you are when you click a link e.g. from the Guide to the API Reference.

![Image](https://github.com/user-attachments/assets/68ce5cb1-cb64-4c3a-8a05-fb09fe1da48a)

Let's check if there's a way to provide some emphasis. For example, by putting the orange background on the active section. (Or some other approach). It should still look reasonable when the current section is emphasized and a user is mousing over other section buttons.

PR #706, opened 2025-06-16
Add underline to active link in top nav

rich-iannone:
This PR is focused on making navigation through the docs site better by highlighting the current section the reader is in within the top navigation.

The proposed fix adds an underline for the current/active section:

image

And it works well with the orange background box that appears when hovering over the topnav items:

image

@machow lmk if this addresses the recent issue well. We can always tweak the appearance of the underline or even go with a different way to highlight the active section. 

Fixes: https://github.com/posit-dev/great-tables/issues/703

### DIFF
--- a/docs/styles.css
+++ b/docs/styles.css
@@ -132,6 +132,13 @@ p,h1,h2,h3,#toc-title,#toc-function-reference,.nav-link,.table {
 	}
 }
 
+.nav-link.active {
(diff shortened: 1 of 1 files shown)


machow:
Note that if you're in a subpage on e.g. Reference, the topnav is no longer marked, but this seems to be a quarto issue. We're checking in with them to see if there's a way to work around this!

PR #719, opened 2025-06-24
feat: remove pandas dependency in vals.fmt_integer

machow:
This PR removes the need for pandas in one of the vals functions, `vals.fmt_integer()`. It does this by removing the need for running the vals formatting through the GT class. Basically, the GT class was used, because it holds a DataFrame, which tells the formatter how to detect DataFrame lib specific NA values. Now, it just uses the Agnostic class, which was designed for this situation.

The process for converting a vals function to not use pandas is roughly this:

* Replace the data annotation with the protocol `data: PFrameData`.
* Pass in FramelessData() for data
* Include any of the data init steps in the format function (e.g. `fmt_integer()` code before it partials `fmt_integer_context()`)

I didn't add any extra tests, so depending on how we test these functions, there's a world where it might not work on pandas or polars (but it should pretty straightforwardly be fine?!)


great_tables/_formats_vals.py
```diff
@@ -254,23 +260,32 @@ def val_fmt_integer(
     ```
     """
 
-    gt_obj: GTData = _make_one_col_table(vals=x)
+    from great_tables._formats import _get_locale_sep_mark, _resolve_locale, fmt_integer_context
```
machow, line 263: oops, these should probably be moved to the top of the module

rich-iannone (approved):
LGTM!

PR #723, opened 2025-07-02
Docs blog images

juleswg23:
This PR adds a blog post on putting plots in tables.


pyproject.toml
```diff
@@ -71,13 +71,15 @@ dev-no-pandas = [
     "jupyter",
     "quartodoc>=0.11.0; python_version >= '3.9'",
     "griffe==0.38.1",
+    "plotnine",
```
machow, line 74: I think the last piece is putting plotnine in dev (line 64), since it induces a pandas dependency. The dev-no-pandas is used to run our pandas free tests.

docs/blog/plots-in-tables/index.qmd
```diff
@@ -0,0 +1,242 @@
+---
+title: "Adding Plots to Great Tables"
+html-table-processing: none
+author: Jules Walzer-Goldfeld and Michael Chow
+date: 2025-07-03
```
machow, line 5: do you mind adding `freeze: true` to the topmatter? Then, if you run quarto render, you should be able to add a folder called something like `_freeze/blog/plots-in-tables`.

machow (changes requested):
Yo, this looks so good! There's just one tiny dependency thing in pyproject.toml, and a small `freeze: true` bit. But I think the post reads really well!

jrycw:
Fantastic blog post — concise and helpful.

@juleswg23,  thanks for writing it! I definitely learned a lot.

Issue #122, opened 2024-01-16
Handle nested data in polars columns


machow:
Polars supports nested data---such as lists and structs---in columns of data.

Here's an example from the [polars guide](https://docs.pola.rs/user-guide/expressions/aggregation/#basic-aggregations)

```python
import polars as pl

url = "https://theunitedstates.io/congress-legislators/legislators-historical.csv"

dtypes = {
    "first_name": pl.Categorical,
    "gender": pl.Categorical,
    "type": pl.Categorical,
    "state": pl.Categorical,
    "party": pl.Categorical,
}

dataset = pl.read_csv(url, dtypes=dtypes).with_columns(
    pl.col("birthday").str.to_date(strict=False)
)

q = (
    dataset.lazy()
    .group_by("first_name")
    .agg(
        pl.count(),
        pl.col("gender"),
        pl.first("last_name"),
    )
    .sort("count", descending=True)
    .limit(5)
)

df = q.collect()
print(df)
```

```
┌────────────┬───────┬───────────────────┬───────────┐
│ first_name ┆ count ┆ gender            ┆ last_name │
│ ---        ┆ ---   ┆ ---               ┆ ---       │
│ cat        ┆ u32   ┆ list[cat]         ┆ str       │
╞════════════╪═══════╪═══════════════════╪═══════════╡
│ John       ┆ 1256  ┆ ["M", "M", … "M"] ┆ Walker    │
│ William    ┆ 1022  ┆ ["M", "M", … "M"] ┆ Few       │
│ James      ┆ 714   ┆ ["M", "M", … "M"] ┆ Armstrong │
│ Thomas     ┆ 454   ┆ ["M", "M", … "M"] ┆ Tucker    │
│ Charles    ┆ 439   ┆ ["M", "M", … "M"] ┆ Carroll   │
└────────────┴───────┴───────────────────┴───────────┘
```

Note that each entry in the gender column is a list of strings. However, I don't think Great Tables is set up to handle this situation.

### Current Behavior

```python
from great_tables import GT

GT(df).render("html")
```

```python
ComputeError: cannot cast List type (inner: 'Categorical(Some(global))', to: 'Utf8')
```

machow:
Note that List columns can have a schema like `List[List[int]]`, so we need some approach that can handle lists of lists (of lists) etc.. It appears that structs are  straightforward to coerce (although a struct can have a list in it).

FBruzzesi:
[...]

PR #729, opened 2025-07-18
perf: Avoid using polars map_elements

FBruzzesi:
# Summary

Avoid using `polars.Expr.map_elements` in `cast_frame_to_string` for performance reasons.

NA

# Checklist

## Additional info

I made a simple script to reproduce the speedup:

<details> <summary> full script </summary>

```py
import random
import time

import polars as pl
import polars.selectors as cs

def new_impl(frame: pl.DataFrame) -> pl.DataFrame:
    list_cols = [
        name for name, dtype in frame.collect_schema().items() if issubclass(dtype.base_type(), pl.List)
    ]

    return frame.with_columns(
        *[
            pl.concat_str(
                pl.lit("["),
                pl.col(c).cast(pl.List(pl.String())).list.join(", "),
                pl.lit("]")
            ).alias(c)
            for c in list_cols
        ],
        cs.all().exclude(list_cols).cast(pl.String()),
    )

def old_impl(frame: pl.DataFrame) -> pl.DataFrame:
    list_cols = [
        name for name, dtype in zip(frame.columns, frame.dtypes) if issubclass(dtype.base_type(), pl.List)
    ]

    return frame.with_columns(
        cs.by_name(list_cols).map_elements(lambda x: str(x.to_list()), return_dtype=pl.String),
        cs.all().exclude(list_cols).cast(pl.Utf8),
    )

SIZES = (10, 100, 1_000, 10_000, 100_000)

results: list[dict[str, float]] = []

for SIZE in SIZES:

    data = {
        "l1": [[random.randint(1, 10) for _ in range(random.randint(1, 10))] for _ in range(SIZE)],
        "l2": [[random.randint(1, 10) for _ in range(random.randint(1, 10))] for _ in range(SIZE)],
        "x": pl.int_range(SIZE, eager=True),
        "y": [random.random() for _ in range(SIZE)]
    }
    df = pl.DataFrame(data)

    start_new = time.perf_counter()
    _ = [new_impl(df) for _ in range(100)]
    end_new = time.perf_counter()

    start_old = time.perf_counter()
    _ = [old_impl(df) for _ in range(100)]
    end_old = time.perf_counter()

    results.append({"size": SIZE, "new": end_new-start_new, "old": end_old - start_old})

print(pl.DataFrame(results))
```

</details>

which in my local results in:

```terminal
shape: (5, 3)
┌────────┬──────────┬───────────┐
│ size   ┆ new      ┆ old       │
│ ---    ┆ ---      ┆ ---       │
│ i64    ┆ f64      ┆ f64       │
╞════════╪══════════╪═══════════╡
│ 10     ┆ 0.017877 ┆ 0.030849  │
│ 100    ┆ 0.018732 ┆ 0.043844  │
│ 1000   ┆ 0.034522 ┆ 0.186577  │
│ 10000  ┆ 0.195836 ┆ 1.602947  │
│ 100000 ┆ 1.607751 ┆ 15.813912 │
└────────┴──────────┴───────────┘
```

*forgot to mention that "new" and "old" columns represent the time taken to run the implementation.

Even with dataframe of size 10, the speedup is significant (in relative terms) and it gets larger and larger as the size increases.

### DIFF
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -516,7 +516,7 @@ def _(df: PdDataFrame):
 def _(df: PlDataFrame):
     import polars as pl
 
-    return df.clear().cast(pl.Utf8).clear(len(df))
+    return df.clear(len(df)).cast(pl.Utf8)
 
 
 @create_empty_frame.register
@@ -569,12 +569,19 @@ def _(df: PlDataFrame):
     import polars.selectors as cs
 
     list_cols = [
-        name for name, dtype in zip(df.columns, df.dtypes) if issubclass(dtype.base_type(), pl.List)
+        name for name, dtype in df.collect_schema().items() if issubclass(dtype.base_type(), pl.List)
     ]
 
     return df.with_columns(
-        cs.by_name(list_cols).map_elements(lambda x: str(x.to_list()), return_dtype=pl.String),
-        cs.all().exclude(list_cols).cast(pl.Utf8),
+        *[
+            pl.concat_str(
+                pl.lit("["),
+                pl.col(c).cast(pl.List(pl.String())).list.join(", "),
+                pl.lit("]")
+            ).alias(c)
+            for c in list_cols
+        ],
+        cs.all().exclude(list_cols).cast(pl.String()),
     )
 
 
--- a/great_tables/_tbl_data.py
+++ b/great_tables/_tbl_data.py
@@ -569,15 +569,15 @@ def _(df: PlDataFrame):
     import polars.selectors as cs
 
     list_cols = [
-        name for name, dtype in df.collect_schema().items() if issubclass(dtype.base_type(), pl.List)
+        name
+        for name, dtype in df.collect_schema().items()
+        if issubclass(dtype.base_type(), pl.List)
     ]
 
     return df.with_columns(
         *[
             pl.concat_str(
-                pl.lit("["),
-                pl.col(c).cast(pl.List(pl.String())).list.join(", "),
-                pl.lit("]")
+                pl.lit("["), pl.col(c).cast(pl.List(pl.String())).list.join(", "), pl.lit("]")
             ).alias(c)
             for c in list_cols
         ],


great_tables/_tbl_data.py
```diff
@@ -516,7 +516,7 @@ def _(df: PdDataFrame):
 def _(df: PlDataFrame):
     import polars as pl
 
-    return df.clear().cast(pl.Utf8).clear(len(df))
+    return df.clear(len(df)).cast(pl.Utf8)
```
FBruzzesi, line 519: This is unrelated but I wonder why a double "clear" is needed

great_tables/_tbl_data.py
```diff
@@ -569,12 +569,19 @@ def _(df: PlDataFrame):
     import polars.selectors as cs
 
     list_cols = [
-        name for name, dtype in zip(df.columns, df.dtypes) if issubclass(dtype.base_type(), pl.List)
+        name
+        for name, dtype in df.collect_schema().items()
```
FBruzzesi, line 573: If it ever comes to support LazyFrame's as well, this sets up the stage for it :)

great_tables/_tbl_data.py
```diff
@@ -569,12 +569,19 @@ def _(df: PlDataFrame):
...
+        if issubclass(dtype.base_type(), pl.List)
     ]
 
     return df.with_columns(
-        cs.by_name(list_cols).map_elements(lambda x: str(x.to_list()), return_dtype=pl.String),
-        cs.all().exclude(list_cols).cast(pl.Utf8),
+        *[
+            pl.concat_str(
+                pl.lit("["), pl.col(c).cast(pl.List(pl.String())).list.join(", "), pl.lit("]")
```
machow, line 580: I think this code might assume that the column can be cast to `pl.List(pl.String())`, but this will fail for other list dtypes?

for example:

```python
import polars as pl

df = pl.DataFrame({"x": [[[1, 2]], [[2, 3]]]})
df["x"].cast(pl.List(pl.String()))   # error: cannot cast List type (inner: 'Int64', to: 'String')
```

Manually converting each entry to a string using map_elements and Python is not fast, but hopefully at the scale of displaying a table, won't slow things down too much (and users can always coerce their data to strings before inputting).

In any event, if there's a way in Polars to stringify anything in the space of List dtypes I'm open to it! If not, there are a lot of other useful changes in this PR, so we could always revert this one piece, and merge everything else while working on the casting of lists to strings. (Down for whatever works best for you!)

FBruzzesi: aaah great catch! You are absolutely right!

> In any event, if there's a way in Polars to stringify anything in the space of List dtypes I'm open to it!

I left an idea here: https://github.com/posit-dev/great-tables/issues/122#issuecomment-3123489484 but I am not sure what's best. I will run some test to check :) 

I will go ahead and revert this change then!

PR #731, opened 2025-07-21
perf: Multiple nitpicks, some best practices and performances

FBruzzesi:
# Summary

So I might (or might not) have entered a rabbit hole, while the initial scope was much smaller (as you might tell from the branch name).

Some of the changes should lead to performance improvements, other are best practices.
Majority of them were spotted via [perflint](https://github.com/tonybaloney/perflint) or [refurb](https://github.com/dosisod/refurb).

There is no logic change, locally all tests pass without changing any of them.

---

If this is something of interest, feel free to help with the PR name, otherwise feel free to close it or cherry pick some of these changes.

NA

# Checklist

### DIFF
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -249,7 +249,7 @@ class ColInfo:
 
     @property
     def is_stub(self) -> bool:
-        return self.type in {ColInfoTypeEnum.stub, ColInfoTypeEnum.row_group}
+        return self.type in (ColInfoTypeEnum.stub, ColInfoTypeEnum.row_group)
 
     @property
     def defaulted_align(self) -> str:
@@ -314,7 +314,7 @@ class Boxhead(_Sequence[ColInfo]):
             )
 
         new_cols = []
-        stub_or_row_group = {ColInfoTypeEnum.stub, ColInfoTypeEnum.row_group}
+        stub_or_row_group = (ColInfoTypeEnum.stub, ColInfoTypeEnum.row_group)
         for col in self:
             # either set the col to be the new stub or row_group ----
             # note that this assumes col.var is always a string, so never equals None
--- a/great_tables/_gt_data.py
+++ b/great_tables/_gt_data.py
@@ -5,6 +5,7 @@ import re
 from collections.abc import Sequence
 from dataclasses import dataclass, field, replace
 from enum import Enum, auto
+from itertools import product
 from typing import TYPE_CHECKING, Any, Callable, Literal, Protocol, TypeVar, overload
 
 from typing_extensions import Self, TypeAlias, Union
@@ -248,7 +249,7 @@ class ColInfo:
 
     @property
     def is_stub(self) -> bool:
-        return self.type == ColInfoTypeEnum.stub or self.type == ColInfoTypeEnum.row_group
+        return self.type in {ColInfoTypeEnum.stub, ColInfoTypeEnum.row_group}
 
     @property
     def defaulted_align(self) -> str:
@@ -311,7 +312,9 @@ class Boxhead(_Sequence[ColInfo]):
                 "rowname_col and groupname_col may not be set to the same column. "
                 f"Received column name: `{rowname_col}`."
             )
+
         new_cols = []
+        stub_or_row_group = {ColInfoTypeEnum.stub, ColInfoTypeEnum.row_group}
         for col in self:
             # either set the col to be the new stub or row_group ----
             # note that this assumes col.var is always a string, so never equals None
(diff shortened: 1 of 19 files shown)


great_tables/_utils_render_latex.py
```diff
@@ -500,14 +489,14 @@ def derive_table_width_statement_l(data: GTData, use_longtable: bool) -> str:
         tw = float(tbl_width.strip("%"))
 
         side_width = (100 - tw) / 200
-        side_width = f"{side_width:.6f}".rstrip("0").rstrip(".")
+        side_width = f"{side_width:.6f}".rstrip(".0")
```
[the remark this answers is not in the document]
FBruzzesi: Good catch! You are right!
Repro: 

```py
"0.0".rstrip(".0")  # ''
"0.0".rstrip("0").rstrip(".")  # '0'
```

great_tables/_gt_data.py
```diff
@@ -390,40 +386,23 @@ def align_from_data(self, data: TblData) -> Self:
...
         align: list[str] = []
 
+        align_to_left = {"object", "utf8", "string"}
         for col_class in col_classes:
             # Ensure that `col_class` is lowercase
             col_class = str(col_class).lower()
 
             # Translate the column classes to an alignment value of 'left', 'right', or 'center'
-            if col_class == "character-numeric":
+            if col_class == "character-numeric" or _str_detect(col_class, r"int|uint|float|date"):
```
machow, line 395: @rich-iannone can you double check?

great_tables/_utils_nanoplots.py
```diff
@@ -935,15 +932,8 @@ def _generate_nanoplot(
...
-    for i in range(len(y_proportions)):
-        y_proportions[i] = safe_y_d + ((1 - y_proportions[i]) * data_y_height)
-
-    for i in range(len(x_proportions)):
-        x_proportions[i] = (data_x_width * x_proportions[i]) + safe_x_d
-
-    data_y_points = y_proportions
-    data_x_points = x_proportions
+    data_y_points = tuple(safe_y_d + ((1 - y_p) * data_y_height) for y_p in y_proportions)
+    data_x_points = tuple((data_x_width * x_p) + safe_x_d for x_p in x_proportions)
```
machow, line 936: It looks like x_proportions and y_proportions are mutated often in the original code. Can we restore this bit of code back?

Even if the tests pass, I want to make sure we do a bigger sweep first over the logic in this function, to be sure its defining / mutating / redefining things like `y_proportions` doesn't produce surprising edge cases (that might not be tested)

great_tables/_utils_render_latex.py
```diff
@@ -295,49 +295,38 @@ def create_columns_component_l(data: GTData) -> str:
         #     ids=True,
         #     omit_columns_row=True,
         # )
-
-        for i in range(len(spanners)):
-            spanners_row = spanners[i]
-
-            for k, v in spanners_row.items():
-                if v is None:
-                    spanners_row[k] = ""
```
machow, line 304: Just noting this mutation that was fixed to not mutate (hopefully isn't leaned on elsewhere in code)

machow (approved):
This LGTM -- thanks for going over the code and making these changes. I do think it would be easier to review if this was split into a couple PRs, but at the same time the changes all seemed very reasonable, so I don't want to slow your roll 😁 

@rich-iannone do you mind looking over the align regex code I marked, just to double check it still matches how you think about the auto align code?

https://github.com/posit-dev/great-tables/pull/731/files#r2234015522

great_tables/_utils_render_latex.py
FBruzzesi: Could you expand on this? I am not sure I understand what you mean

FBruzzesi:
Thanks for the review @machow

> I do think it would be easier to review if this was split into a couple PRs, 

I agree, I even mentioned it in the description:

> So I might (or might not) have entered a rabbit hole, while the initial scope was much smaller (as you might tell from the branch name).

I can split it into different types of improvements

- performance
- best practices
- nitpicks

As also mentioned, it's ok if some changes are not liked 

> but at the same time the changes all seemed very reasonable, so I don't want to slow your roll 😁

I don't want to slow down or disencourange reviews by opening PRs that are too large, I will try to keep that in mind and have more contained changes

great_tables/_gt_data.py
rich-iannone: There's no difference in auto-alignment behavior from this PR compared to `main`. So everything looks alright to me.

great_tables/_utils_nanoplots.py
machow: Thanks for taking the time to look into how these are used. Let's just get this PR merged, since it seems like it shouldn't change things!

machow:
Thanks for all your thoughtful work and response! These kinds of PRs are super helpful, so really appreciate you taking the time to improve so much of the code!

Issue #737, opened 2025-07-29
RFC: Adopt Narwhals for dataframe support


FBruzzesi:
# Description

Hey there 👋🏼

As you might have guessed from a few PRs I submitted, I started to get more comfortable with the codebase and spotted a few things while trying to see how Narwhals would look like in great-tables.

I have already something which is working in the current status: https://github.com/posit-dev/great-tables/compare/main...FBruzzesi:great-tables:feat/narwhalify-tbl-data-module

It's a bit of a messy branch, and I would definitely clean it up before opening a PR.

However I have a couple questions before finalizing the work:

1. Would this be something you are interested in and in scope in the first place? 😂
2. Would you prefer an incremental approach, or a end-to-end change/adoption?

For the incremental approach there is a "lazy" way of doing it, which is wrapping most functionalities in the `_tbl_data.py` module. However this would actually end up in many additional calls (`from_native`/`to_native`), while the final goal would be to convert once to narwhals when ingesting and keeping the data like that as long as possible.

On the other hand, there are some modules which can be "narwhalified" independently (such as `_data_color/base.py`).

Still the main advantages will be noticeable once the full changes are in (I know it's not a perfect metric, but the branch above has almost a -500 net lines of code and it avoids many conversions)

Last but not least, if you decide to move forward, I would suggest to support from narwhals 2.0: I recently fixed a couple of bugs we had, spotted while working on such branch.

Let me know your thoughts (I am also happy to have a call if needed)

machow:
Hey--thanks for looking at the code and making a bunch of useful changes! I definitely think Narwhals would be really useful to implement.

I would lean towards putting narwhals inside `_tbl_data.py` for now, and even making it an optional dependency for now(in the way Pandas and Polars are). That way, if someone passes a Narwhals frame to GT, we can trigger all the logic, but we also ensure the constrained frame API in `_tbl_data.py` remains in place.

IMO cutting 500 lines of code, but introducing the interaction of a full DataFrame API across Great Tables would be a net increase in complexity. Today, I could implement Great Tables over a custom class (call it "SimpleDataFrame") that is just a dictionary of lists, by registering 15-ish implementations in `_tbl_data.py`. But if Narwhals is baked into Great Tables outside that interface, I would have to implement Narwhals over SimpleDataFrame (a more complex task).

In any event, I think my big thoughts are:

* it would be awesome to have support for Narwhals frames.
[...]

PR #738, opened 2025-07-30
refactor: Narwhalify `data_color` function

FBruzzesi:
# Summary

Since I had it ready, I went ahead and narwhalified the `data_color` function. This means adding narwhals as a dependency.

In order to test functionalities I added a `series_constroctor` fixture that (depending on which libraries are installed) return constructors for pandas, pandas nullable, pandas pyarrow, polars and pyarrow series.

I also added a `assert_series_equal` in the test utils. This might need to change in the future, but this version in inspired by what we have in our test utils.

Finally, I added a commit (https://github.com/posit-dev/great-tables/pull/738/commits/ac1094cba2f557931dabe5648f691f0b6d6191db) in which I change all the tests in `test_data_color_utils.py` to use `@pytest.mark.parametrize`. I can revert it and re-scope it in another PR if it's something you would like to see! As you might have noticed already, I tend to lean in these kind of rabbit holes.

Related #737 

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ keywords = ["tables"]
 license.file = "LICENSE"
 authors = [
     { name = "Richard Iannone", email = "rich@posit.co" },
-    { name = "Michael Chow", email = "michael.chow@posit.co" }
+    { name = "Michael Chow", email = "michael.chow@posit.co" },
 ]
 dynamic = ["version"]
 classifiers = [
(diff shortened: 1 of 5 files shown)


pyproject.toml
FBruzzesi, line 1: Apart from adding narwhals as a dependency, all other changes were automatic?!

FBruzzesi:
I am closing this PR for now so that we can first better align following https://github.com/posit-dev/great-tables/issues/737#issuecomment-3137223572.

I will work a bit to refactor/parametrize the existing tests, which in cases of dataframe/series input might help to spot the current gaps.

PR #736, opened 2025-07-27
fix: Always access column names via `get_column_names` instead of `.columns`

FBruzzesi:
# Summary

Replaces `.columns` with `get_column_names` to return column names also for pyarrow (currently the column values are returned).

## Disclaimer

For pyarrow backed tbl_data, `data_color` the code would end up breaking a few lines down the line when performing the follow operation:

```py
column_vals = data_table[col][row_pos].to_list()
```

In fact, slicing with a list on a chunked array raises 

> *** TypeError: 'list' object cannot be interpreted as an integer

I wanted this PR to be atomic enough to solve _one_ issue. I can follow up on this other one

# Related GitHub Issues and PRs

- Closes #667 

# Checklist

- [x] I understand and agree to the [Code of Conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/).
- [x] I have followed the [Style Guide for Python Code](https://peps.python.org/pep-0008/) as best as possible for the submitted code.
- [x] I have added **pytest** unit tests for any new functionality.

PR #741, opened 2025-07-30
fix: Address slicing for pyarrow array in `data_color`

FBruzzesi:
# Summary

Follow up from the disclaimer section in https://github.com/posit-dev/great-tables/pull/736

<details> <summary> summary </summary>
> For pyarrow backed tbl_data, `data_color` the code would end up breaking a few lines down the line when performing the follow operation:

```py
column_vals = data_table[col][row_pos].to_list()
```

In fact, slicing with a list on a chunked array raises 

> *** TypeError: 'list' object cannot be interpreted as an integer

I wanted this PR to be atomic enough to solve _one_ issue. I can follow up on this other one

</details>

# Checklist

### DIFF
--- a/tests/data_color/__snapshots__/test_data_color.ambr
+++ b/tests/data_color/__snapshots__/test_data_color.ambr
@@ -123,6 +123,32 @@
   </tbody>
   '''
 # ---
+# name: test_data_color_autocolor_text_false[pyarrow]
+  '''
+  <tbody class="gt_table_body">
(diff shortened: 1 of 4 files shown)


tests/data_color/__snapshots__/test_data_color.ambr
FBruzzesi, line 1: I manually checked a few vs pandas/polars, but more eyes the better

PR #743, opened 2025-07-31
docs: Extending Great Tables

juleswg23:
This adds a get-started subpage which is a guide page on extensions, with gt-extras as an reference point.

The goal is to help people make their own personal "gt-extras" packages.

### DIFF
--- a/docs/get-started/extensions.qmd
+++ b/docs/get-started/extensions.qmd
@@ -4,7 +4,7 @@ format: html
 ---
 
 <!-- Intro -->
-The **Great Tables** package boasts a diverse range of methods for your creation of beautiful tables. With no shortage of tools, styles, formats and themes, you can produce a very wide range of tables. If you haven't yet, head over to the [examples gallery](../examples/index.qmd) for a flavor of the variety that is possible.
+The **Great Tables** package boasts a diverse range of methods for creating beautiful tables. With no shortage of tools, styles, formats and themes, you can produce a very wide range of tables. If you haven't yet, head over to the [examples gallery](../examples/index.qmd) for a flavor of the variety that is possible.
 
 But what happens when there's something you want to add to your table, but it's not directly supported by **Great Tables**? Let's say you have a very specific use case for your great table, so it's not necessarily broad enough to [contribute to the package](contributing.qmd). Or maybe it *is* generalizable, but it's beyond the scope of what you want to do. There are plenty of reasons to want a specific feature for your table that doesn't warrant a pull request.
 
@@ -78,7 +78,7 @@ def espn_theme(gt: GT) -> GT:
     )
 ```
 
-This is where wrapper functions really shine. Maybe you want a very standardized look, or you want all of your tables to mimic your favorite one. Take advantage of `~~GT.tab_options()` for your own [premade themes](./table-theme-premade.qmd), with styles from any of the [table theme options](./table-theme-options.qmd) of your choosing.
+This is where composed functions really shine. Maybe you want a very standardized look, or you want all of your tables to mimic your favorite one. Take advantage of `~~GT.tab_options()` for your own [premade themes](./table-theme-premade.qmd), with styles from any of the [table theme options](./table-theme-options.qmd) of your choosing.
 
 ### Extensions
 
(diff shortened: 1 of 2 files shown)


juleswg23:
I appreciate the guidance on structure. I've come back with a second draft, which I think (hope) might be closer to your vision. Happy to keep iterating.

I also decided that some of what I was describing wasn't really a wrapper, so I added a third kind of function. I tried to avoid making the page longer, instead repurposed one of the examples.

> My one suggestion is to add minimal examples illustrating each of Wrapper and Extension, so people get the gist quickly.

Should the minimal examples be so minimal so as to not even include Great Tables? A foo/bar kind of example?

docs/get-started/extensions.qmd
```diff
@@ -0,0 +1,170 @@
...
+- **Composing functions** combine multiple methods into a single function, allowing you to build more complex table logic or reusable patterns.
+- **Extensions**, on the other hand, add new functionality by combining multiple methods in creative ways, processing data outside the package, or generating custom content (like HTML or images) that gets integrated into your table.
+
+All approaches are valuable contributions to the **Great Tables** ecosystem, and the choice between them depends on whether you're making it easier to access existing tooling or creating entirely new capabilities.
+
+### Wrappers
+
+Let's wrap `~~GT.tab_style()` to perform a specific action supported by the general function. `~~GT.tab_style()` takes `locations=` and `style=` parameters, so let's assign those. Something this lightweight might be useful for someone who knows what effect they want, but doesn't want to investigate the entire space of options for styling.
+
+```{python}
```
machow, line 27: It took me a second to find where columns was used. I might cut the example down so it's quick to scan... (keeping two locations makes sense to me b/c it's sort of wrapping the user input...):

```{python}
def highlight_cols(gt: GT, columns: SelectExpr) -> GT:
    locations = [
        loc.body(columns=columns),
        loc.column_labels(columns=columns),
    ]

    # Make the wrapped tab_style call
    return gt.tab_style(
        style=style.fill(color="lightblue"),
        locations=locations,
    )
```

docs/get-started/extensions.qmd
```diff
@@ -0,0 +1,170 @@
...
+
+In fact, many of the functions in [**gt-extras**](https://posit-dev.github.io/gt-extras/) are wrappers. What you see above is very similar to the source code for the function [`gt_highlight_cols()`](https://posit-dev.github.io/gt-extras/reference/gt_highlight_cols), which is a wrapper function that doesn't add *new* functionality to **Great Tables**, instead it makes a common action much easier.
+
+### Composing Functions
+
+A little more complex than wrapping, we can also compose functions to perform several steps, such as applying multiple styles, formatting, or themes all at once.
+
+Here's how **gt-extras** implemented the [ESPN](https://posit-dev.github.io/gt-extras/reference/gt_theme_espn) theme:
+
+```{python}
```
machow, line 56: I'd make this something like....

```{python}
# | eval: False
def some_theme(gt: GT) -> GT:
    return (
        gt.opt_all_caps()
        .opt_table_font(font=google_font("Lato"), weight=400)
        .opt_row_striping()
    )
```

docs/get-started/extensions.qmd
```diff
@@ -0,0 +1,170 @@
...
+    )
+```
+
+This is where composed functions really shine. Maybe you want a very standardized look, or you want all of your tables to mimic your favorite one. Take advantage of `~~GT.tab_options()` for your own [premade themes](./table-theme-premade.qmd), with styles from any of the [table theme options](./table-theme-options.qmd) of your choosing.
+
+### Extensions
+
+Let's extend the post-processing step by building an HTML string containing two side-by-side tables. This function takes two separate `~~GT` objects, renders them as HTML using `~~GT.as_raw_html()`, and then wraps them in custom HTML structure to create a side-by-side layout, which would not be possible with **Great Tables** methods alone.
+
+```{python}
```
machow, line 87: Here, I wonder if showing defining a very simple formatter could be useful? idk if this is a bit too complex, but something like this...?

```python
def fmt_threshold(gt, columns, threshold: float):
    """Format values to 0 if they are under threshold.""""

    def _truncate(val: float)
        return "0" if val < threshold else str(val)

    return gt.fmt(_truncate, columns)
```

edit: I just noticed the extending sections do formatters. So in this case maybe the example above, and a stripped down version of the 2 gts combined example?:

```{python}
# | eval: False
def double_table(gt1: GT, gt2: GT) -> str:
    table_1_html = gt1.as_raw_html()
    table_2_html = gt2.as_raw_html()
    return f"""<div>{table_1_html}</div><div>{table_2_html}</div>"""
```

This might help give a sense for the range of activities that could happen here? I think you want people to get a sense for the activity, so when they see the `gt_two_column_layout()` they have a rough sense for how it made things happen.

machow (changes requested):
These examples look really useful! I wrote some comments with stripped down examples before seeing your question, so am including here 😅

juleswg23:
Thanks for the nudge in the right direction! I've replaced the examples, and then changed the language to match the style of using action words.

The very last commit includes the `fmt_threshold` example – I wasn't sure if it made more sense to leave it in or out, so I can revert the last commit if we want it out.

I feel a little weird with the title as "Extending Great Tables" when extending is one of the three approaches. Maybe "Great Tables Add-ons" or "Enhancing Great Tables"?

Issue #711, opened 2025-06-17
Alpha value gets masked by data_color()


juleswg23:
The following call to `data_color()` sets `alpha = 0.2`, but the outcome is identical to setting `alpha = 1`.

``` {python}
from great_tables import GT, exibble

exibble_sm = exibble[["num", "fctr", "row", "group", "currency"]]
GT(exibble_sm).data_color(columns="row", alpha=0.2)
```

<img width="308" alt="Image" src="https://github.com/user-attachments/assets/3432da40-9261-4ead-b6a2-73adfcd1ce5d" />

```{python}
GT(exibble_sm).data_color(columns="row", alpha=1)
```

<img width="307" alt="Image" src="https://github.com/user-attachments/assets/bb9f562d-02ac-4ec8-86a8-86aecd12b01b" />

juleswg23:
`alpha` is first applied to each color in the palette here:

https://github.com/posit-dev/great-tables/blob/0b4c5ca3954706314cb34f2fff36cf6085008d1b/great_tables/_data_color/base.py#L199

But later in the function the `GradientPalette` call creates a list of `RGBColor`s, which drops the `alpha` value.

https://github.com/posit-dev/great-tables/blob/0b4c5ca3954706314cb34f2fff36cf6085008d1b/great_tables/_data_color/base.py#L261

A possible solution would be to apply _html_color later (at line 268), though the type-checker sees `color_vals` as a `list[str | None]`, which I am pretty sure is no longer the case after line 267.

```{python}
if alpha is not None:
    color_vals = _html_color(colors=color_vals, alpha=alpha)
```

https://github.com/posit-dev/great-tables/blob/0b4c5ca3954706314cb34f2fff36cf6085008d1b/great_tables/_data_color/base.py#L266-L284

juleswg23:
As a sidenote, this introduces a bug in applying the `autocolor` flag, because a more transparent value for any color will be lighter, and thus produce a lighter background.

We could set the default `autocolor` to be false when `alpha < 1` (though this has it's own drawbacks), or figure out how to factor in `alpha` to `_ideal_fgnd_color()`.

<img width="322" alt="Image" src="https://github.com/user-attachments/assets/16090335-f74b-4e71-87c6-172eb584fc63" />

PR #747, opened 2025-08-01
Fix: `_ideal_fgnd_color()` takes alpha into account

### LINKED ISSUE #746: bug: applying alpha to background colors produces non-contrasting color
## Prework

## Description

Applying a faint color (via alpha) to any background color option produces non-contrasting font color by virtue of `_ideal_fgnd_color()` ignoring the RGBA alpha value.

## Reproducible example

```python
import pandas as pd
from great_tables import GT

df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})

GT(df).tab_options(table_background_color="#FF000008")
```

The following is returned:

Image

## Expected result

We would have liked to see black text:

Image

## Development environment

- Operating System: macOS
- **great_tables** Version: 0.18.0

juleswg23:
# Summary

See related issue. This PR relies on the fact that background that are not fully opaque are always blended against a white background. I think the way tables set final cell and other background colors is consistent with this fact, but I am open to being corrected and looking for a workaround if that isn't the case.

The PR effectively replaces `_remove_alpha()` with `_alpha_composite_with_white()`.

- Ref: Closes #746 

# Checklist

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -509,21 +509,6 @@ def _add_alpha(colors: list[str], alpha: int | float) -> list[str]:
     return colors
 
 
-def _remove_alpha(colors: list[str]) -> list[str]:
-    # Loop through the colors and remove the alpha value from each one
-    for i in range(len(colors)):
-        color = colors[i]
-        # If the color value is already in the `#RRGGBB` format, then we need to add the
-        # alpha value to it before removing the alpha value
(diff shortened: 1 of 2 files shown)


juleswg23:
Separately, this sets the stage for a fix to https://github.com/posit-dev/great-tables/issues/711#issuecomment-2981026310. If the change in this PR is accepted (or a similar one) I can make the subsequent change to close #711.

PR #748, opened 2025-08-02
Create dedicated CSS group

tylerriccio33:
This is a 2-fold PR:

1. I created a group for css-inline package called css. This would be helpful because I often have to use the css inlining feature in ci, and i must document it as a dependency. However, carrying around selenium can be cumbersome in production, since it has some tough requirements of it's own.
2. Reworked the `_try_import` function call out the dependency group needed to install said dependency instead of the generic pip install command. This would encourage users to follow best practices and add the correct group instead of one-off installing it.

I did these things which are not crucial but development quality of life. Feel free to strike them from the commit!
- I added pre-commit commands in the Makefile so users can more easily install it.
- Added uv.lock to the git ignore so it won't get tracked by users developing/contributing with uv.
- Used uv to run the tests in the Makefile instead of generic `pytest ...`. The old version would require users to activate the environment. Not sure if you want to keep this depending on your philosophy!

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -293,8 +293,8 @@ def data_color(
 
 
 def _ideal_fgnd_color(bgnd_color: str, light: str = "#FFFFFF", dark: str = "#000000") -> str:
-    # Remove alpha value from hexadecimal color value in `bgnd_color=`
-    bgnd_color = _remove_alpha(colors=[bgnd_color])[0]
+    # Compose alpha value from hexadecimal color value in `bgnd_color=`
+    bgnd_color = _alpha_composite_with_white(bgnd_color)
 
     contrast_dark = _get_wcag_contrast_ratio(color_1=dark, color_2=bgnd_color)
     contrast_light = _get_wcag_contrast_ratio(color_1=light, color_2=bgnd_color)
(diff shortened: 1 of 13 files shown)


rich-iannone:
@tylerriccio33 we discussed this PR earlier today and came to the conclusion that splitting out css-inline into its own group would probably make things more complicated for the average user without a clear benefit. I can definitely understand that installing an unneeded selenium dependency makes no sense, but couldn't the alternative be installing css-inline separately (i.e., independent of installing `great-tables`)?

tylerriccio33:
Yea completely get that. The problem with having a separate install of css-inline is it usually fails CI, since it's not documented in the dependency tree and it's never imported, so the linters think it's an extra dependency. It's easy to override so not a big deal!

tylerriccio33:
I would argue though the `_try_import` refactor is worth a merge, since it specifically calls out what dep group you have to add. If you want to proceed with that I can edit the pr.

rich-iannone:
> I would argue though the `_try_import` refactor is worth a merge, since it specifically calls out what dep group you have to add.

But if we’re not adding the second optional dependency group `css` then don’t we only have the `extra` group?

What does look pretty good in this PR are the additions to the Makefile. Would you be willing to close this PR and make a separate one with just the Makefile update?

tylerriccio33:
so even though it's just the extra, i still think it's beneficial to include that in the err message instead of it just saying 'run pip install <pkg>'. In the change it would be 'run pip install[extra]' which is the more accurate install command.

PR #749, opened 2025-08-03
Remove numpy from dependencies

tylerriccio33:
Hi maintainers, I'm seeing if numpy could be removed from the dependencies in order to make the library more agnostic. I know this was a a discussion some time back but wasn't prioritized.

Personally, I'd love to drop numpy from some production images at my work.

I don't believe this current pr has any issues, but it's tough to tell since you're obviously testing it with a full dependency graph available. For this reason I'm marking it as a draft until I'm more confident!

Would love feedback/criticism

### DIFF
--- a/great_tables/_data_color/base.py
+++ b/great_tables/_data_color/base.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-import numpy as np
 from typing_extensions import TypeAlias
 
 from great_tables._locations import RowSelectExpr, resolve_cols_c, resolve_rows_i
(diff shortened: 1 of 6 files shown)


tylerriccio33:
This now passes CI and I have no reason to believe this isn't functional :)

tylerriccio33:
This should be ready to rock

rich-iannone:
@machow I've checked the new nanoplots generated in the docs of the throwaway PR (https://github.com/posit-dev/great-tables/pull/751) based on this PR and it's a perfect match to what's in `main` (no regressions). There are tests in place for `_normalize_to_dict()` and `_normalize_vals()`; and the tests for nanoplots are generally pretty comprehensive. To me, this looks great!

PR #755, opened 2025-08-08
Fix: `fmt()` methods applied to `groupname_col`

juleswg23:
# Summary

This PR ensures data for group labels in groupname_col are fetched from the formatted table values, rather than pure text. The example code in the linked issue now outputs:

image

One caveat is I don't know what the expected behavior should be with None/NA values in the groupname_col. This PR is specifically designed with the intention of not modifying the rendering of None/NA values.

# Related GitHub Issues and PRs

Closes #236 

# Checklist

- [x] I understand and agree to the [Code of Conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/).
- [x] I have followed the [Style Guide for Python Code](https://peps.python.org/pep-0008/) as best as possible for the submitted code.
- [ ] I have added **pytest** unit tests for any new functionality.

PR #756, opened 2025-08-08
Fix: allow seting `groupname_col` without `rowname_col`

### LINKED ISSUE #106: Setting groupname_col drops the group col, unless you set rowname_col
```python

GT(exibble, groupname_col="group")
```

Here is the output (via vs code notebook)

image

```python
GT(exibble, groupname_col="group", rowname_col="char")
```

image

juleswg23:
# Summary

This PR generates the group labels even when the row_stub_column is not present.

The new output from the sample code in the issue is:

```python

GT(exibble, groupname_col="group")
```

image

This PR also renames variables for clarity, including `has_two_col_stub` -> `has_group_stub_column` because in this PR that variable can be set to True even when the row_stub_column is not present.

The variable name changes will have downstream affects on the draft PRs #754 and #755, so I making a note here to ensure if this is accepted that those drafts are revised.

Closes #106 

# Checklist

- [x] **I have added **pytest** unit tests for any new functionality.**

### DIFF
--- a/tests/__snapshots__/test_utils_render_html.ambr
+++ b/tests/__snapshots__/test_utils_render_html.ambr
@@ -35,6 +35,26 @@
   </tbody>
   '''
 # ---
+# name: test_groupname_with_no_rowname
+  '''
+  <tbody class="gt_table_body">
+    <tr class="gt_group_heading_row">
+      <th class="gt_group_heading" colspan="2">A</th>
+    </tr>
+    <tr>
+      <td class="gt_row gt_right">0</td>
+      <td class="gt_row gt_right">22</td>
+    </tr>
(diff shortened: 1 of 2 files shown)


great_tables/_utils_render_html.py
```diff
@@ -471,7 +471,7 @@ def create_body_component_h(data: GTData) -> str:
         body_cells: list[str] = []
 
         # Create table row specifically for group (if applicable)
-        if has_stub_column and has_groups and not has_two_col_stub:
+        if has_groups and not has_group_stub_column:
```
juleswg23, line 474: Removing `has_stub_column` is the only change in the logic. The rest of the modifications are variable name changes

PR #767, opened 2025-08-21
Pre commit makefile targets

tylerriccio33:
This adds pre-commit install and run targets to the makefile per #748 .

This also adds uv.lock to the git ignore file which is now standard, per python's own git ignore file.


tylerriccio33:
Updated to keep pace with main

tylerriccio33:
@rich-iannone Any way this can get merged? Super simple qualify of life imo.

Issue #174, opened 2024-02-06
Implement opt_css


machow:
https://gt.rstudio.com/reference/opt_css.html

PR #775, opened 2025-10-01
feat: add the `.opt_css()` method

rich-iannone:
This PR adds the `.opt_css()` method. It allows users to add custom CSS to HTML-output tables. This `opt_*()` method builds on the existing functionality of the `tab_options(table_additional_css=)` option by allowing users to more easily and safely append CSS rulesets to that Options parameter. 

```{python}
import polars as pl

exibble_mini = pl.from_pandas(exibble).select(["num", "currency"])

(
    GT(exibble_mini, id="one")
    .fmt_currency(columns="currency", currency="HKD")
    .fmt_scientific(columns="num")
    .opt_css(
        css='''
        #one .gt_table {
          background-color: skyblue;
        }
        #one .gt_row {
          padding: 20px 30px;
        }
        #one .gt_col_heading {
          text-align: center !important;
        }
        '''
    )
)
```

image

Notice in that rendered table that: (1) body cells get a 'skyblue' background color, (2) the padding of body cells is increased horizontally and vertically, and (3) column-label text is now centered instead of being right-aligned.

Fixes: https://github.com/posit-dev/great-tables/issues/174

### DIFF
--- a/great_tables/_options.py
+++ b/great_tables/_options.py
@@ -22,7 +22,7 @@ def tab_options(
     table_margin_left: str | None = None,
     table_margin_right: str | None = None,
     table_background_color: str | None = None,
-    table_additional_css: list[str] | None = None,
+    table_additional_css: str | list[str] | None = None,
     table_font_names: str | list[str] | None = None,
     table_font_size: str | None = None,
     table_font_weight: str | int | float | None = None,
@@ -552,6 +552,11 @@ def tab_options(
     if "table_font_names" in modified_args:
         if isinstance(modified_args["table_font_names"], str):
             modified_args["table_font_names"] = [modified_args["table_font_names"]]
+
+    # - `table_additional_css` should be a list but if given as a string, ensure it is list
+    if "table_additional_css" in modified_args:
+        if isinstance(modified_args["table_additional_css"], str):
+            modified_args["table_additional_css"] = [modified_args["table_additional_css"]]
 
     new_options_info = {
         k: replace(getattr(self._options, k), value=v) for k, v in modified_args.items()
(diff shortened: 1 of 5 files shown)


machow:
### Code review (updated)

**MICHAEL NOTES:**

* None of these issues seems too bad

Found 5 issues (note: all scored below 80% confidence threshold, so these are minor):

1. **Test functions test multiple behaviors** (CLAUDE.md says "Test one behavior per case") - Score: 75

   `test_opt_css_basic()` tests both basic CSS addition AND chaining; `test_opt_css_duplicate_handling()` tests both default duplicate prevention AND allowing duplicates.

   https://github.com/posit-dev/great-tables/blob/9fd4016842a6d2b47985954f88ddb37f83ef67f5/tests/test_options.py#L553-L579

2. **Tests don't use kwargs dict pattern** (CLAUDE.md says "Use kwargs dicts, only specify non-defaults") - Score: 75

   Tests use direct method calls instead of `@pytest.mark.parametrize` with kwargs dicts pattern.

   https://github.com/posit-dev/great-tables/blob/9fd4016842a6d2b47985954f88ddb37f83ef67f5/tests/test_options.py#L553-L661

3. **Empty string not filtered in `tab_options()` normalization** - Score: 50

   When `tab_options(table_additional_css="")` is called, it becomes `[""]` instead of `[]`. This is a rare edge case with minimal impact (extra newline in CSS output).

   https://github.com/posit-dev/great-tables/blob/9fd4016842a6d2b47985954f88ddb37f83ef67f5/great_tables/_options.py#L556-L559

4. **Comment capitalization inconsistency** - Score: 50

   One comment starts with lowercase (`# if CSS already exists...`) while all others in the function start with capitals.

   https://github.com/posit-dev/great-tables/blob/9fd4016842a6d2b47985954f88ddb37f83ef67f5/great_tables/_options.py#L1519

5. **Missing test coverage for edge cases** - Score: 75

   No explicit tests for: `opt_css(css="", add=False)` clearing existing CSS, or `opt_css(css="")` preserving existing CSS.

   https://github.com/posit-dev/great-tables/blob/9fd4016842a6d2b47985954f88ddb37f83ef67f5/tests/test_options.py#L599-L610

---

All issues are below the 80% confidence threshold for reporting as definite problems. The implementation is solid overall.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rich-iannone:
> ### Code review (updated)
> **MICHAEL NOTES:**
> 
> * None of these issues seems too bad
> 
> Found 5 issues (note: all scored below 80% confidence threshold, so these are minor):
> 
> 1. **Test functions test multiple behaviors** (CLAUDE.md says "Test one behavior per case") - Score: 75
>    `test_opt_css_basic()` tests both basic CSS addition AND chaining; `test_opt_css_duplicate_handling()` tests both default duplicate prevention AND allowing duplicates.

Fixed by splitting the tests:
- `test_opt_css_basic()` → split into `test_opt_css_add_single_rule()` and `test_opt_css_chaining()`
- `test_opt_css_duplicate_handling()` → split into `test_opt_css_duplicate_prevented()` and `test_opt_css_allow_duplicates()`

Each test now focuses on a single behavior.

> 2. **Tests don't use kwargs dict pattern** (CLAUDE.md says "Use kwargs dicts, only specify non-defaults") - Score: 75
>    Tests use direct method calls instead of `@pytest.mark.parametrize` with kwargs dicts pattern.

Reviewed this suggestion but opted to keep the current approach. The existing tests in `test_options.py` use direct method calls rather than `@pytest.mark.parametrize` with kwargs dicts. Additionally, several `opt_css` tests involve method chaining (e.g., `.opt_css().opt_css()`) which doesn't fit well with the parametrize pattern. Keeping the current style maintains consistency with the rest of the test file.

> 3. **Empty string not filtered in `tab_options()` normalization** - Score: 50
>    When `tab_options(table_additional_css="")` is called, it becomes `[""]` instead of `[]`. This is a rare edge case with minimal impact (extra newline in CSS output).

Fixed by updating the `table_additional_css` normalization in `tab_options()` to strip whitespace and filter empty strings.
Now `tab_options(table_additional_css="")` results in `[]` instead of `[""]`. Added `test_tab_options_empty_string_css()` to verify this behavior.

> 4. **Comment capitalization inconsistency** - Score: 50
>    One comment starts with lowercase (`# if CSS already exists...`) while all others in the function start with capitals.

Fixed by capitalizing the comment.

> 5. **Missing test coverage for edge cases** - Score: 75
>    No explicit tests for: `opt_css(css="", add=False)` clearing existing CSS, or `opt_css(css="")` preserving existing CSS.

Added explicit tests for the edge cases:
- `test_opt_css_empty_string_preserves_existing()` - verifies that `opt_css(css="")` preserves existing CSS when `add=True` (default)
- `test_opt_css_empty_string_clears_with_add_false()` - verifies that `opt_css(css="", add=False)` clears existing CSS

rich-iannone:
The snapshot failures are due to a major update in Pandas (v3.0.0). I've created a new issue (https://github.com/posit-dev/great-tables/issues/809) that identifies the problem and how to address it.

machow:
@rich-iannone can you merge main in so snapshots are fixed, and then if it's ready for review ping me

machow:
Can you add a small note to the docstring contrasting with style.css()? I think we can loop back and pull into the guide later

rich-iannone:
> Can you add a small note to the docstring contrasting with style.css()? I think we can loop back and pull into the guide later

I've added a paragraph explaining the differences between the approaches in the `opt_css()` docstring. Thanks for the suggestion!

great_tables/_options.py
```diff
@@ -1445,6 +1445,12 @@ def opt_css(
     the compiled CSS that Great Tables generates automatically when the object is transformed to an
     HTML output table.
 
+    This method differs from `style.css()` (used with `tab_style()`). Whereas `style.css()` applies
```
machow: What about something like this?

> If you want to set css styles on a specific table location, use `tab_style()` with `style.css()` instead.

That way, people can navigate there and get the gist

Issue #171, opened 2024-02-06
Implement cols_merge*() family of methods


machow:
`cols_merge()` and friends let you combine the content of two columns into one, with special templating syntax. E.g. if I have a mean column with value "X" and variance columns "Y.Z", then I might want to get `X (Y.Z)`.

https://gt.rstudio.com/reference/cols_merge.html

machow:
* https://gt.rstudio.com/reference/fmt_url.html
* https://gt.rstudio.com/articles/case-study-clinical-tables.html
* pharma format sdtm: https://www.cdisc.org/standards/foundational/sdtm

### Combining title, original name, and link

Note that when original name is NA, `(<original_name>)` is excluded.

<img width="772" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/b67be7cb-a088-43e7-8f31-19d90821a740">

### Additional info in parentheses (e.g. stddev), ranges, etc..

<img width="1035" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/b84a93b6-f450-4b23-9c8b-474602379d61">

### More parentheses

<img width="1006" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/cd218955-27e8-4456-8d1f-fc2690dabfbe">

<img width="943" alt="image" src="https://github.com/posit-dev/great-tables/assets/2574498/fdc54ec0-04e5-4bf0-9351-4ac1b73c43e1">

PR #780, opened 2025-10-08
feat: add the `.cols_merge()` method

rich-iannone:
This PR adds the `.cols_merge()` method. It allows for selective concatenation of cell values to a single cell, automatically hiding other columns involved in the merging process (could be set to not hide, however). I'll provide two examples here of how this new method works in practice.

In the first example, we merge the `open` & `close` columns together, and, the `low` & `high` columns (putting an em dash between both). The `close` and `high` columns are hidden, and we take the opportunity to rename `open` and `low`:

```python
from great_tables import GT
from great_tables.data import sp500
import polars as pl

sp500_mini = (
    pl.from_pandas(sp500)
    .slice(49, 6)
    .select("open", "close", "low", "high")
)

(
    GT(sp500_mini)
    .fmt_number(
        columns=["open", "close", "low", "high"],
        decimals=2,
        use_seps=False
    )
    .cols_merge(columns=["open", "close"], pattern="{1}&mdash;{2}")
    .cols_merge(columns=["low", "high"], pattern="{1}&mdash;{2}")
    .cols_label(open="open/close", low="low/high")
)
```

image

This next one is a bit more involved in that we account for missing values in merged cells. We use the `.cols_merge()` twice to merge together the: (1) `trq` and `trq_rpm` columns, and (2) `mpg_c` & `mpg_h` columns. Given the presence of missing values we can use patterns that include `<<`/`>>` to create conditional text spans, excluding results where any of the merged columns have missing values.

```python
from great_tables import GT
import polars as pl
import polars.selectors as cs

gtcars_pl = (
    pl.from_pandas(gtcars)
    .filter(pl.col("year") == 2017)
    .select(["mfr", "model", "trq", "trq_rpm", "mpg_c", "mpg_h"])
)

(
    GT(gtcars_pl)
    .fmt_integer(columns=[cs.starts_with("trq"), cs.starts_with("mpg")])
    .cols_merge(columns=["trq", "trq_rpm"], pattern="{1}<< ({2} rpm)>>")
    .cols_merge(columns=["mpg_c", "mpg_h"], pattern="<<{1} city<</{2} hwy>>>>")
    .cols_label(mfr="Manufacturer", model="Car Model", trq="Torque", mpg_c="MPG")
)
```

image

Note that we formatted the floating point values to be displayed as integers and that the formatted values are present in the merged outputs. The ability to handle formatting separately from the merging (along with the NA handling) makes producing highly customized cell outputs really easy to iterate on. 

This PR partially addresses https://github.com/posit-dev/great-tables/issues/171. The remaining `.cols_merge*()` methods will be added in a later PR.


great_tables/gt.py
```diff
@@ -314,6 +320,98 @@ def _render_formats(self, context: str) -> Self:
...
+                merged_value = _process_col_merge_pattern(
+                    pattern=pattern, col_values=col_values, col_is_missing=col_is_missing
+                )
+
+                # Set the merged value in the target column
+                result = _set_cell(new_body.body, row_idx, target_column, merged_value)
+
+                # For Pandas and Polars, _set_cell() modifies in place and returns None but
+                # for PyArrow, _set_cell() returns a new table
+                if result is not None:
```
machow: Does this custom handling of backends happen in other places outside of `_tbl_data.py`?

great_tables/gt.py
```diff
@@ -314,6 +320,98 @@ def _render_formats(self, context: str) -> Self:
...
+            # Pattern should always be set by `.cols_merge()`, but check here just in case
+            if pattern is None:  # pragma: no cover
+                raise ValueError("Pattern must be provided for column merge operations.")
+
+            # Validate that pattern references are valid
+            pattern_cols = _extract_pattern_columns(pattern)
+
+            # With each column reference in the pattern, check that it is valid
+            for col_ref in pattern_cols:
+                # The pattern syntax uses 1-based indexing so adjust here
```
machow: 1-based indexing seems like a bad idea (but maybe it happens in other places in Python?)

machow (changes requested):
I didn't read any of the code in utils yet, but the main piece I would vote to refactor a bit is the big method on the GT class, and avoiding full GT renders for the unit tests (see comments for suggestions).

rich-iannone:
The review comments are now addressed with a few code changes. Here's a summary of what was done:

### 1. **Moved `_perform_col_merge` out of the GT class** (great_tables/_cols_merge.py)
   - Created a new module `great_tables/_cols_merge.py` containing:
     - `perform_col_merge()` - the main function that replaces the GT method
     - `_apply_single_col_merge()` - handles a single merge operation  
     - `_collect_row_values()` - collects values and missing status for a row

### 2. **Refactored `ColMergeInfo`** (great_tables/_gt_data.py)
   - Added `pattern_columns` property to extract column references from the pattern
   - Added `validate_pattern()` method to validate pattern references
   - Added `merge_values()` method that runs `_process_col_merge_pattern` on provided values
   - Added comprehensive docstrings with examples
   - Note on 1-based indexing: Added documentation explaining this is for **consistency with R's gt package** pattern syntax

### 3. **Updated tests** (tests/test_cols_merge.py)
   - Split into two test classes:
     - `TestColMergeInfoUnit` - 14 unit tests that test `ColMergeInfo` methods directly **without full GT renders**
     - `TestColsMergeIntegration` - 16 integration tests that verify end-to-end behavior
   - The unit tests cover pattern validation, pattern extraction, and merge_values with various patterns including conditional sections

### 4. **Backend handling comment**
   - The pattern `if result is not None: body.body = result` is already used in `_gt_data.py` in the `Body.render_formats()` method - this is consistent with existing code

### 5. **Updated gt.py**
   - Removed the large `_perform_col_merge` method from the GT class
   - Now calls `perform_col_merge(built)` from the new `_cols_merge` module in `_build_data()`
   - Kept `_get_cell` import as it's still used by `_get_column_of_values`

All 1378 tests pass with the refactored code.

rich-iannone:
This is the roborev review (performed *after* the above changes were made):

------

## Review Complete

I've checked the commit for bugs, security issues, testing gaps, regressions, and code quality concerns. **No issues
found.**

### Summary

This is a **clean refactoring commit** that extracts column-merge logic from `gt.py` into a new dedicated module
`_cols_merge.py`. The changes improve code organization and maintainability:

**What changed:**
1. Created `_cols_merge.py` with functions: `perform_col_merge()`, `_apply_single_col_merge()`, and
`_collect_row_values()`
2. Enhanced `ColMergeInfo` class with methods: `pattern_columns` property, `validate_pattern()`, and `merge_values()`
3. Removed `GT._perform_col_merge()` method and replaced call with standalone function
4. Reorganized tests into unit tests (`TestColMergeInfoUnit`) and integration tests (`TestColsMergeIntegration`)

**Why it's safe:**
- Logic is directly extracted with no algorithmic changes
- All original test cases are preserved
- 14 new unit tests added for `ColMergeInfo` methods
- Better separation of concerns (data class methods handle their own validation/processing)

machow:
RE 1-based indexing, it seems like these are the core dynamics that will be at play in Great Tables:

* Polars selectors: 0-based
* Python string formatting: 0-based (e.g. `"{0}".format(1)`)
* cols_merge: 1-based

There may be other examples of 1-based things in Python data libraries though? If so, it might be helpful for us to list some out

machow (changes requested):
It seems like ColMergeInfo is very close to being self contained, but the logic for calculating `cols_values`, `col_is_missing` is outside of it. WDYT of something like below?

I think the big benefit it can be run outside of a loop over columns

```python

# class now self-contained when called over scalar values----

class ColMergeInfo:

    ...

    def merge(self, *values, tbl_data: TblData | None = None) -> str:
        """Merge values according to pattern.

        Parameters
        ----------
        *values
            Raw values to merge.
        tbl_data
            Optional table data for backend-specific NA detection.
            If None, uses generic NA detection.
        """
        from ._tbl_data import is_na, Agnostic

        check_data = tbl_data if tbl_data is not None else Agnostic()

        col_values = {str(i + 1): str(v) for i, v in enumerate(values)}
        col_is_missing = {str(i + 1): is_na(check_data, v) for i, v in enumerate(values)}

        return _process_col_merge_pattern(self.pattern, col_values, col_is_missing)


# outward user scalar interface (for quick experimentation) ----

def merge_pattern(pattern: str, *values) -> str:
    """Merge values into a pattern string.

    Parameters
    ----------
    pattern
        Pattern with {1}, {2}, etc. for positional values.
        Use <<...>> for conditional sections that are omitted
        if any referenced value inside is missing.
    *values
        Values to substitute (1-indexed in pattern).

    Examples
    --------
    >>> merge_pattern("{1} {2}", "John", "Doe")
    'John Doe'
    >>> merge_pattern("{1}<< ({2})>>", "John", None)
    'John'
    """
    from ._gt_data import ColMergeInfo

    info = ColMergeInfo(pattern=pattern, vars=[], rows=[], type="merge")

    # MICHAEL NOTE: could explicitly pass Agnostic() here
    return info.merge(*values)
```

It seems okay to have ColMergeInfo inside _cols_merge.py, and then import it into _gt_data.py, since all of the logic is localized in _cols_merge.py.

rich-iannone:
The last two commits addresses these recent reviewer comments:

> Could you add a method like...
>     def replace_na(*values, tbl_data=None): ... # returns values with None swapped in for NA values
> Then feed the result of that into .merge() (which would then check x is None etc..)?
> That way ColMergeInfo still encapsulates all the logic, and merge_pattern() just uses
> .replace_na w/o passing tbl_data

What was implemented

1. Added `replace_na(*values, tbl_data=None)` static method on ColMergeInfo
   - Returns tuple with NA values replaced by None
   - Uses `tbl_data` for backend-specific NA detection when provided
   - Falls back to generic detection (Agnostic) when tbl_data is None

2. `merge()` now checks `v is None` for missing values
   - Simple and explicit: `col_is_missing = {str(i + 1): v is None for ...}`
   - No longer needs tbl_data parameter

3. `ColMergeInfo` encapsulates all the logic
   - `replace_na()` - NA detection/normalization
   - `merge()` - simple user-facing merge (expects `None` for missing)
   - `merge_values()` - dict-based internal interface (still available)

4. `merge_pattern()` uses `replace_na()` without tbl_data
   ```python
   def merge_pattern(pattern: str, *values: Any) -> str:
       normalized = ColMergeInfo.replace_na(*values)  # no tbl_data
       info = ColMergeInfo(pattern=pattern, vars=[], rows=[], type="merge")
       return info.merge(*normalized)
   ```

5. Internal `_apply_single_col_merge()` builds merge values directly
   - Checks both the body and original for NA (to support `sub_missing()` semantics)
   - Value is missing only if both are NA; `sub_missing()` replacements are NOT missing
   - Passes `None` for truly missing, display value otherwise, then calls `merge()`

rich-iannone:
> RE 1-based indexing, it seems like these are the core dynamics that will be at play in Great Tables:
> 
> * Polars selectors: 0-based
> * Python string formatting: 0-based (e.g. `"{0}".format(1)`)
> * cols_merge: 1-based
> 
> There may be other examples of 1-based things in Python data libraries though? If so, it might be helpful for us to list some out

Examples of 1-based indexing in Python:

**Regular Expressions (built-in `re` module)**
- Backreferences use `\1`, `\2`, etc. (1-based)
- `re.sub(r'(\w+) (\w+)', r'\2 \1', "hello world")` → `"world hello"`

**String Template (built-in `string` module)**
- `string.Template` uses `$1`, `$2` style (though typically named)

**Excel/Spreadsheet Libraries**
- `openpyxl`: Rows and columns are 1-based (`ws.cell(row=1, column=1)`)
- `xlrd`, `xlwt`: Also 1-based for user-facing APIs
- `gspread` (Google Sheets): 1-based

**Database Parameter Binding**
- PostgreSQL (`psycopg2`): Uses `$1`, `$2` for positional parameters
- Oracle (`cx_Oracle`): Uses `:1`, `:2` for positional bind variables

**Statistical/Data Science**
- `statsmodels`: Lag notation often 1-based (lag 1 = first lag)
- `scipy.stats`: Distribution parameters often labeled starting from 1

**Counter-examples (0-based)**
- `str.format()`: `"{0} {1}".format(a, b)`
- Polars/pandas indexing
- Most Python sequences

The strongest precedent is regex backreferences (`\1`, `\2`) since `cols_merge()` patterns serve a similar sort of 'capture and substitute' purpose. This is familiar to anyone who's used regex in Python. (Though `0` is used for the entire match.)

Great Tables uses 0-based indexing for rows everywhere else in the API:

- `loc.body(rows=[0, 1, 2])`: 0-based row indices
- `fmt_number(rows=[0, 1])`: 0-based row indices
- All internal `RowInfo.rownum_i`: 0-based

The only 1-based indexing is in (proposed) `cols_merge()` patterns:

- Pattern placeholders: `{1}`, `{2}`, `{3}`

So `cols_merge()` is actually inconsistent with the rest of Great Tables' API. The `rows=`  parameter uses 0-based indexing, but the pattern placeholders use 1-based indexing.

However, and this is worth noting, the 1-based indexing is ONLY for pattern placeholders within the string (analogous to regex backreferences like `\1`, `\2`), not for row/column selection in the API which remains 0-based throughout.

rich-iannone:
After discussion with @machow we decided to go with 0-based indexing for the `cols_merge()` pattern strings.

rich-iannone:
The last commit makes the changes in the code, docs, and tests to move from 1- to 0-based indexing for the merge pattern. Summary of changes:

**great_tables/_cols_merge.py**: Core merge logic:
- Pattern placeholders: `{1}`, `{2}`, `{3}` → `{0}`, `{1}`, `{2}`
- `validate_pattern()`: removed the `col_idx = int(col_ref) - 1` offset (now just `int(col_ref)`), removed the `< 0` check (no longer needed), updated error message to show valid index range
- `merge()`: dict keys changed from `str(i + 1)` to `str(i)`
- `merge_values()`: docstring updated from "1-based" to "0-based"
- All docstrings and examples updated

**great_tables/_spanners.py**: User-facing `cols_merge()`:
- Default pattern generation: `f"{{{i+1}}}"` → `f"{{{i}}}"`
- All docstring examples and the "How the pattern works" section updated
- Code examples: `"{1}&mdash;{2}"` → `"{0}&mdash;{1}"`, etc.

**tests/test_cols_merge.py**: All 49 tests:
- Every pattern string updated to 0-based
- Dict keys in `merge_values()` tests: `"1"`,`"2"`,`"3"` → `"0"`,`"1"`,`"2"`
- `test_validate_pattern_zero_index_raises` → `test_validate_pattern_zero_index_valid` (positive test)
- `test_zero_based_index_error` → `test_zero_based_indexing` (verifies `{0}-{1}` works)

great_tables/_cols_merge.py
```diff
@@ -0,0 +1,360 @@
...
+        merge_values = []
+
+        for col_name in col_merge.vars:
+            # Get the formatted value from the body
+            formatted_value = _get_cell(body.body, row_idx, col_name)
+
+            # Get the original value from the data table
+            original_value = _get_cell(tbl_data, row_idx, col_name)
+
+            body_is_na = is_na(body.body, formatted_value)
```
machow: I'm confused -- you added a replace_na() method to the class, but manually call is_na(TblData, ...) here. It feels like ColsMergeInfo is now more self-contained, but that path is not used here.

great_tables/_cols_merge.py
```diff
@@ -0,0 +1,360 @@
...
+                # Truly missing: both body and original are NA
+                merge_values.append(None)
+            elif body_is_na and not original_is_na:
+                # Cell is unformatted but has a value in the original data
+                merge_values.append(str(original_value))
+            else:
+                # Cell is formatted (possibly by sub_missing): use formatted value
+                merge_values.append(str(formatted_value))
+
+        # Use merge() as it checks `val is None` for missing detection
```
machow: I don't understand why so much logic is needed here, but not in the merge_values() scalar function. What could they be doing differently? 

* If they're doing something differently, has something gotten misaligned between them?
* If they're not, why is this here?

great_tables/_cols_merge.py
```diff
@@ -0,0 +1,360 @@
...
+
+        Raises
+        ------
+        ValueError
+            If the pattern is None (required for merge operations).
+        ValueError
+            If the pattern references a column index greater than or equal to
+            the number of columns.
+        """
+        if self.pattern is None:
```
machow: Every non-static method has an assertion that self.pattern is set, so it seems like pattern is necessary. Please do not repeat the same assertion in every method

great_tables/_cols_merge.py
```diff
@@ -0,0 +1,360 @@
...
+        The body data to modify.
+    tbl_data
+        The original table data (for checking missing values).
+
+    Returns
+    -------
+    Body
+        The modified body data.
+    """
+    if col_merge.type != "merge":
```
machow: If there are unsupported features that a user tries to set, please raise or warn them the intended behavior will not happen, rather than quietly returning a no-op

machow (changes requested):
Thanks for working on this -- I think the logic isn't quite consolidated yet. I left feedback on piece I think would reduce duplication, merge code paths, and remove unused functions. I went pretty quickly, so it's possible I screwed up when reading / missed things / got things wrong!

rich-iannone:
The comments were addressed in the last three commits. Here's a summary of what was done to address the recent review comments:

1. **`replace_na()` vs `is_na()` inconsistency**: `_apply_single_col_merge()` now uses `ColMergeInfo.replace_na()` consistently instead of calling `is_na()` directly.

2. **Too much logic in `_apply_single_col_merge()`**: the body/original NA branching was simplified to use `replace_na()` for both values, keeping the same 3-way logic (both NA -> None, body-only NA -> use original, else -> use formatted) but through the class's own API.

3. **`_extract_pattern_columns` in wrong file**: moved inline into `ColMergeInfo.pattern_columns` property; removed from `great_tables/_utils.py`. The import was also removed.

4. **Repeated `pattern is None` checks**: made `pattern` a required field. Removed all `if self.pattern is None` guards from `merge()`, `validate_pattern()`, and `pattern_columns`. The pattern is always set at construction time (default generated in `cols_merge()` if not provided by user).

5. **`merge_values()` unused outside tests**: this is now removed entirely. Tests were refactored to use `merge()` directly.

6. **Unsupported merge type silently returns no-op**: changed to `raise NotImplementedError(...)` with a descriptive message.

rich-iannone:
A recent Claude code review flagged other issues (distinct from the human review). Here they are:

```
 Code review                 
                                                                                                                                                       
  Found 2 issues:                            
                                                                                                                                                       
  1. Incorrect pattern index example in docstring -- The pattern parameter description says "{1} {2}" could be used to combine two columns, but the    
  implementation uses 0-based indexing. Using {2} with only 2 columns raises ValueError: Pattern references column {2} but only 2 columns were provided
   (valid indices are 0 to 1). Should be "{0} {1}". All other examples in the same docstring (lines 1092, 1131-1132, 1156-1157) and the ColMergeInfo   
  class docstring correctly use 0-based indices. Commit a1691622 switched to 0-based indexing but this description was not updated.                    
                                                                                                                                                       
  1. https://github.com/posit-dev/great-tables/blob/79ac21ae83411ac857bc4d7713573d161501a68e/great_tables/_spanners.py#L1064-L1072                     
  2. Standard 8 (Code Organization) violation: single-use functions in separate file -- _extract_pattern_columns (line 363) and                        
  _process_col_merge_pattern (line 296) are defined in _utils.py but only imported and used in _cols_merge.py. Standard 8 says: "Avoid single-use
  functions in separate files. If a function is only called in one place, define it next to its caller." The standard's bad example literally names
  _extract_pattern_columns as the anti-pattern. These functions should be moved into _cols_merge.py.

  2. https://github.com/posit-dev/great-tables/blob/79ac21ae83411ac857bc4d7713573d161501a68e/great_tables/_utils.py#L293-L378
  https://github.com/posit-dev/great-tables/blob/79ac21ae83411ac857bc4d7713573d161501a68e/great_tables/_cols_merge.py#L6-L7
```

These review comments will be addressed next.

great_tables/_cols_merge.py
```diff
@@ -0,0 +1,360 @@
...
+
+        Returns
+        -------
+        list[str]
+            List of column reference numbers as strings (e.g., ["0", "1"]).
+        """
+        if self.pattern is None:
+            return []
+
+        return _extract_pattern_columns(self.pattern)
```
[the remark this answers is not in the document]
rich-iannone: Moved inline into ColMergeInfo.pattern_columns property; removed from great_tables/_utils.py. The import was also removed.

great_tables/_cols_merge.py
rich-iannone: Made pattern a required field. Removed all if self.pattern is None guards from merge(), validate_pattern(), and pattern_columns. The pattern is always set at construction time (default generated in cols_merge() if not provided by user).

PR #792, opened 2025-10-31
Fix #791 by handling `interactive_data_values` properly

### LINKED ISSUE #791: `interactive_data_values` gets overwritten in `nanoplot_options()` to always be `True`
## Prework

## Description

Describe the bug clearly and concisely.

## Reproducible example

- [x] Post a minimal reproducible example (MRE) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
  - [x] **Runnable**: post enough code and data so any onlooker can create the error on their own computer.
  - [x] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
  - [x] **Readable**: format your code according to the [Style Guide for Python Code](https://peps.python.org/pep-0008/).

```python
from great_tables import GT, nanoplot_options
from great_tables.data import islands

islands_mini = islands.head(10).astype({"size": float})

# Create a display table showing ten of the largest islands in the world
gt_tbl = GT(islands_mini).fmt_nanoplot(
    columns="size",
    plot_type="bar",
    options=nanoplot_options(interactive_data_values=False),
)

# Show the output table
gt_tbl.show()
```

## Expected result

Data point values should be statically displayed instead of appearing on hover when setting `interactive_data_values=False`. This is [documented behavior](https://posit-dev.github.io/great-tables/reference/nanoplot_options.html#great_tables.nanoplot_options).

## Development environment

- Operating System: Windows 11
- **great_tables** Version: 0.19.0

## Additional context

I am also having trouble getting`show_y_axis_guide=True` to work, but this is another topic.

lorenzo-w:
# Summary

This changes a single line in `nanoplot_options()` to handle `interactive_data_values` option properly instead of overriding it.

- Ref: #791 

# Checklist

### DIFF
--- a/tests/test__utils_nanoplots.py
+++ b/tests/test__utils_nanoplots.py
@@ -1984,6 +1984,22 @@ def test_nanoplot_unknown_plot_type():
         _generate_nanoplot(y_vals=[1, 2, 3], plot_type="unknown")
 
 
+@pytest.mark.parametrize("interactive_data_values", [True, False])
+def test_generate_nanoplot_interactive_data_values_toggles_hover_css(interactive_data_values):
+    svg = _generate_nanoplot(
+        y_vals=[1.0, 2.0, 3.0],
+        interactive_data_values=interactive_data_values,
+    )
+
+    assert _is_nanoplot_output(svg)
+
+    if interactive_data_values:
+        assert ".vert-line:hover rect" in svg
+    else:
+        assert ":hover" not in svg
+        assert ".vert-line rect" in svg
+
+
 @pytest.mark.parametrize(
     "n, bool_",
     [


rich-iannone:
The submitter's fix changes a single line in `nanoplot_options()` in `_helpers.py`. The original code used `interactive_data_values = interactive_data_values or True`, which always evaluates to `True` (so users could never disable interactive data values). The fix replaces it with:

`interactive_data_values = True if interactive_data_values is None else interactive_data_values`, matching the pattern already used for the other boolean options directly above it (e.g., `show_data_points`, `show_data_line`, etc.).

I added a few tests to verify this behavior in `tests/test__utils_nanoplots.py‎`. This example now works correctly:

```python
from great_tables import GT, nanoplot_options
from great_tables.data import islands

islands_mini = islands.head(10).astype({"size": float})

(
  GT(islands_mini)
  .fmt_nanoplot(
    columns="size",
    plot_type="bar",
    options=nanoplot_options(interactive_data_values=False)
  )
)
```

image

And what we get are persistent data labels that aren't 'interactive' (meaning, appearing only on hover; all are shown).

machow:
@rich-iannone I've pushed a simple test up. If it looks good to you, do you mind merging?

PR #793, opened 2025-10-31
feat: support polars expressions in vals functions

machow:
This PR enables the vals functions to work on Polars expressions, by returning an expression that can be used in methods like `polars.DataFrame.with_columns()`.

For example:

```python
import polars as pl
from great_tables import vals

df = pl.DataFrame({"x": [-1.23]})

res = df.select(fmt = vals.fmt_integer(pl.col("x"), accounting=True))
assert res[0, "fmt"] == "(1)"
```

Currently, the vals methods take the slow `.map_elements()` path, which is compounded by many of them instantiating a GT object for each element. However, it should be quick to speed up. The main reason for using `.map_elements()` right now, is that the vals do not propogate Nones (but return the string `"None"`). As it turns out, `.map_element()` was being used because it can skip nulls.

I think we can get it to use `.map_batches()` by...

* propogating None values (breaking change)
* less critical: returning the same type as input (currently we always return a list)

### DIFF
--- a/great_tables/_formats_vals.py
+++ b/great_tables/_formats_vals.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 from functools import wraps
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, Concatenate, ParamSpec, overload
+from typing import TYPE_CHECKING, Any, Callable, overload
 
-from typing_extensions import TypeAlias
+from typing_extensions import TypeAlias, Concatenate, ParamSpec
 
 from ._gt_data import GTData, FramelessData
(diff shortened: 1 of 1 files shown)


machow:
Merging while pairing w/ Rich

jrycw:
This feature looks great! I actually ran into a similar situation before, where I wanted to apply the `vals` functions directly to a Polars DataFrame. I ended up using an approach like this:

```python
from functools import partial

df.select(
    pl.col("x")
    .map_elements(
        partial(vals.fmt_integer, accounting=True),
        return_dtype=pl.List(pl.String),
    )
    .list.get(0)
)
```

PR #800, opened 2025-12-23
add int to RowSelectExpr definition

tylerriccio33:
This is super small but I'm pretty sure int should be a part of `RowSelectExpr` type definition. I found this in a pointblank pr where I'm trying to type all the signatures.


tylerriccio33:
Anything I can do to push this along :)

PR #806, opened 2026-01-14
ci: add Plausible analytics to project website

rich-iannone:
This PR sets up the docs site for Plausible analytics.


rich-iannone:
From the site preview, we can see that the two `<script>` tags were added to the head (scan for `plausible` near the bottom):

image
 
In checking other pages, I see that the script tags exist there as well.

I don't think we can know if this works for tracking until we get verification from Jeroen (which might have to be after the non-preview site is built).

machow:
Alright, let's just let this thing fly and see what happens!

PR #810, opened 2026-01-22
refactor: clean up auto-align code

machow:
This PR is for refactoring the auto-align functionality to clean it up a bit. However, more recent releases of pandas and polars may have interacted with auto-align to change snapshots. So the initial work here will be updating snapshots / validating against CI.

NOTES:

* Original tests are in `tests/test__boxhead.py`: left preserved for refactor, in order to separate concerns (keep old behaviors).
* New tests are in `tests/test_tbl_data_align.py` to reflect that logic has moved out of boxhead, into that file.
* At first glimpse, these changes seem very reasonable

FUTURE WORK:

* This PR tries to operate as a strict refactor, by largely preserving even the approach originally used. I think down the road we will have a nice time using narwhals to do the type representation / detection work.

### DIFF
--- a/tests/test_tbl_data_align.py
+++ b/tests/test_tbl_data_align.py
@@ -2,6 +2,7 @@
 
 import pandas as pd
 import polars as pl
+import pyarrow as pa
 import pytest
 
 from great_tables._tbl_data_align import (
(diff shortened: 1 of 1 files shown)


rich-iannone:
@machow is this ready to merge?

machow:
From pairing w/ Rich Iannone, going to merge

PR #811, opened 2026-01-26
feat: add Claude skill for authoring a table

rich-iannone:
This PR adds a skill that enables Claude (or other LLMs) to better build a table. The instructions provide design guidance on how to build tables more effectively (focus is on principles and less on the specific API methods).


machow:
From pairing w/ Rich, let's re-open this after we've cleared our PR backlog

PR #812, opened 2026-01-27
ci: add tests for the `opt_all_caps()` method

rich-iannone:
This PR adds tests for the `opt_all_caps()` method (which didn't have any tests before this PR).


rich-iannone:
I checked the tests for the other (closed) branch and found that we have sufficient overlap with the tests here. For that reason, I will now merge.

Issue #795, opened 2025-11-07
Unicode characters in column headers break when exporting to png


mrapacz:
## Prework

- [X] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [X] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue. - Haven't found one.

## Description

When using GT.save() to save a table to a PNG file, Unicode characters in the column headers are not rendered correctly.

## Reproducible example

I cloned the repo and installed it with `pip install -e .[all]`.
Then I ran the following code:

```python
import pandas as pd
from great_tables import GT

df = pd.DataFrame({"Żaba": ["1", "2"], "Koń": ["3", "4"]})
GT(df).save("output.png")
```

The output file `output.png` renders the column headers as "Ĺ»aba" and "KoĹ„" instead of the expected "Żaba" and "Koń".

## Expected result

The Unicode characters in the column headers should be rendered correctly.
Expected:
<img width="484" height="232" alt="Image" src="https://github.com/user-attachments/assets/bbe67fe2-c051-4114-99e4-77122a7ef404" />
Actual:
<img width="484" height="232" alt="Image" src="https://github.com/user-attachments/assets/68fb0b70-a5ab-478c-9bf3-bde807763413" />

## Development environment

- Operating System: macOS Sequoia 15.6.1
- **great_tables** Version: Tested on the current main branch: a59301b. Also present in `v0.20.0`.

## Additional context

I did some extra digging to try to understand what the root cause of the problem is:

1. The `as_raw_html()` call in the `save()` function [link](https://github.com/posit-dev/great-tables/blob/main/great_tables/_export.py#L451-L452) is not passed an explicit `make_page` parameter, which means the function runs with `make_page=False` [link](https://github.com/posit-dev/great-tables/blob/main/great_tables/_export.py#L132).

2. This, in turn means the table is rendered as div (inline) element only.

3. This html is later saved to a temporary file [link](https://github.com/posit-dev/great-tables/blob/main/great_tables/_export.py#L462-L463) and opened using a webdriver [link](https://github.com/posit-dev/great-tables/blob/main/great_tables/_export.py#L467).
[...]

PR #823, opened 2026-03-14
fix: prepend charset meta tag to HTML in save()

mrapacz:
# Summary

Headless browsers (like Chrome) can misidentify the encoding of an HTML fragment and corrupt Unicode characters. 

Fix: Added a one-liner change as suggested by @rich-iannone, prepending raw html with  `<meta charset='utf-8'>` to make the webdriver recognize encoding and render characters correctly in the resulting image file.

- Ref: #795 

# Checklist

       N/A - There was no test that would catch this beforehand and I wasn't sure testing 'when we prepend a string with another string, we should test that the prefix is there' warrants a new unit test. But if you think it does, let me know and I'll be happy to add it :)

### DIFF
--- a/great_tables/_export.py
+++ b/great_tables/_export.py
@@ -449,7 +449,8 @@ def save(
         file = str(Path(file).with_suffix(".png"))
 
     # Get the HTML content from the displayed output
-    html_content = as_raw_html(self)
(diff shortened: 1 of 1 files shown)


rich-iannone (approved):
LGTM, thanks!

PR #827, opened 2026-04-10
fix: update snapshots affected by updated `css_inline` library

rich-iannone:
This PR is intended for fixing failing snapshot tests. The dependency library `css_inline` recently updated and now produces HTML with trailing whitespace.


rich-iannone:
@t-kalinowski could you quickly look at this PR? Just a test fix for an updated dependency.

t-kalinowski (approved):

PR #828, opened 2026-05-13
test: add failing footnote target regressions

t-kalinowski:
Add regression coverage for footnote targets that currently produce footer entries without rendering a corresponding mark at the target location.

The new tests cover three public API cases:

- row group footnotes created with loc.row_groups(), where the footer note is present but the row group heading does not receive a footnote mark
- grand summary body footnotes created with loc.grand_summary(), where the footer note is present but the targeted grand summary cell does not receive a footnote mark
- spanner footnotes created with loc.spanner_labels(ids=...) when the spanner id differs from the displayed label, where lookup currently misses the target because rendering matches against label text instead of the spanner id

These tests document the expected rendered HTML behavior and should pass once footnote marks are applied to all supported target locations and spanner footnote matching uses stable spanner ids.

### DIFF
--- a/tests/test_footnotes.py
+++ b/tests/test_footnotes.py
@@ -167,6 +167,62 @@ def test_tab_footnote_all_locations():
     assert re.search(r"Test Subtitle<span[^>]*>2</span>", html)  # Subtitle
 
 
+def test_tab_footnote_row_group_target_gets_mark():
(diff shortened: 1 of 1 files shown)


t-kalinowski:
Opened PR against wrong branch.

PR #860, opened 2026-09-09
cols_*() methods do not raise if a list of column names not all in data

### LINKED ISSUE #817: cols_*() methods do not raise if a list of column names not all in data
The `cols_*()` methods raise if a single column name is passed that is not in the data. However, they do not raise if that missing name is in a list.

Here's a reprex

```python
# %%
import pandas as pd                                                                                                                                                       
import great_tables as gt                                                                                                                                                 
                                                                                                                                                                          
df = pd.DataFrame({"A": [1], "B": [2]})                                                                                                                                   
t = gt.GT(df)                                                                                                                                                             
                                                                                                                                                                          
# String input → raises error                                                                                                                                             
# t.cols_align(columns="NOPE")           # AssertionError
# t.cols_label_with(columns="NOPE", fn=str.lower)  # AssertionError

# List input → silently ignores "NOPE"
t.cols_align(columns=["A", "NOPE"])              # no error, only A aligned
t.cols_label_with(columns=["A", "NOPE"], fn=str.lower)  # no error, only A relabeled
t.cols_hide(columns=["A", "NOPE"])               # no error, only A hidden
t.cols_move(columns=["A", "NOPE"], after="A")               # no error, only A moved
t.cols_move_to_start(columns=["A", "NOPE"])      # no error
t.cols_move_to_end(columns=["A", "NOPE"])        # no error
```

manan28025:
Fixes #817.

`cols_align()`, `cols_label_with()`, `cols_hide()`, `cols_unhide()`, `cols_move()`, `cols_move_to_start()`, `cols_move_to_end()`, and `cols_reorder()` raise when a single invalid column name is passed as a string, but silently drop invalid names when passed inside a list.

The root cause: `resolve_cols_c()` filters out any unmatched string column names before returning the resolved list, so downstream validation checks against the already-filtered list can never catch an invalid name that was in the original request.

Fixed by validating the originally-requested column names directly, in addition to the existing checks.

Tested with the repro from the issue, added regression tests covering all affected methods, and ran the full test suite (`make test`) plus `ruff check`/`ruff format` on the changed files.


rich-iannone (approved):
LGTM, thanks!!

Issue #714, opened 2025-06-19
tab_spanner fails when no columns are found


aloy99:
When a Polars selector is passed to the `columns` parameter in `tab_spanner`, and no columns are found, `NotImplementedError` is raised:

```
    217 if not len(selected_column_names) and not len(spanner_ids):
    218     # TODO: null_means is unimplemented
--> 219     raise NotImplementedError("columns/spanners must be specified")
```

`gt` in R does not behave similarly: there are no errors raised when the same situation is reproduced. The behaviour I would expect is for the spanner to not be created if no columns matching the SelectExpr are found, similar to what happens for the formatting functions. Otherwise, a generalized tabulation function would have to manually check the columns of the input data to cover all cases, and then create the spanners for each selection only if there is at least one match - this seems unintuitive to me.

MRE:

```
from great_tables import GT, md, html
from great_tables.data import islands
import polars.selectors as cs

islands_mini = (
    pl.from_pandas(islands).sort("size", descending=True)
    .head(10)
)

( # spanner is created without issues
    GT(islands_mini,
       groupname_col="name"
    )
    .tab_header(
        title="Large Landmasses of the World",
        subtitle="The top ten largest are presented"
    )
    .tab_stub(rowname_col="name")
    .tab_spanner(
        "size", columns=cs.contains("size"), gather=True
    )
)

( # NotImplementedError
    GT(islands_mini,
       groupname_col="name"
    )
    .tab_header(
        title="Large Landmasses of the World",
        subtitle="The top ten largest are presented"
    )
    .tab_stub(rowname_col="name")
    .tab_spanner(
        "size", columns=cs.contains("area"), gather=True
    )
)
```

PR #865, opened 2026-09-17
fix: make `tab_spanner()` a no-op when no columns or spanners are selected

Rodrigo-Palma:
# Summary

When the `columns=` selector in `tab_spanner()` resolves to zero columns (for example a Polars selector such as `cs.contains("area")` with no match) and no `spanners=` are given, the method raised:

```
NotImplementedError: columns/spanners must be specified
```

This was a placeholder (`# TODO: null_means is unimplemented`). This PR returns the table unchanged instead, which matches R's `gt`:

```r
# If `column_names` and `spanner_ids` have zero lengths then
# return the data unchanged
if (length(column_names) < 1 && length(spanner_ids) < 1) {
  return(data)
}
```

It is also consistent with how other selection based methods here handle an empty match: `fmt_*()`, `cols_align()`, `data_color()` and `tab_style()` all leave the table untouched. This lets generic table code add spanners from selectors without checking the columns first, which is the use case described in the issue.

The validations that run before the selection (negative `level=`, duplicate `id=`) still raise as before.

Changes:

- `great_tables/_spanners.py`: return `self` instead of raising when nothing is selected.
- `tests/test_spanners.py`: `test_tab_spanner_no_columns_raises` pinned the old placeholder error, so it now checks the no-op behavior (`columns=[]`). Added a regression test for a Polars selector with no matching columns.

- Ref: #714

# Checklist


rich-iannone (approved):
LGTM, thanks!

PR #866, opened 2026-09-17
fix: honor output_gt(placeholder=) via output_ui

### LINKED ISSUE #852: output_gt(placeholder=) is accepted and then ignored
## Problem

`great_tables.shiny.output_gt()` accepts a `placeholder` argument and then ignores it completely.

```python
>>> import inspect, great_tables.shiny as s
>>> inspect.signature(s.output_gt)
(id: 'str', placeholder: 'bool' = False) -> 'Tag'
>>> print(inspect.getsource(s.output_gt))
def output_gt(id: str, placeholder: bool = False) -> Tag:
    """Output UI for a great_tables table."""
    return div({"class": "shiny-html-output"}, id=resolve_id(id))
```

The parameter is never read. `placeholder=True` and `placeholder=False` produce identical output.

## Why it matters

The argument shows up in the rendered signature, so users reasonably assume it does something — the obvious guess being "draw an empty bordered box before the table arrives", which is what the equivalent argument does elsewhere in the Shiny ecosystem. Verified with `great-tables` 0.23.0.

## Suggested resolution

Either implement it (emit a placeholder element when `True`) or remove it from the signature.

Found while writing the Great Tables page for the Shiny for Python docs site (posit-dev/py-shiny-site#389). We deliberately left `placeholder` **undocumented** rather than describe behavior that does not exist, so a decision here would let us document it properly.

00200200:
Fixes #852

`output_gt(placeholder=...)` ignored the argument, so True and False produced the same Shiny UI. I forward it to `shiny.ui.output_ui()`.

Tested with:
`pytest tests/test_shiny.py`

# Checklist


00200200:
Closing this: on a closer look the approach doesn't actually fix #852. `output_ui(id, placeholder=True)` only adds an inert `placeholder` HTML attribute, which Shiny never reads (the placeholder box for text/code outputs is a CSS class, not an attribute), and the added test only checks that the two rendered strings differ, so it can't catch that. Sorry for the noise — happy to send a proper implementation or a removal of the parameter if the maintainers prefer one of those.

PR #867, opened 2026-09-18
fix(nanoplots): warn when a curved data line is dropped for x_vals

### LINKED ISSUE #256: nanoplots: If `x_vals` is passed it silently overrides `line_type` curved to be straight
Currently, if you try to specify line_type curved, but also supply x_vals, we quietly set line_type back to straight.

```python

_generate_nanoplot(
    y_vals=[1, 2, 3],
    x_vals=[2, 4, 6],
    show_data_line=True,
    plot_type="line",
    line_type="curved"
)
```

If we want to do this, then let's add a guard for x_vals at the very top, that does this one thing (warns user, and overrides their option). That way it will be very easy for us to spot, and users will know what's happening.

Currently, it happens pretty high up, but there's a little bit of extra processing that's going on in tandem.

Rodrigo-Palma:
# Summary

Closes #256.

When `x_vals` are supplied, `_generate_nanoplot` reset `data_line_type` to `"straight"`
partway through the function, after the `x_vals` length checks and NA removal had
already run. A caller asking for `data_line_type="curved"` got a straight line with no
indication that the option had been dropped:

```python
from great_tables._utils_nanoplots import _generate_nanoplot

curved = _generate_nanoplot(y_vals=[1, 2, 3], x_vals=[2, 4, 6], show_data_line=True, data_line_type="curved")
straight = _generate_nanoplot(y_vals=[1, 2, 3], x_vals=[2, 4, 6], show_data_line=True, data_line_type="straight")
curved == straight  # True, with no warning
```

Without `x_vals` the two differ as expected: `curved` emits `<path d="M ... C ...">`
and `straight` emits `<polyline>`.

# Change

This follows the approach in the issue: a guard at the top of the function that does
the one thing, warns and overrides the option. The silent assignment further down is
removed, along with its `TODO`, so the fallback now lives in a single place.

The rendered output is unchanged. The only behavioural difference is the `UserWarning`.
`data_line_type` is validated by `_match_arg` just above the guard, so it can only be
`"curved"` or `"straight"` at that point.

- Ref: #256

# Checklist

Two tests in `tests/test__utils_nanoplots.py`: one asserting the warning and the
fallback to the straight output, one asserting that a curved line without `x_vals`
still renders as a curve. The first fails without the source change (`DID NOT WARN`).
Full suite is green locally, including the 192 snapshots.


Rodrigo-Palma:
The workflow runs on this PR are still sitting in `action_required`, so there are no check results here at all, just an empty rollup. The sibling PRs were approved yesterday and are green (#865 and #870), so this one looks like it was missed in that batch.

To have the result available in advance, I ran this repository's own CI on my fork, on the same commit as this PR (`74c0acb`):

* [CI Tests](https://github.com/Rodrigo-Palma/great-tables/actions/runs/35448185220): build on 3.10, 3.11, 3.12, 3.13 and 3.14, plus test-windows and test-pandas, all passing.
* [Code Checks](https://github.com/Rodrigo-Palma/great-tables/actions/runs/35448185214): pre-commit and the optional-deps static check, both passing.
* [CI Docs](https://github.com/Rodrigo-Palma/great-tables/actions/runs/35448185213): Build Docs and Preview Docs, both passing.

Happy to have the runs approved here whenever it is convenient. The change is independent of #865 and #870: it only touches the nanoplot path, and GitHub reports it mergeable against main.

rich-iannone:
Apologies for the delay in approving. And also, thank you for the series of PRs! These are great contributions and I really appreciate all of this :)

rich-iannone (approved):
LGTM, thanks!

PR #869, opened 2026-09-18
fix(formats): pad the year in the iso date style

### LINKED ISSUE #868: fmt_date(date_style="iso") drops leading zeros from the year
## Prework

- [X] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [X] If there is [already a relevant issue](https://github.com/posit-dev/great-tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

`date_style="iso"` drops the leading zeros of the year, so the result is not a valid
ISO 8601 date. The style is defined as `"y-MM-dd"`, and in CLDR `y` is the year with no
minimum width, while `yyyy` is the zero-padded four-digit year.

## Reproducible example

```python
from datetime import date
from great_tables import vals
from great_tables._formats import _iso_str_to_date

out = vals.fmt_date(["0999-01-05"], date_style="iso")[0]
print(out)                  # '999-01-05'

_iso_str_to_date(out)       # ValueError: Invalid isoformat string: '999-01-05'
date.fromisoformat(out)     # ValueError: Invalid isoformat string: '999-01-05'
```

`vals.fmt_date(["0001-01-05"], date_style="iso")` gives `'1-01-05'`.

## Expected result

`'0999-01-05'` and `'0001-01-05'`. ISO 8601 requires four digits for the year, and the
output of the `iso` style should be readable back as an ISO date. Right now it is rejected
by `great_tables`' own `_iso_str_to_date` and by `datetime.date.fromisoformat`.

Years with four digits are unaffected either way.

## Development environment

- Operating System: macOS
- **great_tables** Version: 0.24.0 and `main`

## Additional context

The R gt package uses the same `"y-MM-dd"` code for this style, so changing it is a small
divergence. It seemed worth raising anyway because the style is named for a standard that
fixes the year width, and because the current output cannot be parsed back. Only the `iso`
entry would change; the other styles that show a year (`year`, `year.mn.day`, and the
long-form ones) read naturally without padding and would stay as they are.

Happy to open a PR.

Rodrigo-Palma:
# Summary

Closes #868.

The `iso` date style is defined as `"y-MM-dd"`. In CLDR, `y` is the year with no minimum
width, so a year below 1000 loses its leading zeros and the result stops being an ISO 8601
date:

```python
from datetime import date
from great_tables import vals
from great_tables._formats import _iso_str_to_date

out = vals.fmt_date(["0999-01-05"], date_style="iso")[0]   # '999-01-05'
_iso_str_to_date(out)                                      # ValueError: Invalid isoformat string
date.fromisoformat(out)                                    # ValueError: Invalid isoformat string
```

`"0001-01-05"` comes out as `"1-01-05"`.

# Change

One line: the `iso` entry in `_get_date_formats_dict()` becomes `"yyyy-MM-dd"`, which is
the zero-padded four-digit year in CLDR. The style now round-trips through
`date.fromisoformat()`.

Four-digit years produce the same output as before, so existing tables are unaffected, and
the 192 snapshots pass unchanged.

The R gt package uses `"y-MM-dd"` for this style too, so this is a small divergence. It
seemed worth it because the style is named after a standard that fixes the year width, and
because the current output cannot be read back. Only `iso` changes; the other styles that
show a year (`year`, `year.mn.day`, and the long-form ones) read naturally without padding
and are left alone.

- Ref: #868

# Checklist

`test_fmt_date_iso_pads_the_year` in `tests/test_formats.py` checks the padded output and
that every value parses with `date.fromisoformat()`. It fails without the source change.
Full suite green locally, snapshots included.


rich-iannone (approved):
LGTM!

Issue #661, opened 2025-04-04
Wide columns in latex output with spans


malcolmbarrett:
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and [contributing guidelines](https://github.com/posit-dev/great_tables/blob/main/.github/CONTRIBUTING.md).
- [x] If there is [already a relevant issue](https://github.com/posit-dev/great_tables/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

When using exporting a GT to PDF, I get unusual output. The base problem is that the formatting is off with a span: the first column is very wide.

## Reproducible example

Here is a quarto file

````markdown
---
format: pdf
---

Example using the `gtcars` dataset.

```{python}
from great_tables import GT
from great_tables.data import gtcars
import polars as pl

gtcars_pl = (
    pl.from_pandas(gtcars)
    .select(["mfr", "model", "hp", "trq", "mpg_c", "msrp"])
)

gt_tbl = (
    GT(
        gtcars[["mfr", "model", "hp", "trq", "msrp"]].head(5),
    )
    .tab_stub(rowname_col="mfr", groupname_col="model")
    .tab_spanner(
        label="Performance",
        columns=["hp", "trq"]
    )
)

gt_tbl
```

````

And the result is 

<img width="710" alt="Image" src="https://github.com/user-attachments/assets/90ee0088-6af5-4c00-83c0-889b065a399a" />

When I use `gt_tbl.as_latex()` here, I get an error. 

Relatedly, I seem to be getting different output based on if I end a cell with a gt object (implicitly printing/showing it) vs. `as_latex()`. Try it with this example without `tab_stub()` and you will see what I mean.

## Expected result

The table should not float off the page, and the first column should not be so wide.

## Development environment

- Operating System: Mac

Here's my uv toml:

```toml
[project]
name = "gt-reprex"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "great-tables>=0.17.0",
    "jupyter>=1.1.1",
    "pandas>=2.2.3",
    "polars>=1.26.0",
]
```

PR #870, opened 2026-09-18
fix(latex): emit one cell per column in a spanner row

Rodrigo-Palma:
# Summary

Ref #661.

A spanner row in the LaTeX output carries fewer cells than the table has columns, and the
`\cmidrule` under a spanner points at columns to the left of the ones it covers.

```python
from great_tables import GT

sub = gtcars[["mfr", "model", "hp", "trq", "msrp"]].head(2)
print(GT(sub).tab_spanner(label="Performance", columns=["hp", "trq"]).as_latex())
```

Before:

```latex
\begin{tabular*}{\linewidth}{@{\extracolsep{\fill}}llrrr}
\toprule
 & \multicolumn{2}{c}{Performance} &  \\
\cmidrule(lr){2-3}
mfr & model & hp & trq & msrp \\
```

The table has five columns, the spanner row covers four (`1 + 2 + 1`), and `hp` and `trq`
are the third and fourth columns, not the second and third. After:

```latex
 &  & \multicolumn{2}{c}{Performance} &  \\
\cmidrule(lr){3-4}
```

A row that is short of cells is what makes the first column stretch under
`\extracolsep{\fill}`, which is the symptom reported in #661. That issue also involves a
stub, and this fixes the spanner part of it; I have not verified the rendered PDF, only the
emitted LaTeX.

## Cause

Two things in `create_columns_component_l`:

- A group of columns with no spanner was written as `spanner_labs.append("" * span)`.
  `"" * span` is always the empty string, so a group covering n columns contributed one
  cell instead of n.
- `span_accumulator` was only advanced inside the spanner branch (`span - 1`), and the
  `cmidrule` bounds were derived from `j`, the index of the *group*. Groups of unspanned
  columns therefore never moved the offset.

Both are fixed by emitting one cell per covered column and advancing the accumulator by
every group's span, spanned or not.

- Ref: #661

# Checklist

Two new tests in `tests/test_utils_render_latex.py`: one parametrized test asserting every
spanner row covers the same number of columns as the headings row (with and without a stub,
and with stacked spanners), and one asserting the `cmidrule` bounds match the spanned
columns. All five fail without the source change.

**Two existing tests were updated**, because they recorded the short rows:

- `test_create_columns_component_l_one_spanner` expected
  `\multicolumn{2}{c}{Spanner} &  \\`, which is 3 columns for the 9 columns of `exibble`.
  It now expects the 7 trailing empty cells.
- `test_create_columns_component_l_many_spanners` expected 7 columns on its upper spanner
  row, and now expects 9. Its lower row and every `cmidrule` in it were already correct,
  since each unspanned group there happens to be one column wide, which is the case where
  the old code coincidentally worked.

`test_create_columns_component_l_adjacent_spanners_hiding` is unchanged for the same
reason. Full suite green locally: 2513 passed, snapshots included.


Rodrigo-Palma:
Two things that may help this along.

**The invariant now has wider evidence.** I could not verify a rendered PDF, as I said above, so instead I checked the emitted LaTeX structurally across seven spanner layouts: every header row must cover exactly as many columns as the table has, counting `\multicolumn{n}` as n, and each `\cmidrule` must point at the columns its spanner actually covers.

On `main`, three of the seven fail:

| case | header row covers | `\cmidrule` | correct |
|---|---|---|---|
| spanner over the middle columns (c, d) | 4 of 5 | `{2-3}` | `{3-4}` |
| spanner over the first columns (a, b) | 3 of 5 | `{1-2}` | `{1-2}` |
| **spanner over the last columns (d, e)** | 3 of 5 | `{2-3}` | `{4-5}` |
| two spanners | 5 of 5 | correct | correct |
| spanner over every column | 5 of 5 | correct | correct |
| with a stub | 5 of 5 | correct | correct |
| no spanner | 5 of 5 | correct | correct |

With this branch all seven pass. The last-columns case is worth singling out: the rule there is off by two columns, pointing at `b` and `c` while labelling `d` and `e`. I had only reported the middle-column case originally.

**It still applies cleanly.** The branch is 16 commits behind `main` now, but merging `main` into it produces no conflict, so there is nothing for me to rebase unless you would prefer the history linear.

Happy to add the seven-case check as a test in this PR if you would like the invariant pinned rather than just measured.

Rodrigo-Palma:
Follow-up on the one caveat in my description: I have now compiled the LaTeX and measured the rendered PDF, rather than only reading the emitted source.

Same example as above, compiled with Tectonic, then measured on the rasterized page. The table spans x = 1041 to 1508 px, and the column labels sit at:

| column | x range |
|---|---|
| mfr | 1042-1057 |
| model | 1143-1172 |
| hp | 1286-1297 |
| trq | 1378-1390 |
| msrp | 1478-1501 |

The `\cmidrule` under `Performance` lands at:

- **`main`**: x **1142-1298**, which covers `model` and `hp`
- **this branch**: x **1271-1392**, which covers `hp` and `trq`

So in the rendered document the rule really is sitting under the wrong two columns, one column to the left of the ones the spanner labels, and this branch puts it under the right ones. The `\toprule`, `\midrule` and `\bottomrule` are byte-identical between the two (x 1041-1508 each), so nothing else about the table geometry moves.

That was the only part of #661 I could not confirm before.

rich-iannone (approved):
LGTM!

rich-iannone:
I visually verified that the LaTeX tables (with different variations of spanners set) all look correct after the fix. No LaTeX errors either in all cases. Thanks for this valuable contribution!

rich-iannone:
Approved and will merge after CI passes.

PR #511, opened 2024-11-17
dev: fix some `mypy` errors and switch to ruff

jrycw:
Hello team,  

I ran `mypy` on our codebase and resolved around 60 reported errors in this PR (leaving 300+ errors still to address). Most of the changes involve refining type hints.  

Based on my observations, `mypy` seems to struggle with the following scenarios:  
* Redefining a variable within the same scope while changing its type.  
* Fully understanding the logic behind our `DataFrame` backend.  
* Recognizing that `GTSelf` is a `TypeVar` for `GTData`, which causes many `GT` attributes to be flagged as "not found."  
* Handling returned values that might include `None`, which can be challenging to annotate correctly.  

Feel free to accept, reject, or modify anything in this PR.

PR #872, opened 2026-09-19
fix(latex): escape backslash, tilde and circumflex with their macros

Rodrigo-Palma:
# Summary

`_latex_escape` prefixes a backslash to every character it matches:

```python
latex_escape_regex = "[\\\\&%$#_{}~^]"
text = re.sub(latex_escape_regex, lambda match: "\\" + match.group(), text)
```

That is right for `&`, `%`, `$`, `#`, `_`, `{` and `}`, and wrong for the other three characters in the class, because none of them has a prefixed form:

| input | current output | what LaTeX does with it | correct output |
|---|---|---|---|
| `a\b` | `a\\b` | `\\` is a line break | `a\textbackslash{}b` |
| `a~b` | `a\~b` | `\~` is an accent, and it takes `b` as its argument | `a\textasciitilde{}b` |
| `a^b` | `a\^b` | `\^` is an accent, and it takes `b` as its argument | `a\textasciicircum{}b` |

Reproduced on `main` at `dfc9b40de`:

```python
import polars as pl
from great_tables import GT

print(GT(pl.DataFrame({"path": ["C:\\Users\\me"]})).as_latex())
```

The body row is emitted as `C:\\Users\\me \\`, so a single cell carries two line breaks inside a `tabular*` row. A value ending in a backslash produces `\\ \\`, which LaTeX rejects outright.

The R **gt** package, which this function mirrors, maps exactly these three to macros. Its inverse, `decode_latex_special_chars` in `R/utils.R`, is explicit about it:

```r
text <- gsub("\\textasciitilde{}", "~",  text, fixed = TRUE)
text <- gsub("\\textasciicircum{}", "^", text, fixed = TRUE)
text <- gsub("\\textbackslash{}", "\\", text, fixed = TRUE)
```

So this PR replaces the blanket prefix with a small lookup for the three, leaving the other seven exactly as they were.

**One existing test changed.** `test_latex_escape` asserted `_latex_escape("\a_\d") == "\\a\_\\d"`, which recorded the doubled backslash. Its line comes from #511, a mypy/ruff cleanup, rather than from a decision about escaping, so I updated the expectation to `\textbackslash{}a\_\textbackslash{}d` and added two tests: one covering the three macro characters and one pinning the seven prefixed ones so they cannot regress.

Full suite is green locally (`pytest tests/`, 192 snapshots passed).

- Ref: no existing issue. Found while checking how user content survives the LaTeX output path.

# Checklist


rich-iannone (approved):
LGTM!!

PR #873, opened 2026-09-19
fix(utils): match options on a prefix instead of a substring

Rodrigo-Palma:
# Summary

`_match_arg` selects an option with `x in el`, which is a substring test anywhere in the option, and then resolves ties with `matched.pop()`, which returns the last one. Together those two make a mistyped option into a silently accepted one:

```python
from great_tables import GT
import polars as pl

df = pl.DataFrame({"a": [1]})

GT(df).opt_align_table_header(align="ight")   # accepted, becomes "right"
GT(df).opt_align_table_header(align="e")      # accepted, becomes "center"
GT(df).opt_align_table_header(align="")       # accepted, becomes "right"
GT(df).opt_stylize(color="")                  # accepted
```

The empty string is the clearest case: it is contained in every option, so it always matches, and `pop()` hands back whichever option is last in the list. Nothing warns, and the user sees a table that is aligned or coloured in a way they never asked for.

This reaches users through `opt_align_table_header(align=)`, `opt_table_outline(style=)`, `opt_stylize(color=)`, the footnote marks option, and four call sites in the nanoplot code.

The function is named after `match.arg` in R, which allows an abbreviation but matches it as a **prefix** (via `pmatch`) and raises when the abbreviation is ambiguous. This PR brings `_match_arg` in line with that:

* an option may still be abbreviated, by a prefix;
* an exact match wins over a prefix shared with longer options, so `("red", ["red", "reddish"])` stays `"red"`;
* an ambiguous abbreviation raises instead of quietly picking the last candidate;
* a value that is not a prefix of any option, the empty string included, raises the existing "not an allowed option" error.

**This is a behaviour change**, and worth calling out: code that today passes a substring such as `align="ight"` currently works by accident and will start raising. That is the point of the change, but if you would rather not break those callers in a patch release, I am happy to emit a warning for one cycle and raise afterwards, or to restrict the change to rejecting the empty string only. Tell me which you prefer and I will adjust.

Full suite is green locally (`pytest tests/`, exit 0, 192 snapshots passed), including the four nanoplot call sites.

- Ref: no existing issue. Found while auditing option validation for values that are accepted without matching anything.

# Checklist


rich-iannone (approved):
Looks good, thank you!

PR #874, opened 2026-09-19
fix(formats): accept locale names regardless of case

Rodrigo-Palma:
# Summary

The case of a locale tag currently decides whether it is accepted:

```python
from great_tables import vals

vals.fmt_number([1234.5], locale="pt-BR")   # '1.234,50'
vals.fmt_number([1234.5], locale="pt-br")   # ValueError
vals.fmt_number([1234.5], locale="en-us")   # ValueError
vals.fmt_number([1234.5], locale="de-ch")   # ValueError
```

`_validate_locale` replaces underscores with hyphens and then tests membership with `not in locales_list`, an exact string comparison, and `_normalize_locale` resolves default locales the same way. Neither touches case, so only the conventional spelling works. The error message is itself a hint that this was not intended:

```
The normalized locale name `pt-br` is not in the list of locales.
```

The value is called normalized, and the only normalization applied was the underscore.

Case in a BCP 47 tag carries no meaning: the convention is lowercase language, titlecase script, uppercase region, but `en-us` and `en-US` identify the same locale, and an application that takes a locale from a header, a URL or a config file frequently has the lowercase form in hand.

This PR adds a small helper that resolves a supplied tag to the spelling used in the locale data, matching without regard to case, and uses it in both `_validate_locale` and `_normalize_locale`. Behaviour otherwise is unchanged:

* every spelling that worked before still works and still returns the same result;
* `pt-br`, `EN-US` and `de-ch` now resolve to `pt-BR`, `en` and `de-CH`;
* an unknown locale such as `xx` still raises, and so does an unknown tag that babel rejects.

Full suite is green locally (`pytest tests/`, exit 0, 192 snapshots passed), and the new behaviour is covered by two tests next to the existing locale tests.

- Ref: no existing issue. Found while checking which locale spellings the formatting functions accept.

# Checklist


rich-iannone (approved):
Thanks!

PR #876, opened 2026-09-21
fix(utils): match the http scheme case-insensitively

Rodrigo-Palma:
# Summary

`is_valid_http_schema()` compares the prefix verbatim, so it only recognizes a lowercase scheme:

```python
is_valid_http_schema("https://posit.co/logo.png")  # True
is_valid_http_schema("HTTPS://posit.co/logo.png")  # False
```

URI schemes are case-insensitive: RFC 3986, Section 3.1 says "schemes are case-insensitive", and Python's own `urllib.parse.urlparse("HTTPS://posit.co/a.png").scheme` returns `https` for exactly that reason.

The user-visible effect is in `fmt_image()`, where this function chooses between the remote-URL branch and the local-file branch. An uppercase scheme takes the local-file branch, `Path()` collapses the double slash, and the image is read from disk:

```python
import pandas as pd
from great_tables import GT

GT(pd.DataFrame({"img": ["HTTPS://posit.co/logo.png"]})).fmt_image(columns="img").as_raw_html()
```

```
FileNotFoundError: [Errno 2] No such file or directory:
'/current/dir/HTTPS:/posit.co/logo.png'
```

With `encode=False` there is no error at all, which is worse: the `src` attribute silently becomes a local path built from the remote URL.

After the change, both cases render the same tag, and the scheme is passed through as the user wrote it (browsers match schemes case-insensitively too):

```
<img src="HTTPS://posit.co/logo.png" ...>
```

Changes:

- `is_valid_http_schema()` lowercases before matching the prefix, with a comment naming the RFC.
- The existing parametrized test gains four uppercase and mixed-case spellings, plus a new negative test (`posit.co`, `ftp://posit.co`, `/tmp/http://x.png`, `httpx://posit.co`, `""`) so the fix cannot be read as "accept anything containing http".

One honest note on precedent: the R **gt** package solves the same question differently, with `grepl("https?://", path)`, which is also case-sensitive but unanchored, so it matches the substring anywhere in the value. I kept the Python behavior anchored at the start and only made the comparison case-insensitive, which is the narrower change and matches what `urlparse` does.

Full test suite passes locally (including the snapshot tests).

- Ref: #874, which fixed the same class of bug for locale names (`pt-br` vs `pt-BR`).

# Checklist


rich-iannone (approved):
Thanks, looks good!

PR #877, opened 2026-09-21
test(helpers): stop the parametrized _intify_scaled_px test from being shadowed

Rodrigo-Palma:
# Summary

`tests/test_helpers.py` defines `test_intify_scaled_px` twice, once as a parametrized test and once as a plain one:

```python
# line 408
@pytest.mark.parametrize(
    "value, scale, expected", [("0.5px", 0.5, 0), ["1px", 1, 1], ["2.1px", 2.1, 4]]
)
def test_intify_scaled_px(value: str, scale: float, expected: int):
    assert _intify_scaled_px(value, scale) == expected

# line 488
def test_intify_scaled_px():
    from great_tables._helpers import _intify_scaled_px

    assert _intify_scaled_px("10px", 2.0) == 20
    assert _intify_scaled_px("5px", 1.5) == 7
```

The second definition rebinds the module-level name, so the parametrized one is unreachable by the time pytest collects the module. Its three cases have never run:

```
$ pytest tests/test_helpers.py -k intify --collect-only -q
before: tests/test_helpers.py: 1
after:  tests/test_helpers.py: 5
```

This matters because the shadowed cases are the interesting ones: `("0.5px", 0.5, 0)` is the case where a scaled padding collapses to zero, and `("2.1px", 2.1, 4)` pins the truncation. The surviving test covered neither.

The change merges all five cases into the parametrized test and deletes the duplicate. I also added a comment on `("5px", 1.5, 7)` noting that the result is truncated toward zero rather than rounded, so a future reader does not mistake the assertion for a rounding bug.

I checked the rest of the suite for the same problem with an AST scan of every top-level function in `tests/`: this was the only shadowed definition.

No production code is touched and the full suite passes.

- Ref: #

# Checklist


rich-iannone (approved):
LGTM, thanks!

PR #879, opened 2026-09-21
fix(options): keep the fractional part when scaling padding

### LINKED ISSUE #878: opt_vertical_padding() and opt_horizontal_padding() truncate the scaled value, diverging from R gt
## Prework

- [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) and contributing guidelines.
- [x] Searched existing issues. #822 is about the `px` suffix being stripped from padding and #153 is the original request for these methods; neither covers the rounding question below.

## Description

`opt_vertical_padding()` and `opt_horizontal_padding()` truncate the scaled padding to an integer, so some scale factors produce a visibly different table than the same call in the R **gt** package, and small paddings can collapse to `0px`.

Both methods go through `_intify_scaled_px()`:

```python
def _intify_scaled_px(v: str, scale: float) -> int:
    return int(float(v.removesuffix("px")) * scale)
```

and both call sites carry the note `TODO: another limitation is that the returned values must be in integer pixel values`.

R **gt** does not do this. `get_padding_option_value_list()` in `R/opts.R` multiplies and re-attaches the unit, keeping the fractional value:

```r
padding_options$px <- padding_options$px * scale
create_option_value_list(padding_options$parameter, paste0(padding_options$px, "px"))
```

CSS accepts fractional pixel lengths, so `padding: 2.5px` is valid and is what R emits.

## Reproducible example

```python
from great_tables._helpers import _intify_scaled_px

for value, scale in [("5px", 0.5), ("4px", 0.1), ("5px", 0.3), ("5px", 2.9)]:
    print(value, scale, "->", f"{_intify_scaled_px(value, scale)}px",
          "| R:", f"{float(value.removesuffix('px')) * scale}px")
```

```
5px 0.5 -> 2px  | R: 2.5px
4px 0.1 -> 0px  | R: 0.4px
5px 0.3 -> 1px  | R: 1.5px
5px 2.9 -> 14px | R: 14.5px
```

The defaults make this reachable without unusual input: `heading_padding` is `4px` and the horizontal paddings are `5px`, so `opt_horizontal_padding(scale=0.5)` already diverges, and any scale at or below `0.2` flattens `heading_padding` to `0px` while R keeps a visible sliver.

Truncation also goes toward zero rather than rounding, which the suite pins today: `_intify_scaled_px("5px", 1.5) == 7`, not `8`.

## Expected result

Either the scaled value keeps its fractional part, matching R (`2.5px`), or the truncation is a deliberate decision, in which case the `TODO` should be turned into a comment saying so, since "limitation" reads like something meant to be lifted.

I did not send a patch because this changes rendered output for existing tables and the call is yours. Happy to open one in whichever direction you prefer: keeping the fraction is a two-line change plus snapshot updates, and rounding instead of truncating is smaller still.

Rodrigo-Palma:
# Summary

Implements what you asked for in #878: the scaled padding keeps its fractional part, matching R **gt**.

`_intify_scaled_px()` is replaced by `_scaled_px()`:

```python
def _scaled_px(v: str, scale: float) -> int | float:
    scaled = round(float(v.removesuffix("px")) * scale, _SCALED_PX_DIGITS)

    return int(scaled) if scaled.is_integer() else scaled
```

Both `TODO: another limitation is that the returned values must be in integer pixel values` comments are gone, and the call sites are unchanged apart from the function name, since `px()` already accepts `int | float`.

Two decisions worth flagging, because neither was in the issue:

**A whole-number result comes back as an `int`.** Otherwise `8px` scaled by `1.5` would render as `"12.0px"`, and R emits `"12px"` there. `paste0(12, "px")` in R drops the trailing zero, so returning an `int` is what keeps the two implementations matching on the common case rather than only on the fractional one.

**The result is rounded to 4 decimal places, and that is not cosmetic.** `5px` scaled by `0.07` is `0.35000000000000003` in binary floating point, and without rounding that string goes into the stylesheet. I measured it across the accepted scale range (`0.01` to `3.00`, in `0.01` steps, against the padding defaults): **263 of 1800 combinations produce an artifact like that**. Four decimals is far below anything a browser renders differently, and it keeps the CSS readable.

## Before and after

```python
from great_tables._helpers import _scaled_px

for value, scale in [("5px", 0.5), ("4px", 0.1), ("5px", 0.3), ("5px", 2.9), ("8px", 1.5)]:
    print(value, scale, "->", f"{_scaled_px(value, scale)}px")
```

| value | scale | before | after | R gt |
|---|---|---|---|---|
| 5px | 0.5 | `2px` | `2.5px` | `2.5px` |
| 4px | 0.1 | `0px` | `0.4px` | `0.4px` |
| 5px | 0.3 | `1px` | `1.5px` | `1.5px` |
| 5px | 2.9 | `14px` | `14.5px` | `14.5px` |
| 8px | 1.5 | `12px` | `12px` | `12px` |

End to end, on the rendered table:

```python
GT(pd.DataFrame({"x": [1]})).opt_horizontal_padding(scale=0.1).as_raw_html()
```

`padding-left` goes from `0px` to `0.5px`, and `scale=1.0` and `scale=2.2` still produce `5px` and `11px` with no trailing zero.

## Snapshots

You asked me to update them, and I want to be explicit about why the `.ambr` files are untouched: **no snapshot exercises this path**. `opt_vertical_padding` and `opt_horizontal_padding` appear only in `tests/test_options.py`, which asserts on option values directly rather than against a snapshot. The full suite passes with all 192 snapshots green and none regenerated. If you would rather have a snapshot covering a scaled table, say the word and I will add one.

## Tests

- `test_scaled_px` rewritten for the new behavior, covering whole-number results (`8px` × 1.5 → `12`), fractional ones (`5px` × 0.5 → `2.5`), the case that used to collapse (`4px` × 0.1 → `0.4`), and the floating point artifact (`5px` × 0.07 → `0.35`).
- `test_scaled_px_renders_without_a_trailing_zero` pins the rendered string through `px()`, which is the part that actually reaches the stylesheet.
- `test_opt_vertical_padding` and `test_opt_horizontal_padding` updated, each with one extra case at a small scale, which is the regression this fixes.
- Removed an unused `_intify_scaled_px` import in `tests/test_options.py`.

## Verification

Full suite passes (192 snapshots). On `ruff` 0.8.0, the version pinned in `.pre-commit-config.yaml`, the four touched files report **9 errors with this branch and 10 on `main`**: all of them pre-existing `E712` comparisons in `tests/test_options.py` lines 268 to 430, which I did not touch. The one that disappeared is the unused import above. `ruff format --check` is clean.

- Closes #878

# Checklist


rich-iannone (approved):
This is very good, thank you!

PR #880, opened 2026-09-21
fix(render): render booleans, missing values and zero the same on every backend

### LINKED ISSUE #875: Boolean, missing and signed-zero cells render differently for pandas, polars and pyarrow
## Prework

- [x] Searched the open and closed issues; I did not find this reported.

## Description

The same logical table renders differently depending on which dataframe it was built from. `cast_frame_to_string` dispatches to each backend and lets that backend decide how a value becomes text, so a boolean, a missing value and a signed zero each come out in a different notation for pandas, polars and pyarrow.

I am filing this rather than opening a pull request because fixing it means deciding which spelling is the canonical one for **great_tables**, and that is a call for the maintainers rather than something to settle inside a patch.

## Reproducible example

```python
import re

import pandas as pd
import polars as pl
import pyarrow as pa
from great_tables import GT

data = {"flag": [True, False, None], "zero": [-0.0, 0.0, None]}

for name, df in [
    ("pandas", pd.DataFrame(data)),
    ("polars", pl.DataFrame(data)),
    ("pyarrow", pa.table(data)),
]:
    html = GT(df).as_raw_html()
    print(name, re.findall(r'<td[^>]*class="gt_row[^"]*"[^>]*>(.*?)</td>', html, re.S))
```

Output:

```
pandas  ['True', '-0.0', 'False', '0.0', '<NA>', '<NA>']
polars  ['true', '-0.0', 'false', '0.0', 'None', 'None']
pyarrow ['true', '-0',   'false', '0',   'None', 'None']
```

| value | pandas | polars | pyarrow |
|---|---|---|---|
| `True` | `True` | `true` | `true` |
| missing | `<NA>` | `None` | `None` |
| `-0.0` | `-0.0` | `-0.0` | `-0` |

## Expected result

One table, one rendering. A reader should not be able to tell which dataframe library produced the table, and moving a script from pandas to polars should not silently change what the cells say.

Which spelling wins is the open question. `true`/`false` matches two of the three backends, `True`/`False` matches Python and the current pandas output, and the R **gt** package prints `TRUE`/`FALSE`. The same question applies to the missing marker, where `None` and `<NA>` are both internal repr spellings leaking into the rendered table, and to whether `-0` and `-0.0` should agree.

## Development environment

- Operating System: macOS (arm64)
- **great_tables** Version: reproduced on `main` at `dfc9b40de`
- pandas 3.0.5, polars 1.44.2, pyarrow 25.0.1, Python 3.12.13

## Additional context

The dispatch lives in `great_tables/_tbl_data.py`, in `cast_frame_to_string`: the pandas branch calls `astype("string")`, the polars branch `cast(pl.Utf8)`, and the pyarrow branch `cast(pa.string())`. Each one inherits its engine's notation, which is where the three spellings come from.

`fmt_tf()` and `sub_missing()` let a user pin the output explicitly, so this is about the default for a column that is not formatted.

Happy to open a PR once you decide which spelling should win, including the tests to keep the three backends in agreement.

Rodrigo-Palma:
# Summary

Implements the decision you gave in #875: booleans render as `True`/`False`, missing values as `None`, and zero carries no sign, on all three backends.

Before:

```
value    pandas   polars   pyarrow
True     True     true     true
missing  <NA>     None     None
-0.0     -0.0     -0.0     -0
```

After, for the reproducer in the issue, all three produce `['True', '0.0', 'False', '0.0', 'None', 'None']` (pyarrow writes `0` rather than `0.0`, see the note at the end).

## How each rule is applied

**Booleans** are normalized in the polars and pyarrow branches, which spelled them lowercase. This is done through `cs.boolean()` and a `pa.types.is_boolean` check rather than a string replacement, so **a text column that happens to contain `"true"` is left untouched**. There is a test for exactly that, since a blanket replace would have been the easy way to introduce a worse bug than the one being fixed.

**Zero** loses its sign by adding `0.0` to floating point columns before the cast: `-0.0 + 0.0` is `+0.0` and every other value is unchanged, including `NaN` and infinities. One wrinkle worth flagging: in polars adding zero is optimized away and the `-0.0` survives, so that branch uses an explicit `when(col == 0).then(0.0)`.

**Missing values** were the only part that reached beyond `cast_frame_to_string`. pandas produces `pd.NA`, which renders as `<NA>`, and converting it to `None` is not enough on its own: under the `"string"` dtype, a `None` coming from the replacement frame is coerced straight back to `pd.NA` by `fillna`. So `replace_null_frame` for pandas now casts to `object` first. That is the one line in this PR outside the function named in the issue, and I wanted to call it out rather than have you find it in the diff.

## Snapshots

Seven snapshots updated, in `test_export.ambr`, `test_formats.ambr` and `data_color/test_data_color.ambr`. I audited every changed line: **all 21 of them are `<NA>` → `None`**, with no other difference. The remaining 185 snapshots pass untouched.

## Tests

Three new tests in `tests/test_tbl_data.py`, each parametrized over pandas, polars and pyarrow:

- `test_cast_frame_to_string_is_consistent_across_backends` — booleans and missing values agree, and the `"true"` string column is preserved.
- `test_cast_frame_to_string_drops_the_sign_on_zero` — `-0.0` and `0.0` render identically, while `-1.5` keeps its sign.
- `test_render_body_is_consistent_across_backends` — checks the rendered body directly, asserting `<NA>` appears nowhere.

All nine cases fail on `main` and pass here.

## One thing I deliberately did not change

pyarrow still renders a whole float as `0` where pandas and polars render `0.0`. #875 settled the spelling of booleans, missing values and signed zero, and this is float formatting instead, which feels like it belongs to `fmt_number()` rather than to a cast. Happy to take it on in a follow-up if you want the three backends to agree there too.

## Verification

Full suite passes (192 snapshots). `ruff` 0.8.0 (the pinned version) reports no errors on the touched files and `ruff format --check` is clean.

- Closes #875

# Checklist


rich-iannone (approved):
LGTM!

